Issue #64, opened 2026-04-16
feat: optimize() should route through forward() to support NU mesh


bk-squared:
## Problem

`optimize()` and `topology_optimize()` call `sim._forward_from_materials()` directly, bypassing the public `sim.forward()` path. This means:

1. Non-uniform mesh is not available for `optimize()` — NU support lives in `forward()` → `_forward_nonuniform_from_materials()`
2. New features added to `forward()` (checkpoint_every, emit_time_series) are not available in `optimize()`
3. The API split creates maintenance burden — features must be added in two places

## Current Code Path

```
optimize()  → _forward_from_materials()  → uniform runner only
forward()   → _forward_nonuniform_from_materials()  → NU + checkpoint + emit_ts
```

## Desired Code Path

```
optimize()  → forward()  → dispatches to uniform or NU automatically
```

## Impact

Users who want to combine `DesignRegion` continuous optimization with NU mesh (e.g., DRA on thin substrate with fine z-cells in substrate) currently cannot use the `optimize()` convenience API. They must write a manual `jax.value_and_grad` loop with `forward(eps_override=...)`.

Same for `topology_optimize()` — though soft PEC on NU is a separate blocker (#45).

## Suggested Approach

Refactor `optimize()` internal forward call from:
```python
result = sim._forward_from_materials(grid, materials, ...)
```
to:
```python
result = sim.forward(eps_override=eps_r, checkpoint=True, n_steps=_n_steps)
```

This requires `forward()` to handle the material assembly internally (which it already does for the NU path).

bk-squared:
Implemented in `70a8be9`. `optimize()` now builds `eps_override` from the design region overlay and delegates to `sim.forward()` — uniform/NU dispatch happens inside `forward()`. New kwargs forwarded through: `checkpoint_every`, `n_warmup`, `emit_time_series`.

Test: `tests/test_optimize_nonuniform.py` — dz_profile optimize() now produces finite loss (previously raised ValueError). 27 existing optimize/auto-preflight tests still pass.

Issue #63, opened 2026-04-16
preflight: warn when minimize_reflected_energy used with soft source instead of port


bk-squared:
## Problem

`minimize_reflected_energy()` computes the ratio of late-time to early-time energy in a probe time series. This is physically meaningful only when the probe is co-located with an impedance-matched port (`add_port(impedance=50)`), which absorbs backward-traveling waves.

When used with `add_source()` (soft source, zero impedance), the source does not absorb reflected waves. They pass through, re-reflect from boundaries (CPML), and create a complex standing-wave pattern at the probe. The energy ratio becomes physically meaningless.

## Discovery Context

rfx-TAP ex6 absorber used `add_source` at z=26mm + `add_probe` at z=24mm + `minimize_reflected_energy`. The optimization converged (loss decreased) and reported avg S11=-21.9 dB, but this number was likely an artifact of the measurement setup, not genuine absorber performance.

After switching to `add_port(impedance=50)` co-located with the probe, the S11 is being re-evaluated (results pending).

## Proposed Check

When `minimize_reflected_energy` is called on a ForwardResult, check if the simulation has at least one port. If only soft sources exist:

```python
if not self._ports and self._sources:
    warnings.warn(
        "minimize_reflected_energy() is most accurate with an impedance-matched "
        "port (add_port). Using it with a soft source (add_source) may give "
        "incorrect S11 estimates because reflected waves are not absorbed.",
        UserWarning,
    )
```

Or alternatively, check at `forward()`/`optimize()` entry when the objective is known.

## Severity

MEDIUM — the optimization still converges (loss decreases), but the physical interpretation is wrong. Difficult to diagnose without domain expertise.

bk-squared:
Fixed on branch `fix/issue-62-63-preflight-warnings` and submitted for review in PR #65: https://github.com/bk-squared/rfx/pull/65

Issue #62, opened 2026-04-16
preflight: detect probe placed inside PEC volume


bk-squared:
## Problem

When a probe is placed inside a PEC volume, it records zero E-field at all timesteps. If used with `minimize_reflected_energy`, the loss becomes 0/0 = NaN, causing the optimizer to diverge with all-NaN parameters.

## Discovery Context

rfx-TAP ex6 absorber had `PROBE_Z = 0.5mm` inside PEC slab (z=0..1mm). The optimizer produced `per_layer_eps = [NaN, NaN, NaN, NaN, NaN]`. No warning was emitted.

## Proposed Check

At preflight or `forward()`/`run()` entry:

```python
for probe in self._probes:
    for geo in self._geometry:
        if geo.material == 'pec' and geo.box.contains(probe.position):
            warnings.warn(
                f"Probe at {probe.position} is inside PEC volume '{geo.name}'. "
                f"E-field is zero inside PEC — probe will record no signal.",
                UserWarning,
            )
```

## Severity

HIGH — NaN gradients with no diagnostic. Users will spend hours debugging.

bk-squared:
Fixed on branch `fix/issue-62-63-preflight-warnings` and submitted for review in PR #65: https://github.com/bk-squared/rfx/pull/65

PR #65, opened 2026-04-16
Add preflight warnings for PEC probes and soft-source reflected-energy use

bk-squared:
## Summary

This PR fixes two small but high-signal diagnostics issues:

- #62: warn when a probe is placed inside a PEC volume
- #63: warn when `minimize_reflected_energy()` is used in a soft-source-only measurement context

### #62 probe-inside-PEC detection
- Added a preflight warning in `Simulation._validate_simulation_config()` for probes inside PEC geometry
- Added focused regression coverage in `tests/test_api.py`
- Added the same validation hook to `optimize()` and `topology_optimize()` so the warning also appears on optimization paths that currently bypass `forward()`

### #63 reflected-energy proxy warning
- Added source/port context metadata to `Result` and `ForwardResult`
- `minimize_reflected_energy()` now warns once per objective instance when the result came from a soft-source-only setup with no matched port
- Added focused regression coverage in `tests/test_optimize_proxy_objectives.py`

## Why

Both issues are easy to miss and can waste a lot of debugging time:
- a probe inside PEC records zero field and can produce NaN optimization behavior
- reflected-energy proxy loss can look reasonable while the physical S11 interpretation is wrong if there is no matched port

## Verification

```bash
python -m compileall rfx/api.py rfx/optimize_objectives.py rfx/optimize.py rfx/topology.py tests/test_api.py tests/test_optimize_proxy_objectives.py
pytest -q tests/test_api.py -k 'probe_inside_pec or upml_rejects_subgridding_refinement'
pytest -q tests/test_optimize_proxy_objectives.py -k 'warns_when_only_soft_source_context_is_present or no_warning_when_matched_port_context_is_present or forward_attaches_measurement_context_counts or actual_soft_source_forward_result or optimize_warns_when_probe_is_inside_pec'
pytest -q tests/test_optimize_proxy_objectives.py tests/test_optimize.py tests/test_topology.py -k 'not distributed and (optimize or reflected_energy or transmitted_energy or topology_optimize)'
```

## Notes for review

This PR is intentionally narrow:
- warning-only behavior, not semantic changes to simulation math
- preserves differentiable paths
- does not attempt the larger `optimize()/topology_optimize() -> forward()` reroute from #64

### DIFF
--- a/rfx/api.py
+++ b/rfx/api.py
@@ -144,6 +144,8 @@ class Result(NamedTuple):
     grid: object = None
     dt: float | None = None
     freq_range: tuple | None = None
+    matched_port_count: int | None = None
+    soft_source_count: int | None = None
 
     def find_resonances(self, freq_range=None, probe_idx=0,
                          source_decay_time=None, bandpass=None):
@@ -228,6 +230,8 @@ class ForwardResult(NamedTuple):
     grid: object = None
     s_params: object = None
     freqs: object = None
+    matched_port_count: int | None = None
+    soft_source_count: int | None = None
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


bk-squared:
Closing — probe-in-PEC check was already added to P1.8 on main (commit 444eb4f, combined ports+probes into one loop). The soft-source reflected-energy warning is a good idea but this branch has substantial merge conflicts with the current main (preflight was heavily refactored in the v1.6.0 session). Please rebase and open a focused PR for just the soft-source warning if still wanted.

PR #79, opened 2026-04-30
feat(api): plumb WirePort S-params through forward(port_s11_freqs=...)

bk-squared:
## Summary

Wire ports added via `add_port(extent=...)` were silently dropped from the `forward(port_s11_freqs=...)` S-param dispatch — the api layer only registered `LumpedPortSParamSpec` for single-cell lumped ports, leaving the underlying `WirePortSParamSpec` and the `wire_sparam_meta` JIT-scan path unused. As a result, `ForwardResult.s_params` was `None` and `ForwardResult.wire_port_sparams` was `None` for any wire-fed device, blocking AD-based optimisation of substrate-spanning probe-fed antennas.

## Mechanism

`rfx/simulation.py` already exposes both:
- `WirePortSParamSpec` (line 51) — V/I DFT spec at the wire midpoint cell
- The `wire_sparam_meta` accumulator path inside both the Python-loop and JIT-scan bodies — accumulates `(v_dft, i_dft, v_inc_dft)` BEFORE source injection (per the `update_sparam_probe` docstring contract; PR #78)

But `Simulation._forward_from_materials` (api.py) registered nothing for the wire path:

```python
if pe.extent is not None:
    # build WirePort + sources + clear PEC mask under wire cells
    ...
    continue   # ← never registers a WirePortSParamSpec
```

Even when `_s11_freqs_arr is not None`. So `_run(...)` always got `wire_port_sparams=None`.

## Fix

1. In the WirePort branch of the port loop in `_forward_from_materials`: when `_s11_freqs_arr is not None`, build a `WirePortSParamSpec` anchored at the wire's midpoint cell (`wp_cells[len(wp_cells)//2]`) and append it to a new `wire_port_sparam_specs` list.
2. Pass `wire_port_sparams=wire_port_sparam_specs or None` to `_run(...)` (kwarg already supported).
3. Post-`_run`: when `result.wire_port_sparams` is populated, run the same wave-decomposition extraction as the lumped path — `extract_lumped_s11` works for wire ports because both use the FDTD sign convention `V = -E·dx`, so `S11 = (V + Z0·I) / (V − Z0·I)` is identical.
4. `ForwardResult` gains a `wire_port_sparams` field mirroring the existing `lumped_port_sparams`.

## Tests

New `tests/test_wire_port_sparams_forward.py`:
- `test_forward_wire_port_sparams_populated` — closed PEC cavity with WirePort (`extent=0.015`), confirms `result.s_params.shape == (n_freqs,)` and `|S11| ≈ 1` in band.
- `test_forward_wire_port_grad_flows` — `jax.grad` of mean(|S11|²) wrt `eps_override` returns finite, non-zero gradient → wire-port path is now AD-friendly.

Existing test parity (15/15 pass): `tests/test_lumped_port_sparams_jit.py` (3/3) + `tests/test_wire_sparam.py` (Python-loop, 6/6) + `tests/test_twoport_wire_port.py` (3/3) + `tests/test_sparam.py` (3/3).

## Out of scope

- Multi-port wire S-matrix (S21, S22, …): `extract_lumped_s11` reuse only computes self-reflection. Multi-port AD objectives still compose their own wave decomposition from raw `(v_dft, i_dft, v_inc_dft)`. The existing `extract_s_matrix_wire` (non-AD Python-loop) remains for that workflow.
- WirePort current is currently the curl-H loop integral × dx at the midpoint cell (same as lumped). For a multi-cell wire the more accurate I would be the cell-by-cell average; this PR keeps the midpoint-cell approximation consistent with `wire_sparam_meta`.

## Downstream user

This unblocks `rfx-tap` Phase 2: Yi 2019 TSPAA reproduction at 24.125 GHz on RO4350B, where field-level analysis showed clean traveling-wave phase distribution along the 14-element array but the api couldn't return a usable `|S11|` because the coax-modelled WirePort was unplumbed.

### DIFF
--- a/rfx/api.py
+++ b/rfx/api.py
@@ -1809,6 +1809,7 @@ class Simulation:
                 pec_faces=self._pec_faces,
                 pmc_faces=self._boundary_spec.pmc_faces(),
                 face_layers=face_layers,
+                conformal_faces=self._boundary_spec.conformal_faces(),
             )
         return Grid(
             freq_max=self._freq_max,
@@ -1821,6 +1822,7 @@ class Simulation:
             pec_faces=self._pec_faces,
             pmc_faces=self._boundary_spec.pmc_faces(),
             face_layers=face_layers,
+            conformal_faces=self._boundary_spec.conformal_faces(),
         )
 
     def _resolve_face_layers(self) -> dict:
@@ -1988,16 +1990,24 @@ class Simulation:
                 corner_lo = [-big, -big, -big]
                 corner_hi = [big, big, big]
                 if side == "lo":
+                    # Skip when the wall coincides with the grid origin
+                    # (y=0 PEC face): the binary apply_pec_faces handles
+                    # this exactly and a Dey-Mittra Box at corner_hi=0
+                    # would impose a spurious 0.5 weight on the cell at
+                    # j=0. Only inject when an actual interior region
+                    # past the lo face needs to be PEC-fied.
                     if wall_lo <= 0.0:
-                        # No fractional cell on the lo face: domain edge
-                        # already coincides with the physical wall.
                         continue
                     corner_hi[axis_idx] = wall_lo
                 else:  # hi
-                    if wall_hi >= float(self._domain[axis_idx]):
-                        # No fractional cell on the hi face: domain
-                        # edge already coincides with the physical wall.
-                        continue
+                    # Always inject on the hi side. The grid often
+                    # extends past ``self._domain`` due to dx-snap or
+                    # CPML padding on other axes, so a fractional cell
+                    # exists at the wall even when ``wall_hi`` equals
+                    # the user-declared domain extent. When no
+                    # fractional cell is present (grid edge ≤
+                    # wall_hi), the SDF naturally produces weight=1
+                    # everywhere and the Box is a harmless no-op.
                     corner_lo[axis_idx] = wall_hi
                 pec_shapes.append(Box(tuple(corner_lo), tuple(corner_hi)))
 
--- a/rfx/api.py
+++ b/rfx/api.py
@@ -2199,23 +2199,39 @@ class Simulation:
             full-record DFT is finite-energy on the recorded transient
             so no gating is needed even on strong reflectors.
         normalize : bool
-            When True, run a two-run normalization that cancels Yee-grid
-            numerical dispersion.  A reference simulation with vacuum (no
-            user geometry) is run automatically, and the device S-params
-            are divided by the reference incident waves.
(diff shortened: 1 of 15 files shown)


bk-squared (commented):
## Review summary

**Verdict**: APPROVE-CONDITIONAL — math + plumbing look correct, but I'd like 3 confirmations before merge.

### What works

The diagnosis (`Mechanism` section in PR body) is exactly right: the api `_forward_from_materials` registered `LumpedPortSParamSpec` for the lumped path but silently dropped the wire-port path before `_run`. The fix is symmetric to the lumped registration.

Reusing `extract_lumped_s11(v_dft, i_dft, z0)` for the wire path is a defensible choice given the FDTD sign convention `V = -E·dx` is the same for both — the wire's midpoint-cell V has identical units to the lumped single-cell V.

Two new tests are load-bearing: closed PEC cavity → `|S11|≈1` is the natural sanity, and `jax.grad` finite + non-zero confirms AD reachability through the new spec registration.

### 3 items to confirm before merge

1. **CI status** (`gh pr checks 79` shows "no checks reported"). Did the repo's CI workflow ever run, or is this PR not gated by automated tests? Confirm the new `tests/test_wire_port_sparams_forward.py` passes locally / on VESSL before merge.

2. **`extract_lumped_s11` reuse — inline comment**. The PR body justifies the reuse, but the code at line ~4493 just imports + calls. Recommend a 1-line comment in the elif branch noting *why* it's safe (same V scale at midpoint cell), so a future reader doesn't have to dig the PR body out of git history.

3. **Midpoint cell convention `wp_cells[len(wp_cells)//2]`**. For an even-length `wp_cells`, this picks the upper half's first cell (asymmetric). Does this match the convention used in `wire_sparam_meta` at `rfx/simulation.py:51` (`WirePortSParamSpec` definition + scan accumulator)? If yes, log it; if no, prefer `(len(wp_cells)-1)//2` for symmetry.

### Out-of-scope items the PR self-disclosed (good)

- Multi-port wire S-matrix (S21/S22) deferred to existing `extract_s_matrix_wire` non-AD path — correct boundary.
- Cell-by-cell I averaging deferred — consistent with `wire_sparam_meta` midpoint approximation.

### Bottom line

Looks ready to merge once items 1-3 are confirmed. The fix unblocks the rfx-tap Phase 2 use case (24.125 GHz TSPAA on RO4350B) per the PR body — this is real downstream value, not speculative.

🤖 Reviewed during Stage 2 step 3b work (Stage 2 commits 73adf5d..9eaafc5 in parallel; this PR is independent of Stage 2 scope).

PR #81, opened 2026-05-16
Guarded FDTD subgrid envelope (experimental) — Stage-1 mesh intelligence + Phase-3b decomposition

bk-squared:
## Summary

This branch lands the **guarded FDTD subgrid envelope** plus its supporting structural
work, developed and reviewed in an isolated clone of `rfx`. It is delivered as an
**experimental, opt-in** capability — see the honest scope note below.

The work is the union of three landed branches:

- **Branch A — Stage-1 mesh intelligence.** Mesh-refinement planning artifacts: the
  Stage-1 nu-physics gate, AD-memory estimation, and the memory-reduction planning
  report.
- **Branch B1 — guarded subgrid clean core.** The structurally-clean subgrid surface:
  `validation.py`, `material_sat.py`, the disjoint runner contract, the `add_refinement`
  public API, and the fail-closed `validate_subgrid` lane. Cut from `main` and verified
  **zero-regression** against it (`test_sbp_sat_jit.py` + `test_api.py`: 57 passed / 1
  pre-existing fail — identical to `main`).
- **Branch B2 — subgrid runner + Phase-3b decomposition.** The `jit_runner.py` /
  `disjoint_3d.py` runner files, plus the Phase-3b structural cleanup.

## Phase-3b structural work (GATE-B2)

A mechanical gate (`scripts/gate_b2_check.py`) enforces four structural invariants. **All
gates now pass:**

| Gate | Check | Status |
|---|---|---|
| G1 | no `jit_runner.py` function > 600 AST lines | PASS |
| G2 | no function > 15 params | PASS |
| G3 | 0 dead-knob occurrences (`material_sat_e_h_trace_*` / `box_shadow_sync_*`) | PASS |
| G4 | ≤ 1 `step_disjoint_z_slab*` stepper | PASS |

- **US-09** decomposed `run_subgridded_jit`: 52→6 params via a `SubgridRunOptions`
  NamedTuple bundle (unpack-at-top, byte-unchanged body), the two coupling functions
  22/23→6 params likewise, and the 477-line nested `step_fn` hoisted into a module-level
  `_make_step_fn(ctx)` factory (ctx built from `locals()` before the `lax.scan`, `ctx.get`
  unpack — preserving the original closure's lazy free-variable semantics).
- **US-10** stripped the dead-knob surface (`jit_runner.py` 3179→2557 lines) and kept the
  production-validation tombstones so stale overrides still fail closed.
- **US-11** collapsed 9 `step_disjoint_z_slab_*` steppers into 1.

**G5 behavior suite:** the jit / coupling / disjoint test suite passes (21 tests);
behavior is preserved through the decomposition. The long-run instability test is locked
as `xfail(strict=True)`.

## Honest scope note

This is an **experimental, guarded one-sided z-slab subgrid envelope** — not the
North-star general embedded subgrid, which **remains incomplete**. The Stage-2 disjoint
prototype is research-only and **not long-time energy-stable**; it is labelled as such in
the code and examples. The `add_refinement` → `validate_subgrid` lane is fail-closed: an
out-of-envelope config produces a structured `SubgridValidationReport`.

## Review summary

The work passed through three review passes before landing.

**1. Entry review of the prior subgrid prototype.** The pre-existing Codex-built subgrid
program was reviewed multi-agent (opus + scientist + critics + program-reviewer). Verdict:
structurally heavy, with honesty gaps in the status docs and a true general embedded
subgrid not achieved. That review defined the forward scope (Phase 2 doc honesty → Phase
3a/3b structural work).

**2. Consensus planning (Planner / Architect / Critic).** The forward plan was reviewed to
consensus. The review **caught a real plan flaw**: the proposed blanket N-1 honesty guard
(`run()` on any refined sim raises) would have regressed 12 of `main`'s passing subgrid
tests, since B1 inherits `main`'s working `overlap_z_slab` runner. The guard was narrowed
to fire only for `topology="stage2_disjoint_3d"` — the new, unfinished runner — restoring
zero-regression.

**3. Parallel Claude + Codex critics on the implementation.**

- **Claude critic** verified **US-11 solid** — the 9→1 stepper collapse lost no live
  coverage; the deleted steppers' tests were genuinely orphaned.
- **Codex critic** returned `unsound` on **one** finding: US-10's dead-knob strip had also
  dropped the production-validation *rejection* of those knobs — a fail-closed regression
  where stale `material_sat_e_h_trace_*` / `box_shadow_sync_*` overrides would be silently
  accepted. **Fixed in `3327992`** before landing: `_active_diagnostic_overrides` now
  tombstones the removed prefixes (verified — `test_subgrid_validation.py` 65 passed).
- Codex independently **confirmed the narrowed B1 N-1 guard correct** and ran a B2 test
  subset green (38 passed / 1 xfailed).

Two minor Codex notes are carried as non-blocking follow-ups: (a) US-11 also narrowed some
live *private* `disjoint_stepper` selector values (intentional research-surface
narrowing); (b) the US-06 long-run lock exercises `step_disjoint_sat_3d` while the
surviving keep-path stepper is `step_disjoint_z_slab_sat_3d` — consider a long-run lock for
the keep-path.

## Verification

- GATE-B2: all of G1–G4 pass (`scripts/gate_b2_check.py`).
- B1 zero-regression vs `main`: 57 passed / 1 pre-existing fail.
- G5 behavior suite: 21 passed; long-run instability test `xfailed` as designed.
- All 12 forward-program user stories (US-01 … US-12) delivered and verified.


bk-squared:
## External Codex critic review — verdict: ITERATE

An external Codex critic reviewed the full PR diff (`8fcac3c..b61486d`, 41 files). Codex's
sandbox blocked local test execution (`bwrap: Permission denied`), so it ran a **static**
review with GitHub cross-checks; every concrete claim below was independently verified
against the code and test runs.

### Findings

| # | Severity | Status | Finding |
|---|----------|--------|---------|
| 1 | medium | ✅ **Fixed** (`da4c24d`) | Broken crossval example import |
| 2 | high | ✅ **Fixed** (`d8fdfc3`) | Overlap fine-shape convention mismatch |
| 3 | low–med | ⚠️ Open (non-blocking) | Tombstone enforcement is production-mode-only |
| 4 | **high** | ⛔ **Open — confirmed failing** | `test_subgrid_n_minus_1_guard.py` — 2 tests fail on the PR head |
| 5 | low | ⚠️ Open (non-blocking) | `_make_step_fn` `ctx.get` latent fragility |

### Fixed in this push (`d8fdfc3`, `da4c24d`)

**#2 — overlap fine-shape convention.** `build_subgrid_region` reported the overlap
fine-grid shape as cell-extent `(hi-lo)*ratio`, while the overlap runner
`rfx/runners/subgridded.py` builds the executed grid node-aligned as `(hi-lo-1)*ratio+1`
— a `ratio-1` per-axis mismatch — and the code comment falsely claimed the two agreed.
US-04 unified the two *validation* region builders to each other, never against the
runner. Fixed by extracting `overlap_fine_extent()` as a single source of truth called by
both `build_subgrid_region` and the runner, so the reported region and the executed grid
cannot drift apart. `build_stage2_disjoint_region` correctly keeps cell-extent (it matches
its own `build_disjoint_runner_contract` default) — the two topologies legitimately
differ, so the parity test now checks each builder against its own runner.

**#1 — broken crossval example.** `examples/crossval/12_subgrid_disjoint_prototype.py`
imported `scripts.stage2_disjoint_full_physics_gate`, which is absent from the repo →
`ModuleNotFoundError` on the import line. Rewritten to drive the committed
`rfx.runners.disjoint.run_disjoint_stage2_path`; an import-smoke test was added.

Regression for this push: subgrid + SBP-SAT suite **96 passed**, GATE-B2 **all gates
pass**, ruff clean.

### ⛔ Finding #4 — escalation: 2 tests actually FAIL on the PR head

Codex flagged `test_subgrid_n_minus_1_guard.py` as having only a stale docstring.
Verification shows it is worse: **2 of its 3 tests fail on the pristine PR head
`b61486d`** — independent of this push's fixes:

```
$ pytest tests/test_subgrid_n_minus_1_guard.py
FAILED test_run_on_disjoint_topology_raises_not_implemented
FAILED test_run_on_overlap_topology_does_not_trip_the_n1_guard
2 failed, 1 passed
```

Both fail because production validation raises `ValueError`
(`z_slab_requires_guarded_boundary` / unsupported-production-z-slab) where the tests
expect `NotImplementedError`. Per `progress.txt`, US-07 left this file at **"3 passed" on
Branch B1** — so the 2 tests **silently regressed during the B2 / US-09 landing**, and the
"all 12 stories pass / GATE-B2 green" sign-off did not catch it.

This needs a decision — is `ValueError` from production validation the intended N-1 guard
signal now that `run_disjoint_stage2_path` exists? — plus a test update. It is out of
scope for the #1/#2 fix. **Recommend addressing before merge.**

### Non-blocking follow-ups

- **#3** — `_active_diagnostic_overrides` tombstone rejection sits inside the
  `mode == "production"` block in `validation.py`. In `research`/`off` mode a stale
  removed-knob key is not flagged. Low impact: the knobs are dead-stripped from the
  runner, so a stale key is inert in every mode — only the advisory is mode-gated.
- **#5** — `_make_step_fn` uses `ctx.get(name)`, so a missing *required* capture would
  resolve to `None` rather than raising as the original closure did. No currently
  misbound capture exists (behavior is preserved as shipped); a typed `StepContext` /
  `ctx[name]` would be safer against future edits.

### Optimality

Codex affirmed `SubgridRunOptions` is sound (immutable `NamedTuple`, no mutable-default
trap). It flagged the `locals()` + `ctx.get` step-fn hoist as auditable-but-not-cleanest
(a typed `StepContext` preferred) — captured as follow-up #5.

bk-squared:
### Finding #4 resolved (`6c525ac`)

The 2 failing `test_subgrid_n_minus_1_guard.py` tests are fixed. Root cause:
the tests carried Branch B1's placeholder `NotImplementedError` expectation,
but Branch B2 replaced that placeholder with the real runner dispatch — so
the loud-failure signal for an unsupported topology is now production
validation's `ValueError`. The tests were rewritten against the current,
correct behaviour:

- disjoint topology + production validation → loud `ValueError`;
- disjoint topology + `validation="research"` → dispatches to
  `run_disjoint_stage2_path`, returns a finite result;
- a guarded one-sided overlap z-slab validates as supported and runs.

No production code changed — the behaviour was already correct, only the
stale tests were wrong. `pytest tests/test_subgrid_n_minus_1_guard.py` →
4 passed.

Updated finding status: **#1, #2, #4 fixed**; **#3, #5** remain as the
non-blocking follow-ups noted above.

bk-squared:
## Verified full-suite run — closing evidence

Ran the complete subgrid + SBP-SAT + mesh-intelligence test surface — 23 files spanning
Branch A (mesh intelligence) + B1 + B2 + the six review-fix commits — at branch tip
`83120c1`, plus the GATE-B2 mechanical check.

**GATE-B2: ALL GATES PASS** — G1 (function size) · G2 (param count) · G3 (dead knobs) ·
G4 (stepper count).

**Test suite: `197 passed, 2 xfailed, 4 failed`** (493 s).

- **197 passed** — the entire subgrid / SBP-SAT / mesh-intelligence surface is green,
  including the review-fix tests (fine-shape parity, crossval example import, step-fn
  capture guard, removed-knob validation, the rewritten topology-guard tests).
- **2 xfailed** — the disjoint long-run instability locks (`step_disjoint_sat_3d` ~14×,
  `step_disjoint_z_slab_sat_3d` ~33× energy growth over 3000 steps). `xfail(strict=True)`
  by design — these now measure real energy divergence (previously they xfailed on a
  `ConcretizationTypeError`; fixed in `83120c1`).
- **4 failed** — all in `tests/test_physics_gate_reporting.py`, and **verified
  pre-existing on `main`**: re-running them on pristine `8fcac3c` gives the identical
  `AssertionError: assert 4 == 7` (only 4 of 7 RF port families have broad-E5 evidence).
  That file is a separate, in-progress **RF port-crossval gate** — not subgrid code — and
  the PR diff does not touch it.

  *Caveat for the reviewer / CI:* PR story US-05 declared the `pyyaml` dependency so this
  file *collects* (it previously hit a `ModuleNotFoundError: yaml` collection error). A
  side effect is that these 4 pre-existing RF-gate failures become **visible** in a full
  run post-merge, where the collection error previously masked them. This is not a
  regression introduced by the subgrid work — but expect them, or treat
  `test_physics_gate_reporting.py` as the separate RF-gate work item it is.

**Net:** the guarded subgrid envelope this PR lands is fully green and GATE-B2-compliant;
the only failures are pre-existing, unrelated RF-port-gate tests confirmed identical on
`main`.

### Review fixes landed on the branch since the initial review

| Commits | Finding |
|---|---|
| `d8fdfc3` | #2 — overlap fine-shape unified on the node-aligned runner convention |
| `da4c24d` | #1 — broken crossval example rewired to the committed disjoint runner |
| `6c525ac` | #4 — stale N-1 topology-guard tests repaired |
| `2891aa5` | #3 — removed-knob tombstones now fail closed in every validation mode |
| `0377023` | #5 — required step_fn captures fail loudly if absent from ctx |
| `83120c1` | disjoint long-run instability locks now measure real energy; keep-path lock added |

Issue #80, opened 2026-05-13
compute_msl_s_matrix reports wrong S11 resonance for thin-substrate patch at 10 GHz (|q|>1, 13% offset vs analytic Balanis)


bk-squared:
# `compute_msl_s_matrix` produces wrong S11 resonance frequency for edge-fed patch on RO4003C at 10 GHz

## TL;DR

For an edge-fed HAMMERSTAD patch on RO4003C (εr=3.38, h=0.787 mm, W=10.13 mm, L=8.59 mm, 50 Ω microstrip feed), `compute_msl_s_matrix` reports the S11 minimum at **10.11 GHz** while the analytic Balanis prediction is **9.21 GHz** and the actual patch Ez field DFT confirms resonance at **9.2 GHz** (16× more patch energy at 9.2 GHz vs 10.0 GHz). The 3-probe extractor is **not** producing the patch's intrinsic S11 — instead it converges to a geometry-dependent feature governed by the trace length and probe placement.

The `|q|` parameter that the wave-decomposition uses is observed at **> 1** for all tested configurations (1.05 → 1.30), which is non-physical for a passive line.

This affects any user doing AD-style inverse design of patch antennas on thin substrates at multi-GHz frequencies through `compute_msl_s_matrix`: the |S11| loss landscape is **not** the antenna's intrinsic |S11|, so the optimization minimizes the wrong quantity.

## Reproduction

Repository / branch / commit: rfx-ref / main / HEAD as of 2026-05-13
Hardware: NVIDIA RTX 4090, 24 GB

```python
import jax.numpy as jnp
from rfx import Simulation, Box

EPS_R = 3.38; H_SUB = 0.787e-3
W = 10.129e-3; L = 8.595e-3
W_MSL = 1.8e-3; L_MSL = 8.0e-3
PORT_MARGIN = 5.0e-3; DX = 0.197e-3
DOM_X = 29.747e-3; DOM_Y = 18.130e-3; DOM_Z = 12.787e-3
y_c = DOM_Y / 2

sim = Simulation(freq_max=15e9, domain=(DOM_X, DOM_Y, DOM_Z),
                 dx=DX, cpml_layers=8, boundary="cpml")
sim.add_material("ro4003c", eps_r=EPS_R, sigma=0.0)
sim.add(Box((0,0,4e-3), (DOM_X, DOM_Y, 4e-3+DX)), material="pec")
[...]

PR #99, opened 2026-05-29
fix(msl): assemble S from voltage-only spatial fit, not V·I split (issue #80)

bk-squared:
## Summary

Fixes the issue #80 MSL S-parameter extractor. `compute_msl_s_matrix` assembled S11/S21 from a **V·I single-plane wave split** (`a=(V+Z0_hj·I)/2`, `b=(V−Z0_hj·I)/2`, `S11=b/a`) using the analytic Hammerstad-Jensen Z0. On a discretized mesh the actual line Z0 differs from Z0_hj, so a pure forward wave gets a spurious `b` → `|S11|>0` and `|S11|²+|S21|²>1` even on a matched line, with error that grows with trace length (a thru-line must be length-invariant). This is the |q|>1 / wrong-resonance family reported in #80.

## Root cause (verified, not assumed)

Synthetic test on the real `extract_msl_nprobe` vs the V·I formula:

| case | true \|Γ\| | V·I split | spatial fit |
|---|---|---|---|
| thru (forward only) | 0 | **0.115** ✗ | 5e-5 ✓ |
| Γ=0.3 | 0.3 | **0.401** ✗ | 0.300 ✓ |
| Γ=0.85 (strong) | 0.85 | 0.879 | 0.850 ✓ |

The spatial fit recovers the true reflection to 4–5 digits in every regime, including the strong reflector the code comment claimed it "blew up" on (that was a real-FDTD probe-placement issue, not the math).

## Fix

Assemble S from the spatial fit (`extract_msl_nprobe`, fitting `V_n = α e^{−jβx_n} + γ e^{+jβx_n}`). With signed per-port probe x (`msl_probe_x_coords_n`), α is the global +x wave and γ the −x wave on every port, so a direction-keyed assembly applies:
```
leaving  = α if dir=="+x" else γ      # into the structure
arriving = γ if dir=="+x" else α      # back from the structure
S_ii = arriving_i/leaving_i ;  S_ji = arriving_j/leaving_i
```
each de-embedded probe-0→feed by `rot=exp(+jβ·d0)`, `d0=n_probe_offset·dx` (cancels in S_ii, corrects cross-term phase; magnitudes unaffected). This reads reflection purely from the standing-wave shape — independent of Z0 and the Ampere-loop current — and is passive by construction when the 2-wave fit is well-conditioned. The closed-loop current + N-probe Z0 remain diagnostics (honesty guard).

Also makes `extract_msl_nprobe` **x64-aware** (`complex128` under jax x64, else `complex64`): it previously pinned `complex64`, so once it feeds the S-matrix the AD/inverse-design "f64 path" was not actually f64. The float32 default path is unchanged.

## Validation (CPU)

- Synthetic-exact incl Γ=0.85.
- Live known-good RO4350B thru gate **passes**: mean|S11|=0.111, |S21|=0.9996, passivity 1.025.
- Length-invariant (|S11| 8mm vs 16mm differ 0.04).
- `test_replay_float64_equivalence` jnp-f64 == numpy-f64 to <1e-5 (extractor now truly f64 under x64).
- Full MSL suite green (12 synthetic + thru gate + AD smoke + AD/FD + preflight); golden fixtures re-baselined from the validated assembly (documented; the V·I-split goldens were the #80 bug — this is NOT loosening, the strict patch acceptance keeps its assertions). Capture-script numpy mirror updated to match (f32-vs-f64 delta 3.6e-5).

## Scope / honest limits

- This fixes the **extractor**. The strong-reflector **patch acceptance** (`test_issue80_patch_s11_regression`, num_periods=200) is GPU and is being validated separately on VESSL; its strict-xfail will be flipped to a green gate only if it XPASSes (assertions unchanged).
- A **separate, coupled limitation** remains on coarse/short geometries: when the probe offset < λ/(2π) the probes sit in the source reactive near-field and a spurious length-invariant |S11| enters the FIELD that no extractor can remove (issue #80 "Fix B" far-field placement). Documented in `docs/research_notes/20260529_msl_broad_e5_extractor_blocker.md`.

## Test plan

- [x] synthetic extractor (12) + AD smoke + float32-delta
- [x] live RO4350B thru gate; length-invariance
- [x] replay f64-equivalence <1e-5; E2E golden re-baselined
- [ ] GPU patch acceptance (VESSL) → flip xfail iff XPASS


bk-squared:
## GPU patch acceptance result (run 369367240214)

The strong-reflector patch acceptance (`test_issue80_patch_s11_regression`, num_periods=200) **still xfails** — but the extractor fix measurably improves it:

| | max \|S11\| | verdict |
|---|---|---|
| V·I split (before) | **1.44** | xfail (non-physical) |
| spatial-fit (this PR) | **1.166** | xfail (improved, still >1.05) |

The runtime honesty guard now pinpoints the residual cause:
> spatial-fit \|S11\| = 1.166 > 1 ... The voltage 2-wave fit is likely ill-conditioned (**probes inside the source reactive near-field**) ... Increase n_probe_offset (move probes into the far field).

**Honest scope:** this PR fixes the **extractor** (removes the Z0-mismatch corruption; validated synthetic-exact, known-good RO4350B thru gate passes, length-invariant, replay f64-equivalence, full MSL suite 37 passed / 2 xfailed). It reduces the patch non-passivity but does **not alone** resolve the resonant patch — that additionally needs the **coupled near-field probe-placement fix (issue #80 "Fix B")**: with the offset < λ/(2π) the probes sit in the source reactive near-field, so a spurious component lives in the measured FIELD that no extractor can remove.

The strict-xfail is therefore **kept** (assertions unchanged — not loosened); its reason string is updated to reflect the new state (1.44→1.166, near-field cause, Fix B remaining). When Fix B lands and the patch XPASSes, the xfail is removed and promoted to a green gate.

This PR is complete as the **extractor** fix. Fix B (far-field probe placement) is the natural follow-up to fully close issue #80.

bk-squared:
## Correction — the near-field / "Fix B" diagnosis is FALSIFIED (verify-then-fix)

My earlier comment attributed the residual patch `|S11|=1.166` to source
reactive **near-field** probe placement and proposed **Fix B** (increase
`n_probe_offset`). Before changing the shared default (which would re-baseline
every MSL golden), I verified that hypothesis with three CPU experiments. **It
is wrong**, and so is the implied fix. Pushed in `b3afadf` (xfail reason +
docstring corrected; **assertions and `strict=True` unchanged — not loosened**).

Method: one FDTD with a dense 30-probe array, dump raw probe voltages, re-fit
`|S11|` offline with the **production** extractor (`extract_msl_nprobe`) on
sliding/sub-sampled windows — isolating offset and span on identical field data.

| experiment | result | verdict |
|---|---|---|
| matched line, |S11| vs **offset** | flat 0.11–0.18 from 0.64→9.3 mm (past λ/2π); no decay | **near-field FALSIFIED** |
| matched line, |S11| vs **span** | cond 34→4 as span grows; only scatters |S11| on low-SWR | span = conditioning lever, but bounded |
| **clean open stub** (|Γ|≈1) | `max|S11| = 1.000–1.002` at **every** span & offset, even cond≈32 | **conditioning-alone FALSIFIED** |

**Conclusion:** the PR #99 voltage-only spatial fit is **robust on any clean
single-mode line at any SWR** (open stub reads 1.000). The patch's `|S11|>1` is
**neither** near-field **nor** generic conditioning — it is best explained by
**single-mode model mismatch at the short patch feed**: the reference plane sits
in the patch-junction's non-TEM higher-order/fringing field, which
`V = α e^{−jβx} + γ e^{+jβx}` cannot represent, so lstsq returns the
non-physical `|γ|>|α|`. (`num_periods=200` settling is a secondary, not-yet-
excluded alternative.)

Confirming the patch cause needs a **GPU raw-probe dump on the actual patch**
(CPU proxies are clean single-mode lines and cannot reproduce it). The honest
resolution meanwhile is to **flag** (this strict xfail + the runtime honesty
guard), not to manufacture a ≤1 number.

**Net for PR #99:** the extractor fix stands on its own and is validated; the
patch's residual is a *separate*, now-correctly-diagnosed limitation, not a flaw
in this change. Repro: `scripts/diagnostics/issue80_{nearfield_offset_sweep,
span_conditioning,highswr_span}.py`.

bk-squared:
## Definitive root cause (GPU raw-dump run 369367240331) — corrects my prior comment

My previous comment said the CPU experiments "falsified" near-field and pointed
to patch-junction model mismatch. A definitive raw-probe dump on the **actual
patch** (dense 18-probe feed array, `num_periods` 200 AND 400) shows that was an
**over-correction**. The real cause is two separable, now-proven effects — and
the near-field one is at the **SOURCE**, not the junction.

**1. Source near-field (real).** Single-mode fit residual vs probe position:

| probe x (mm) | cells from source | rel. residual @11.46 GHz | \|S11\| @11.46 GHz |
|---|---|---|---|
| 5.9 | ~5 | **0.178** | 1.003 |
| 6.7 | ~9 | **0.187** | 1.047 |
| 7.5 | ~13 | 0.028 | 0.438 |
| ≥9.1 | ≥21 | **<0.001** | 0.64–0.74 (passive) |

Residual is highest **near the MSL source** and collapses past ~13 cells. The
patch's **default `n_probe_offset` is only 4 cells** (`= 0.5·λ/2π/dx` at
f_max=15 GHz) — *inside* that transient — so near-source probes corrupt the fit
(|S11| ~5 full-array, ~1.166 at the 3-probe default). Dropping near-source probes
removes the spike; clean-region probes are passive. **The transient scales with
h_sub/w, not λ/2π**, so the default offset under-provisions at fine mesh. (The
matched-line CPU sweep saw no offset effect only because its default offset, 31
cells, already cleared its transient — that doesn't generalise to 4 cells.)

**2. Low-freq settling.** After cleaning probes, the only residual >1 is at
1.5–1.84 GHz and is settling-bound: **np200 = 1.29 → np400 = 1.01**.

➡️ **clean probes + np400 give max|S11| ≈ 1.01 ≤ 1.05 — passivity would pass.**

**Separate, still-failing (NOT extractor):** the simulated resonance dip is at
**10.78 GHz** (np200 == np400), not analytic Balanis **9.21 ± 0.2 GHz** — full-
wave-vs-analytic + Yee staircase on the patch dimensions. The dip-location
assertion fails independently of probe placement.

**Net for PR #99:** the extractor change is sound and unaffected. The patch's
residual is now fully diagnosed — a default-offset under-provisioning + settling
(both fixable) plus a separate resonance-target accuracy issue. The real fix
(floor `n_probe_offset` to clear the source transient, ~`K·h_sub/dx`) is a
broad-blast-radius default change that re-baselines MSL goldens, so it's tracked
separately, not bundled into this PR. Repro: `scripts/diagnostics/
issue80_patch_{rawdump_diag,cleanfit}.py`; details in
`20260530_issue80_nearfield_falsified.md` (ROUND 3).

bk-squared:
Closing as **superseded by #116** (merged).

#116 resolved the issue-#80 passivity defect by **flooring `n_probe_offset`** to clear the
MSL source fringing transient, keeping the existing OpenEMS-canonical V·I split. This
PR's own final diagnostic comment (GPU run `…240331`) reached the same root cause and
states the floor "is the real fix … tracked separately, not bundled into this PR" — that
separate fix is #116.

The voltage-only / `S11=γ/α` spatial-fit alternative proposed here is **ill-conditioned at
resonance** (high SWR → `|α|→0`): the design-round in #80 measured `max|S11| = 4.16→1.30`
for the voltage-only N-probe fit, vs `0.985–0.986` (passive, non-worsening with mesh) for
the floored V·I split. The simpler, verified path won, so the extractor rewrite here isn't
needed. Diagnostics preserved in the thread; reopen if a distributed extractor for
near-total reflectors (`|Γ|→1`) is pursued.

PR #100, opened 2026-05-30
[codex] P0 host-side Touchstone interop and batch provenance

bk-squared:
## Summary

P0 foundation for host-side RF interop and repeatable sweep provenance.

- Adds `TouchstoneData`, `read_touchstone_full()`, and a bounded Touchstone 2.0 S-parameter metadata subset while preserving legacy tuple APIs.
- Supports explicit standard-vs-legacy multiport layouts, per-port `[Reference]`, suffix/finite-data validation, and nonuniform-reference rejection in the legacy API.
- Adds manifest-backed sequential batch execution with stable case IDs, parameter digests, run fingerprints, artifact size/SHA validation, and resume-safe completed-case checks.
- Adds docs and tests, including passive/reciprocal network roundtrip and tiny physical FDTD sweep resume gates.

## Stack

Base: `main`.
Follow-up PRs P1-P4 are independent branches based on this P0 branch.

## Validation

- `python -m py_compile rfx/io.py rfx/batch.py rfx/__init__.py`
- `python -m pytest -q tests/test_io.py tests/test_touchstone_interop.py tests/test_batch.py tests/test_batch_provenance.py` → 48 passed
- `git diff --check`

## Notes

No FDTD solver/update-kernel paths are modified.

### DIFF
--- a/docs/guides/network_interop_batch_provenance.md
+++ b/docs/guides/network_interop_batch_provenance.md
@@ -1,0 +1,89 @@
+# Network Interop and Batch Provenance Roadmap
+
+This note records the next two bounded contribution topics after runtime artifact
+bundles. It is intentionally scoped to host-side interoperability and repeatable
+workflow infrastructure; it does not change FDTD update equations or solver
+physics paths.
+
+## 1. Touchstone / RF network interop hardening
+
+### Current rfx surface
+
+- `rfx.io.write_touchstone()` / `read_touchstone()` provide Touchstone-style
+  S-parameter I/O for RI, MA, DB formats and Hz/kHz/MHz/GHz frequency units.
+- The legacy API returns `(s_params, freqs, z0)` and writes scalar reference
+  impedance in the option line.
+- Existing tests cover round-trips and basic ordering, but the contract does not
+  yet expose structured metadata or Touchstone 2.0 keywords.
+
+### Gap vs RF ecosystem
+
+- Touchstone 2.0 formalizes metadata such as `[Number of Ports]`,
+  `[Number of Frequencies]`, `[Reference]`, `[Matrix Format]`, and explicit
+  two-port data ordering.
+- For three or more ports, the standard arranges each frequency block in
+  matrix row-wise order. rfx must preserve any existing legacy behavior while
+  adding an explicit standard path.
+- RF network tooling such as scikit-rf exposes richer impedance/write controls;
+  rfx should at least preserve impedance metadata and fail clearly on unsupported
+  mixed-mode or non-S-parameter variants.
+
+### Proposed bounded contribution
+
+- Add a metadata-aware Touchstone reader/writer path while preserving the legacy
+  tuple API.
+- Keep existing v1 calls on the legacy rfx multi-port layout unless callers opt
+  into `layout="standard"` or `version="2.0"`.
+- Support standard 3+ port row-wise layout and retain a named legacy layout for
+  backwards compatibility.
+- Parse/write useful Touchstone 2.0 metadata: version, number of ports,
+  frequency count, matrix format, two-port order, and per-port `[Reference]` for
+  S-parameters. Legacy tuple reads must raise on nonuniform references instead
+  of silently collapsing them.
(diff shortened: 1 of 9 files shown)


bk-squared:
## Review (deep-research, adversarially verified) — 2 CONFIRMED HIGH Touchstone bugs before merge

Batch provenance/resume is merge-ready and robust. The Touchstone **multiport layout** has two confirmed-real bugs, both masked by round-trip-only tests that exercise only N=2 / N=4:

### HIGH-1 (read): external v1 3+ port files read column-major → off-diagonals transposed
`rfx/io.py:566-567` resolves `layout='auto' → 'legacy-rfx'` for any file lacking v2 `[...]` keywords. Real ADS/CST/HFSS/scikit-rf `.s3p/.s4p` are **row-major** per Touchstone v1 spec. Repro: a row-major `.s3p` with S12=12/S21=21 reads back as `s_params[0,1]=21, [1,0]=12`. Reciprocal nets survive (S=Sᵀ); **couplers/circulators/isolators silently corrupt.** 2-port unaffected.

⚠️ **This is a format-POLICY decision, not a one-liner:** rfx also *writes* `legacy-rfx` (column-major) by default for v1, so naively flipping the read default to row-major would transpose rfx's own round-tripped files. Pick one:
- (a) adopt spec row-major as the v1 default for both read+write (spec-correct; breaks compat with previously-written rfx v1 files), or
- (b) keep `legacy-rfx` internal default but auto-detect / require `layout='standard'` for external reads (documented prominently).

Note the v1 special case when fixing: **2-port v1 is column-major** (S11,S21,S12,S22); **N≥3 is row-major** — don't regress the 2-port path.

### HIGH-2 (write): standard-layout writer breaks one-row-per-line for N≠4
Hard-coded `_MAX_PAIRS_PER_LINE=4` chunking is applied in `standard` layout regardless of N. IBIS Touchstone 2.1 (Examples 6/7) mandates one matrix **row** per physical line. A 3-port emits `freq S11 S12 S13 S21` on line 1 (mixes rows). Self round-trip passes only because the reader re-chunks identically; **4-port coincidentally aligns** (the only N tested).

### Required gates (currently missing)
- parse a hand-written **row-major external `.s3p`** with distinct off-diagonals; assert S[0,1]/S[1,0] map correctly
- a 3-port v2-standard external read
- a 3-port standard **write** asserting one matrix row per line

*Verified NOT real (skip):* the bare-`R` Z0=50 default concern — non-numeric `R` already raises; pre-existing on main.

#101–104 stack on this branch, so this is the gate. Round-trip API itself is internally consistent; fixes are localized to multiport layout.

bk-squared:
## Merge runbook (codex P0–P4 interop/provenance stack)

This stack was rebased onto current `main` (`0b8daf3`, includes #108). All five PRs are MERGEABLE/CLEAN and `public-docs-source` is green. Three public-guide docs were re-resolved against main's canonical #105 versions (the earlier rebase had duplicated `probes-sparams.mdx` and dropped the `Aside` import — both fixed; only valid APIs are referenced).

**P1–P4 are siblings off P0** (each = P0 + 1 commit), intentionally not linearized. Merge **P0 first**, then P1–P4 in any order, with two trivial, expected merge-time conflicts for the *second* PR of each pair:

- **P1 ↔ P2** — `tests/test_touchstone_interop.py` import block: resolve by union of imports (`TouchstoneData, network_quality_metrics, read_touchstone, read_touchstone_full, write_touchstone`).
- **P3 ↔ P4** — `rfx/batch.py` `run_batch_with_manifest` exception handler + `tests/test_batch_provenance.py` + the shared guide: merged handler must keep both behaviours, i.e. `manifest['last_run'].update({...'failed'...}); _write_manifest_atomic(...); if not continue_on_error: raise` (verify failed-case provenance is still recorded when `continue_on_error=True`).

P1↔P3, P1↔P4, P2↔P3, P2↔P4 merge cleanly. If you'd prefer these linearized (P0→P1→P2→P3→P4) to remove the merge-time conflicts entirely, that's a quick follow-up.

PR #101, opened 2026-05-30
[codex] P1 preserve Touchstone information and D exponents

bk-squared:
## Summary

P1 hardens Touchstone metadata compatibility for files emitted by external RF tools.

- Preserves `[Begin Information]` / `[End Information]` blocks as raw lines in `TouchstoneData.information`.
- Adds Touchstone 2.0 writer support for `information={...}` or information line sequences.
- Accepts Fortran-style `D` exponent numeric tokens in option, reference, and network-data fields.
- Adds roundtrip and invalid-version tests.

## Stack

Base: `codex/p0-network-interop-batch-provenance`.
Independent of P2/P3/P4.

## Validation

- `python -m py_compile rfx/io.py rfx/__init__.py`
- `python -m pytest -q tests/test_io.py tests/test_touchstone_interop.py` → 33 passed
- `git diff --check`

### DIFF
--- a/docs/guides/network_interop_batch_provenance.md
+++ b/docs/guides/network_interop_batch_provenance.md
@@ -40,6 +40,9 @@ physics paths.
   frequency count, matrix format, two-port order, and per-port `[Reference]` for
   S-parameters. Legacy tuple reads must raise on nonuniform references instead
   of silently collapsing them.
+- Preserve bounded `[Begin Information]` / `[End Information]` metadata blocks
+  as raw lines so external tool provenance can survive rfx import/export.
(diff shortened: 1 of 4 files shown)


bk-squared:
## Review — 1 confirmed bug

**Unterminated `[Begin Information]` silently swallows the rest of the file** (incl. `[Network Data]`) and returns a degenerate `(2,2,0)` object with no error — reachable for external files. One-line guard after the read loop: `if in_information: raise ValueError("unterminated [Begin Information] block")`. Add a test for it, and for the orphan `[End Information]` raise path (currently untested).

D-exponent / info-preservation logic itself checks out. Low-severity polish (sorted-key write order, comment capture) is optional.

PR #102, opened 2026-05-30
[codex] P2 add S-matrix quality metrics

bk-squared:
## Summary

P2 adds host-side S-matrix report diagnostics for interop/visualization workflows.

- Adds `network_quality_metrics()` for finite-data, max magnitude, max column power, max singular value, passivity excess, and reciprocity error.
- Exports the helper from `rfx.__init__`.
- Documents report-oriented use and explicitly keeps physics-claim validation tied to support-matrix evidence.
- Adds passive/reciprocal, nonreciprocal, and active-network tests.

## Stack

Base: `codex/p0-network-interop-batch-provenance`.
Independent of P1/P3/P4.

## Validation

- `python -m py_compile rfx/io.py rfx/__init__.py`
- `python -m pytest -q tests/test_io.py tests/test_touchstone_interop.py` → 31 passed
- `git diff --check`

### DIFF
--- a/docs/guides/network_interop_batch_provenance.md
+++ b/docs/guides/network_interop_batch_provenance.md
@@ -42,6 +42,8 @@ physics paths.
   of silently collapsing them.
 - Validate writer inputs against the file suffix and finite numeric data so rfx
   does not emit self-inconsistent `.sNp` files.
+- Provide `network_quality_metrics()` for host-side passivity, reciprocity,
+  finite-data, and magnitude diagnostics that can be recorded in reports.
 - Add physical/interop gates based on passive/reciprocal known networks, not only
   random numeric round-trips.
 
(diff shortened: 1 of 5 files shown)


bk-squared:
## Review — math is sound (σ_max passivity, not weak column-power); 2 items

1. **Empty cube (n,n,0) crashes** at the `reciprocity_error` line (`rfx/io.py:651`) with `ValueError: zero-size array to reduction`. `max_abs_s` is guarded `if s.size else 0.0`; reciprocity is not — add the same guard or reject `n_freqs==0`.
2. **Passivity definition diverges from `rfx/validation.py`** (σ_max vs column-power; absolute vs relative tol). They disagree on real matrices, e.g. `[[0.9,0.9],[0,0]]` (passive by validation.py, non-passive here). Not a bug in new code, but add a docstring note / reconcile follow-up. *Should-do, not merge-blocking.*

(The red `public-docs-source` check this PR surfaced was fixed by #105, now merged.)

PR #84, opened 2026-05-20
fix(api): refactor cleanup + Kottke port dilation guard (#82, #83)

bk-squared:
## Summary
- Fix _assemble_materials type hint/docstring (7-tuple, not 6) — #83
- Update stale line-number references from pre-refactor api.py — #83
- Clear Kottke dilation neighbors around port cells — #82

## Kottke port dilation guard

Kottke occ_dilated = max(f, roll(f,+/-1,...)) picks up non-zero occupancy from cells adjacent to each port. For probe-fed patches (port 1 cell below PEC sheet), this leaks the patch occupancy gradient into the port inv_eps tensor.

Before fix: 60x-223% AD-vs-FD gradient mismatch on |S11|^2 objectives
After fix: Port cell + 6 face-neighbors zeroed in pec_occupancy_local before kottke_inv_eps_from_occupancy call

Closes #82, closes #83

PR #107, opened 2026-05-31
fix(runners): land 6-vs-7-tuple _assemble_materials unpack fix on main

bk-squared:
## Why

The scheduled `scientific-validation` workflow on `main` has failed on every run (2026-05-11, -18, -25):

```
FAILED tests/test_boundary_pmc_composition.py::test_oq9_distributed_v2_cpml_path_enforces_pec_face_via_cpml_init
  ValueError: too many values to unpack (expected 6)
  rfx/runners/distributed_v2.py:615
```

`Simulation._assemble_materials()` has returned a **7-tuple** (`…, pec_shapes, boundary_pec_shapes, kerr_chi3`) since `b7683f2` (2026-05-17), but several runner / sweep call sites still unpacked **6** targets and raise on any distributed / vmap-sweep path. These paths are latent under the default CPU gate, so the break only surfaced in the full scheduled suite.

The cluster was already repaired in commit **`73c3f3b`** ("fix(runners): repair 6-vs-7-tuple _assemble_materials unpack cluster + numpy-2 trapz"), but that commit only ever lived on the issue-#80 branches and was **never merged to main**. This PR lands it on main, decoupled from issue #80.

## What

- **`2efc83b`** — cherry-pick of `73c3f3b`: fixes `distributed.py`, `distributed_v2.py`, `topology.py`, `differentiable_material_fit.py`, `tests/test_topology.py`, `tests/test_thin_conductor.py` (trailing `*_`), plus the numpy-2 `np.trapezoid` guard in `coaxial_port.py`. The `_compile.py` docstring conflict was resolved in favour of main (no diff there).
- **`7638c4a`** — completes the cluster: `73c3f3b` itself **missed** `rfx/vmap_sweep.py:553` (also 6 targets, same latent crash). Fixed with the same trailing `*_`.

## Verification

- **Exhaustive audit**: every `_assemble_materials()` unpack site in `rfx/` and `tests/` now uses 7 targets or `*_`. No 6-target sites remain in the package/test surface.
- `python -m py_compile` passes on all 7 changed files.
- JAX is not installable in the authoring env, so the distributed test itself runs in CI/VESSL — the fix is a pure unpack-arity correction matching the exact CI error.

## Out of scope

Seven one-off diagnostic scripts under `scripts/` still carry the latent 6-target unpack (`scripts/_wr90_port_oracle_matrix.py`, `scripts/isolate_extractor_vs_engine.py`, etc.). They are not part of the package or CI and were left untouched.

## Falsifier

The named test `test_oq9_distributed_v2_cpml_path_enforces_pec_face_via_cpml_init` going green on a main that includes this branch.

### DIFF
--- a/rfx/differentiable_material_fit.py
+++ b/rfx/differentiable_material_fit.py
@@ -391,7 +391,7 @@ def differentiable_material_fit(
         sim = sim_factory(eps_inf, debye_poles, lorentz_poles)
 
         # Assemble materials (geometry masks are static, but eps_inf flows through)
-        materials, debye_spec, lorentz_spec, pec_mask, _, _ = sim._assemble_materials(grid)
+        materials, debye_spec, lorentz_spec, pec_mask, *_ = sim._assemble_materials(grid)
 
         # Setup ports (fold port impedance into materials)
         for pe in sim._ports:
--- a/rfx/runners/distributed.py
+++ b/rfx/runners/distributed.py
@@ -1273,7 +1273,7 @@ def run_distributed(sim, *, n_steps, devices=None, exchange_interval=1,
 
     # Build grid and materials (full domain)
     grid = sim._build_grid()
-    base_materials, debye_spec, lorentz_spec, pec_mask, pec_shapes, _ = (
+    base_materials, debye_spec, lorentz_spec, pec_mask, pec_shapes, *_ = (
         sim._assemble_materials(grid)
     )
     materials = base_materials
--- a/rfx/runners/distributed_v2.py
+++ b/rfx/runners/distributed_v2.py
@@ -611,7 +611,7 @@ def run_distributed(sim, *, n_steps, devices=None, exchange_interval=1,
         pec_shapes = None
     else:
         grid = sim._build_grid()
-        base_materials, debye_spec, lorentz_spec, pec_mask, pec_shapes, _ = (
+        base_materials, debye_spec, lorentz_spec, pec_mask, pec_shapes, *_ = (
             sim._assemble_materials(grid)
         )
     materials = base_materials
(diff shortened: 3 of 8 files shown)


bk-squared:
## CI verification (manual dispatch on this branch)

Dispatched `validation.yml` on `fix/distributed-tuple-unpack-main` (run [26717382391](https://github.com/bk-squared/rfx/actions/runs/26717382391)).

**This PR's fix is verified:** the original failure
`test_oq9_distributed_v2_cpml_path_enforces_pec_face_via_cpml_init`
(`ValueError: too many values to unpack (expected 6)` @ `distributed_v2.py:615`)
is **gone** — the error string no longer appears anywhere in the suite log, and
465 tests pass.

**However, `main` is not yet green** — a *separate, pre-existing* failure remains,
unrelated to this PR:

```
FAILED tests/test_kottke_inv_eps_from_occupancy.py::test_half_fill_y_normal_kottke_signature
  AssertionError: boundary inv_yy (perp) should be 0.25 (with Heaviside proj); got mean 0.0000
```

Root cause is in the Kottke subpixel path (`rfx/geometry/smoothing.py`), introduced
by the **#84 Kottke dilation guard** (commit `60939e0`, "1-cell PEC dilation via
neighbor-max"): the boundary cell at `occ=0.5` adjacent to a PEC cell is now
neighbor-max-dilated to full PEC (`inv_yy → 0`), while the test still expects the
pre-dilation subpixel value `0.25`. This is a test-vs-code consistency issue that
landed via #84 (2026-05-28), after the last scheduled run (2026-05-25) — so it had
not been surfaced until now. **Out of scope for this PR; tracked separately.**

This PR is still correct and necessary (it removes one of the two red tests); it is
just not sufficient on its own to green the scheduled CI.

Issue #61, opened 2026-04-16
preflight: detect geometry/material placed inside CPML region


bk-squared:
## Problem

When a user places geometry (material boxes, PEC structures) inside the CPML absorbing boundary region, the simulation runs without warning but produces physically meaningless results. CPML modifies the field update equations with absorbing coefficients — any dielectric or PEC placed there is effectively "eaten" by the absorber.

## Discovery Context

rfx-TAP paper experiments (ex1 DRA, ex6 absorber) both had this bug:
- **ex1**: PEC ground plane at z=0, DRA at z=0..10mm, but CPML occupies z=0..8mm → DRA resonance at wrong frequency (7.4 GHz vs 10 GHz target), HPBW=152° (near-isotropic instead of directional)
- **ex6**: PEC backing plane at z=0..1mm inside 8mm CPML → no proper reflection → NaN gradients during optimization

Both ran silently with no warnings. The only clue was a `UserWarning: Material extends into CPML region` on the post-optimization `run()` call — but this warning came from the NEW preflight code in PR #34, and only for `run()`, not during `optimize()`/`forward()`.

## Proposed Check

In `preflight()` and/or at `Simulation.add()` time:

```python
cpml_z_lo = cpml_layers * dx  # bottom CPML boundary
cpml_z_hi = domain_z - cpml_layers * dx  # top CPML boundary
# (same for x, y)

for geo in self._geometry:
    if geo.box.lo[2] < cpml_z_lo or geo.box.hi[2] > cpml_z_hi:
        warnings.warn(
            f"Geometry '{geo.material}' at z=[{geo.box.lo[2]*1e3:.1f}, {geo.box.hi[2]*1e3:.1f}]mm "
            f"overlaps CPML region z=[0, {cpml_z_lo*1e3:.1f}]mm. "
            f"CPML absorbs the geometry — results will be incorrect. "
            f"Shift geometry above z={cpml_z_lo*1e3:.1f}mm.",
            UserWarning, stacklevel=2,
        )
```

Should fire on ALL paths: `run()`, `forward()`, `optimize()`, `topology_optimize()`.

## Severity

HIGH — this silently produces wrong results that look plausible. Cost us ~4 hours of debugging + multiple VESSL GPU-hours.

PR #119, opened 2026-06-02
fix: green the slow-test suite — preflight CPML carve-out + 6 stale/brittle test re-pins

bk-squared:
## Summary

CI's `slow-tests` job runs `pytest -x`, which stops at the first failure
(`test_lossy`) and **masked six more failures behind it**. Running the full suite
without `-x` (clean `origin/main` worktree, `-n 24`) and re-running each failure
**serially** to drop xdist parallelism artifacts surfaced **7 real deterministic
failures** — none from the recent issue-80/Kottke merges; all pre-existing and
hidden by `-x`. Each is root-caused and fixed against the underlying physics/state,
not by loosening a gate.

| # | test | class | fix |
|---|------|-------|-----|
| 1 | `test_periodic_cpml::…non_periodic_z_axis` | preflight regression | CPML-intrusion exemption was too broad (a slab buried in the absorber also "touches the edge"); now intentional only if it reaches past the CPML into the interior, with a degenerate full-CPML carve-out so the thin-substrate false-positive stays silent |
| 2 | `…rf_e5_blocker_ladder` | committed progress | re-pin 7→ exact 4-family set (waveguide/coaxial closed their broad-E5 envelopes) |
| 3 | `test_lossy::higher_sigma_faster_decay` | measurement artifact | same-time ½εE²+½μH² energy oscillates at 2ω; measure conservation tail-averaged + σ-ordering early (also fixes a latent floor-equality bug) |
| 4 | `…external_reference_audit` | test-design bug | required gitignored `.omx/` VESSL artifacts; now asserts the real clean-checkout contract (gate correctly blocked) |
| 5 | `…shard_report_incomplete_family` | committed progress | coaxial status advanced to `pec_short_calibrated_…_pending` |
| 6 | `…shard_execution_manifest` | test-design bug | skipped gitignored `**/vessl*.yaml` instead of unconditionally reading them |
| 7 | `test_v173a_baseline_bit_identity` | numerics-refactor drift | rebump baseline f_res 1.6923→1.9409 GHz (deterministic; old mode gone, not an amplitude flip) — caused by intentional merged refactors (77daeb8 inverse-eps E-update; EPS_0/MU_0 post-2019 SI) |

## Verification

- Full suite after fix (`-n 24`): **1319 passed**, only the pre-existing
  `test_sparam_ad_end_to_end::test_msl_s_matrix_ad_end_to_end` **xdist parallelism
  artifact** remains red — it **passes serially** (214 s). It is not touched by this
  PR and is unrelated to issue-80/#116 (the `n_probe_offset` floor does not change
  that geometry's offset).
- The two conflicting preflight tests (full-domain-silent vs buried-slab-warn vs
  inset-leak-warn) all pass together.

## Reviewer notes (please scrutinize)

- **#7 (v173a rebump)** is the one change that *blesses current numerics* as the new
  reference. Evidence: bit-identical over 2 trials, the 1.69 GHz mode is gone (genuine
  resonance move, not a harminv amplitude flip), uniform mesh rules out the NU
  Yee-spacing fix, and every intervening commit is an intentional merged change. It
  is an equivalence/bit-identity pin, **not** a correctness gate — 1.9409 GHz still
  deserves an independent analytic sanity-check. Easy to revert if undesired.
- **#4 / #6** now assert the clean-checkout "blocked" contract because the broad-E5
  evidence artifacts are VESSL outputs that repo policy keeps gitignored. No artifacts
  were fabricated.


bk-squared:
## Independent adversarial review — passed (with one provenance correction)

Ran 5 independent reviewers (each forbidden from "make red→green by masking"), all
verifying with real commands (pytest, `git log -S`, live `build_external_reference_audit`,
harminv mode dumps, and a **git-worktree bisect** for v173a). Result: **all 5 sound /
high confidence; 0 masked bugs; 0 loosened gates.**

- **preflight**: the old code was the *actual* defect (silent on the issue-#61 footgun); the fix tightens, not loosens. 62 preflight tests pass; new rule re-derived by hand + 8 synthetic geometries (incl. PEC-ground microstrip, asymmetric lo-CPML/hi-PEC) all correct.
- **lossy**: independent leak probe shows the lossless envelope is flat (0.0037% drift over 500 steps → no leak); σ-ordering margins ×4.1; base commit is genuinely red. The two-measurement split is necessary (late-time both lossy cases floor to bit-identical 9e-4).
- **physics-gate**: all 33 audit "missing" artifacts are confirmed gitignored (29 `.omx/`, 4 `vessl*.yaml`); **zero** committed JSON is schema-invalid; the 4-family set + coaxial status are data-derived from committed JSON and code-backed by a passing test; gate keeps teeth (`status=blocked`, `rc=1`).
- **v173a**: bisect confirms the rebump is justified **and a correctness improvement** — analytic TM010 cavity ≈ 2.0 GHz, so new **1.94 GHz is ~4% low** vs the old 1.69's **~16% CPML-bleed deficit**.

**Correction applied (amended):** the bisect showed my original cause attribution was
wrong. The shift is **not** 77daeb8 (its inverse-eps path needs
`subpixel_smoothing='kottke_pec'`, unused here) nor EPS_0/MU_0 (zero movement). The real
cause is the v1.7.5 per-face CPML/PMC-padding fix **e340644** (1.6923→1.8895, which removes
spurious CPML absorption of the patch-cavity modal field) + **c8b17f8** (→1.9409). The
baseline `_note` and commit message now reflect this. Also removed two stray scratch logs
from the working tree.

### Non-blocking follow-ups (filed for later, not in this PR)
1. Analytically sanity-check the new 1.9409 GHz v173a baseline before the next release tag (the bisect's TM010 ≈ 2.0 GHz already supports it).
2. Optional preflight defense-in-depth: warn when summed CPML thickness exceeds the domain on an axis (degenerate over-thick-CPML config) — predates this PR.
3. `test_lossy` `_EARLY_STEP=30` is tied to the current dt/dx; an in-window guard would fail loudly if a future numerics change shifts the valid window.

bk-squared:
## Follow-up ① closed: v173a 1.9409 GHz analytically validated

Balanis cavity model (TM010, resonant along the 36 mm length, W=28 mm, FR4 εr=4.3,
h=1.5 mm, Hammerstad fringing): **f₀₁₀ = 2.021 GHz** (eps_eff=3.94, L_eff=37.4 mm).

| quantity | value | vs analytic |
|---|---|---|
| analytic TM010 | 2.021 GHz | — |
| **FDTD dominant (new baseline)** | **1.9409 GHz** | **−3.9%** ✅ |
| FDTD weak mode | 3.81 GHz | −5.7% vs ~TM020 (4.04 GHz) |
| old baseline | 1.6923 GHz | −16.3% ❌ (CPML-bleed) |

Both FDTD modes sit a *consistent* ~4–6% below their analytic counterparts — exactly the
uniform low-bias expected from 1 mm-cell staircasing — so the mode structure is physically
sound and the **rebump to 1.9409 GHz is correct** (the old 1.69 GHz was the 16% CPML-bleed
deficit, now fixed by e340644). The "wants an analytic sanity-check" caveat is resolved.

PR #134, opened 2026-06-09
fix(msl): re-land the voltage-only spatial-fit extractor (issue #80) — main still used the non-passive V·I split

bk-squared:
## Summary

`compute_msl_s_matrix` on main assembles the MSL S-matrix from a **V·I single-plane wave split** (`a=(V+Z0_hj·I)/2`, `b=(V−Z0_hj·I)/2`, `S11=b/a`) using the analytic Hammerstad–Jensen `Z0`. On a discretized mesh the actual line `Z0` differs (52–79 Ω vs 50), so a pure forward wave on a **uniform matched thru-line** picks up a spurious `b` ⇒ `|S11|>0` and `|S11|²+|S21|²>1`, with the error **accumulating along the line**. A thru-line is length-invariant by physics, so a length-dependent extractor is structurally unsound.

This re-lands the extractor fix originally written as **PR #99** (commit `f92410a`), which was closed unmerged — only its `n_probe_offset` floor reached main (as #116), leaving the core extractor still on the V·I split.

## Fix

Assemble S from the **voltage-only spatial fit** already produced by `extract_msl_nprobe` (`V_n = α e^{−jβx_n} + γ e^{+jβx_n}`). Direction-keyed: `leaving = α if direction=="+x" else γ`, `arriving` the opposite, `S_ii = γ/α` at the de-embedded reference plane. This reads reflection from the standing-wave shape alone — **independent of `Z0` and the Ampère-loop current** — and is passive by construction when the 2-wave fit is well-conditioned. `extract_msl_nprobe` is also made x64-aware so the AD / inverse-design f64 path is genuinely f64. Goldens re-baselined from the validated spatial-fit assembly; the closed-loop current and N-probe `Z0` are retained as soft diagnostics (honesty guard).

## Validation (CPU)

| uniform 50 Ω thru-line (dx = h_sub/4, 3–4.5 GHz) | before (V·I split) | after (spatial fit) |
|---|---|---|
| 8 mm max passivity \|S11\|²+\|S21\|² | 1.072 | 1.045 |
| 8 mm mean \|S11\| | 0.233 | 0.183 |
| 8 mm extracted Z0 | **−41.6 Ω** | +41.4 Ω |
| length dependence \|Δ max_pass\| (8 ↔ 16 mm) | large (1.30 → 4.74 over 8 → 37 mm) | **0.006** |

- `test_msl_port_integration.py::test_msl_thru_line_passive_gate` (real FDTD, dx = 80 µm recipe) — **PASS**
- `test_msl_sparam_ad.py::test_compute_msl_s_matrix_end_to_end_matches_historical_base` (golden replay) — **PASS**
- fast MSL suite — **44 passed / 1 xfail** (eigenmode J+M source, unrelated)
- new `tests/test_msl_length_invariance.py` (slow) — **PASS** — a uniform thru-line at two lengths must give length-invariant passivity & `|S11|`; guards regression to the V·I split.

## Known follow-up (out of scope here)

The residual matched-line `|S11| ≈ 0.12` at the recipe mesh is a **separate** source near-field / Yee-staircase floor (itself length-invariant), to be addressed in the broad-E5 envelope work (far-field probe placement / finer mesh). This PR fixes the **extractor**; it does not by itself flip `microstrip_line_port` to `broad_e5_passed`.


bk-squared:
Closing after an evidence review against the issue-#80 record (decision 2026-06-11).

**Why closed, not merged:**

1. **This re-lands PR #99**, which was closed as superseded by #116 (`n_probe_offset` floor). The superseding decision was made *with* the spatial-fit data in hand, not in ignorance of it.
2. **The spatial-fit (γ/α) extractor is documented ill-conditioned at high-SWR resonant loads.** Curve-level crossval (2026-05-31, dx sweep 300→98µm on the issue-80 patch): at the patch resonance the forward-wave amplitude |α| collapses (~17× at 9.30 GHz, dx=98µm) and γ/α blows up to **|S11| = 6.4 (non-passive)** — and *mesh refinement makes it worse* (higher Q → higher SWR → smaller α). The V·I split with the #116 offset floor is passive on the same device at every tested mesh (max|S11| 0.994 @197µm, 0.986 @98µm, non-worsening — VESSL run 369367240504).
3. **All validation in this PR is on a matched thru-line** — the low-SWR regime where the 2-wave fit is well-conditioned. There is no resonant-device evidence. The PR's own caveat ("passive by construction *when the 2-wave fit is well-conditioned*") is exactly the condition that fails on the device issue #80 is about.
4. **Both golden fixtures are re-baselined to the new extractor**, so the replay gates can no longer catch the regression this PR would introduce. The patch passivity gate (`test_issue80_patch_s11_regression.py`) is gpu+slow and does not run in PR CI, so a merge would land unguarded.
5. Branch is 9 commits behind main and predates the #138 `_sparams.py` changes.

**What survives from this PR:**

- The thru-line findings (claimed length-dependent passivity error 1.30→4.74 over 8→37mm under the V·I split; extracted Z0 = −41.6 Ω) are a real, *separate* claim about the matched-line regime — tracked as a new issue with these numbers quoted as claimed-but-unverified.
- The `extract_msl_nprobe` x64-awareness fix and `tests/test_msl_length_invariance.py` are salvageable pieces, noted in that issue.
- Branch backed up locally (git bundle) before deletion.

**Reopen condition:** a committed, non-stub green run of the issue-80 resonant patch under the spatial-fit assembly at **both** the recipe mesh (197µm) **and** ≤98µm showing max|S11| ≤ 1.05, with per-frequency dump. Without that, swapping extractors trades a claimed thru-line defect for a documented resonance blow-up.

Issue #332, opened 2026-07-12
preflight/post-run: energy-based truncation witness for open-domain fixed-num_periods runs (G1)


bk-squared:
Under-settled runs manufactured spurious ~3.1 GHz 'resonances' twice in the 2026-07-11/12 arc (np=60/200mm domain; np=120/334mm domain — required settling scales with domain size and Q), nearly flipping a correct STOP decision. Design VERIFIED (synthetic battery in the 2026-07-12 known-issues settling entry): two-window harminv comparison is two-sided broken (dominant-identity false-fires on settled multi-Q amplitude-ranking crossovers; matched-mode false-passes transient mush) — do NOT implement that. CORRECT check: for open(CPML) + pulsed-source + fixed-num_periods runs, report end-of-run total energy (or probe envelope) vs post-source peak; below −40 dB → advisory 'truncation: Harminv/DFT/NTFF quantities may carry transient error; consider until_decay' (openEMS EndCriteria pattern; reuses #169/#176 total-domain-energy machinery). Scope OUT: PEC-closed domains (energy conserved by design), CW drives. Always states a measured fact. Field-validated: the witness ran in the patch arc scripts (−55 dB OK). Advisory only + test + docs sync.

bk-squared:
Shipped in PR #346 (merged): post-run advisory fires when an open-boundary pulsed fixed-num_periods run ends above −40 dB of peak, recommending until_decay. Demonstrated live in examples/tutorials/run_control_and_fields.py (7-period run trips it; until_decay clears it).

PR #377, opened 2026-07-15
fix(cv05): register substrate to smoothed grid (6 cells/1.5mm) + far-field mode-ID; supersede keep-buggy note (#325)

### LINKED ISSUE #325: cv05 z-mesh: smooth_grading without preserve_regions mis-registers the ground/substrate/patch stack (substrate built 2 cells/1.361mm, not 6/1.5mm)
## Summary

`examples/crossval/05_patch_antenna.py:199` builds the non-uniform z-mesh with
`smooth_grading(raw_dz, max_ratio=1.3)` and **no `preserve_regions`**. Because
`smooth_grading` inserts transition cells before any >1.3× ratio jump, the fine
substrate/ground/patch cells are silently shoved upward and the stack is
mis-registered. This is the exact failure mode issue #48 fixed via
`preserve_regions` (regression test `tests/test_smooth_grading_preserve.py`,
whose docstring geometry is essentially cv05's) — cv05 never adopted the fix.

## Verified (independent reproduction — grid build only, no FDTD run)

Building cv05's exact `raw_dz` (air_below=15mm, 1 ground cell + 6 substrate cells
of 0.25mm + air_above=25mm, dx=1mm) through `smooth_grading(max_ratio=1.3)`:

| Feature | Intended | Actually built |
|---|---|---|
| substrate | 6 cells × 0.25mm = 1.500mm at z=[15.00, 16.50]mm | **2 cells, 1.361mm** (9.3% thin) |
| the 7 fine 0.25mm cells | ground + substrate stack | shoved to z=17.44–18.94mm (in the air ~2.7mm ABOVE the patch) |

raw 47 cells → smoothed 57 (10 transition cells inserted; ~5 land before the
fine block and push it up). Both the Harminv (source-excited) and port lanes
share this `dz_profile`, so both are affected — it is upstream of either.

## Why it matters (R5)

The primary gate (Harminv-vs-openEMS resonance 2.65%) currently PASSES despite
this, so the metric hides a wrong geometry: the FR4 substrate the script
documents as "6 cells / 1.5mm" is physically 2 cells / 1.361mm and severely
under-resolved. Whether the current openEMS agreement survives the corrected
1.5mm/6-cell substrate is itself a useful check (the fix's before/after
resonance shift + gate margin must be dumped, not assumed).

Not the #118/#127 "wrong observable" class — a plain geometry-registration bug
with an existing library fix, just never applied to this example.

## Fix

Add `preserve_regions` (the fine substrate+ground+patch z-band) to the
`smooth_grading` call, mirroring the #48 pattern. Before/after witness: substrate
rebuilds as 6 cells / 1.5mm; resonance shift and the Harminv-vs-openEMS gate
margin recorded. Also worth checking in the same pass: `06_msl_notch_filter.py`
uses the same unprotected `smooth_grading` for its xy fine-region (MSL/stub)
placement — its z-profile is immune by construction (fine block is first) but the
xy calls are unverified.

_Found by the cv05 settings audit (2026-07-11); reproduction is grid-build only._

bk-squared:
Resolves #325. Two coupled fixes that **supersede** the cv05 docstring's "keep the mis-registered substrate" decision, with new evidence. **VESSL openEMS gate confirmation pending** (run `369367247281`).

## 1. Geometry (#325)
`smooth_grading` inserts transition cells below the fine 0.25mm stack (ramping the coarse air down within max_ratio), shifting the fine block **up** in absolute z. The geometry Boxes were placed at fixed pre-smoothing z, so the substrate Box landed on coarse transition cells → built as **2 cells / 1.361 mm** instead of 6 / 1.5 mm. Fix: `preserve_regions` + **register the ground/substrate/patch stack to where the fine cells actually land** (asserted: substrate now 6 cells / 1.500 mm).

## 2. Resonance extraction — far-field broadside mode-ID
The harminv resonance was picked by **closest-to-analytic** — mode-ambiguous (the patch is multi-mode; the selector flips between modes as geometry/subpixel shift, e.g. 2.65 ↔ 2.23 GHz). Fix: add an NTFF box + select the **radiating** mode by far-field broadside (|E/H peak| ≤ 20°, highest directivity), analytic-independent. Verified **stable across subpixel**: it picks the same physical mode off and on (2.6472 → 2.7539 GHz — a shift, not a swap), where closest-to-analytic would have swapped.

## Why this supersedes the "kept as-is" docstring note (R1 lifecycle, new evidence)
That note rejected re-pinning because it "degraded resonance agreement" — but the agreement was harminv-vs-**analytic** under the ambiguous selector, and the analytic is 5–8 % approximate (**not the gate**). The **primary gate is rfx-vs-OpenEMS < 20 %**. The corrected 6-cell substrate reads ~9 % above analytic (the coarse-mesh high-bias of the radiating mode, physically expected), and rfx's radiating mode (2.6472 GHz off) vs OpenEMS (~2.42 GHz) is ~9.5 % — **to be CONFIRMED on VESSL** (openEMS absent locally → cv05 exits 2/SKIP). Docstring updated to match.

## Verification
- Local (both subpixel off/on): `[#325] substrate z=[14.686,16.186]mm = 6 cells/1.500mm`; far-field picks the same broadside mode (off 2.6472 GHz D=6.68 dBi, on 2.7539 GHz D=6.83 dBi). ruff clean.
- **VESSL** (`369367247281`, remilab-c0, openEMS source build): the decisive rfx-vs-OpenEMS < 20 % gate on the corrected geometry — **result pending** (~hours). Merge gated on it.
- Out of scope (pre-existing #332): num_periods=60 undersettles to ~−25 dB; applies equally before/after.

R3: memory=supersedes the cv05 docstring "keep buggy" note with new evidence (far-field mode-ID removes the ambiguity that drove the rejection; openEMS is the real gate) | R2-attempts=1 | falsifier=rfx-vs-OpenEMS < 20 % on VESSL 369367247281.

### DIFF
--- a/validation/crossval/05_patch_antenna.py
+++ b/validation/crossval/05_patch_antenna.py
@@ -12,11 +12,20 @@ Structure (stack from bottom to top):
     ``pec_faces={"z_lo"}`` — an infinite boundary GP turns the antenna
     into a cavity and shifts the resonance ~8 % high (see build_patch
     docstring).
-  - FR4 substrate: εr=4.3, 1.5 mm nominal. NOTE: with the committed
-    graded z-mesh the substrate Box RASTERIZES to 2 coarse cells, not
-    the intended 6 fine cells (issue #325); re-pinning it onto the fine
-    band was measured to DEGRADE the resonance agreement, so the
-    committed envelope is kept as-is. (OpenEMS side: 4 uniform z-cells.)
+  - FR4 substrate: εr=4.3, 1.5 mm nominal, now correctly RASTERIZED to
+    6 fine cells / 1.5 mm (issue #325 fixed): smooth_grading shifts the
+    fine block up in absolute z, so the stack is registered to where the
+    fine cells actually land (see the #325 block below). An earlier note
+    kept the mis-registered 2-cell/1.361 mm substrate because re-pinning
+    "degraded resonance agreement" — but that agreement was
+    harminv-vs-analytic under a mode-AMBIGUOUS selector (closest-to-
+    analytic), which flips between modes as the geometry/subpixel changes.
+    This script now IDs the radiating mode by far-field broadside (below),
+    which is stable, and the primary gate is rfx-vs-OpenEMS (not analytic),
+    re-verified on the VESSL external-crossval lane. The corrected 6-cell
+    substrate reads ~9 % above the (5-8 %-approximate) analytic — the
+    coarse-mesh high-bias of the radiating mode, not a registration bug.
+    (OpenEMS side: 4 uniform z-cells.)
   - Patch: PEC rectangle on top of the substrate
   - Air region above the patch (MUR / CPML open boundaries)
 
@@ -193,6 +202,7 @@ from rfx.boundaries.spec import BoundarySpec
 from rfx.sources.sources import GaussianPulse, ModulatedGaussian
 from rfx.auto_config import smooth_grading
 from rfx.harminv import harminv
+from rfx.farfield import compute_far_field, directivity
 import jax.numpy as jnp
 
 # Non-uniform z mesh: coarse air below, fine substrate, coarse air above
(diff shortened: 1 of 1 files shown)


bk-squared:
Closing — this branch re-applies an approach a PRE-DECLARED FALSIFIER ALREADY STOPPED, and its flagship mode-ID is unsound on this fixture. A from-first-principles cv05 review (with adversarial physics/rigor critics) caught it; the load-bearing evidence is verified against source:

1. **R2 rediscovery of a fired STOP.** `docs/research_notes/20260711_cv05_multiobservable_design.md:58-61`: an earlier session already re-registered the substrate to 6 fine cells, ran the openEMS gate (VESSL 369367246676, np=120 settled), and recorded: buggy 2-cell = **2.65% vs openEMS** (clean fundamental 2.5528 GHz); fixed 6-cell = **6.45% vs openEMS (WORSE)** — the 6-cell substrate **SPLITS** the mode (2.1425 / **2.6471(dom)** / 3.4456, all high-Q) because the graded-mesh fine↔coarse transition sits adjacent to the resonant dielectric. "**Pre-declared falsifier FIRED → STOP; z-mesh fix #325 NOT committed; gate untouched.**" My commit 65e2aa9's harminv 2.6472 GHz is exactly that split-off dominant peak — I did not cite this note (R1 miss).

2. **The far-field broadside mode-ID (this branch's flagship) is unsound on cv05's geometry.** cv05's 60×55 mm ≈ 0.48λ ground plane radiates OFF-broadside (±32° double-hump / broadside DIP; design note :124-128). The rfx "broadside peak" exists only because the sub-λ/2 NTFF box smooths the real off-broadside structure into a false peak; give the box ≥λ/2 clearance and it fails its own ±20° broadside test → falls back to closest-to-analytic.

3. **The 20% gate is near-vacuous** — it provably PASSES on both the broken 2-cell geometry AND the split 6-cell geometry; the "agreement" is a documented z-under-res(HIGH)/staircase(LOW) error cancellation, not convergence.

Correct path (see #325): cv05 should be demoted claims-bearing→diagnostic-reporter and its accuracy claims delegated to committed tests (canonical-patch NF2FF + a uniform-mesh cavity oracle + fine-mesh harminv tests), wiring the already-landed G1–G5 guards. The naive 6-cell re-registration on the graded mesh is NOT the fix — the resonant substrate needs a uniform-fine band with no adjacent grading transition. Full review filed as a research note.

PR #379, opened 2026-07-17
test(#325): lock the uniform-fine substrate grid build; fail-closed on the split geometries

bk-squared:
First deliverable of the #325/#330 patch-validation redesign (design workflow + adversarial critics). A **@fast grid-build lock — no FDTD, no openEMS**.

## Root cause (verified)
cv05 places the ground/substrate/patch stack at a FIXED pre-smoothing z (`air_below=12mm`, `05_patch_antenna.py:138,161-163`) while `smooth_grading` inserts transition cells that shift the fine band up — so the 1.5mm FR4 substrate builds as **2 coarse cells, not 6 fine cells** (#325).

## The lock
Imitates the verified, already-run z-derivation in `scripts/diagnostics/patch_tutorial_rfx.py::build` (NOT reinvented — the external-example rule): DERIVE the stack z from where `smooth_grading` actually put the fine band, and add uniform-fine BUFFER cells so the coarse↔fine transition sits clear of the resonator.

| geometry | substrate cells | transition clearance | lock |
|---|---|---|---|
| uniform-fine N_BUF=8/12/16 | **6** (1.500mm) | 2.13 / 3.13 / 4.13 mm | **PASS** |
| committed cv05 (fixed air_below) | 2 coarse | — | **FAIL** (the #325 bug) |
| zero-buffer 6-cell re-registration | 6 | ~0 (adjacent) | **FAIL** (fired STOP-1) |

The zero-buffer 6-cell case is the **fired STOP-1** (research note 20260711: transition-adjacent substrate splits the mode 2.14/2.65/3.45 and worsens openEMS 2.65%→6.45%). It appears here **only as a rejection guard** — a fails-closed check so that broken geometry can't be committed silently.

## Discipline
- **Does NOT reopen STOP-1**: uniform-fine (buffered) is the note-sanctioned correct direction, genuinely untried; the zero-buffer case is a guard, not an attempt.
- The **physics-decisive** step (does the buffered geometry give a single clean fundamental at the cv05 FR4 point, vs still split) is the staged **@slow two-buffer invariance witness (§1-B)**, not this grid-only lock.

## Verification
5 passed in 0.69s, ruff clean.

R3: memory=consistent (encodes the 20260711 STOP as a guard; #325 correct-direction; imitates patch_tutorial_rfx) | R2-attempts=0 (no STOP reopened) | falsifier=lock PASSES uniform-fine, FAILS committed + STOP-1 (shown).

### DIFF
--- a/scripts/diagnostics/patch_uniform_fine_fr4_witness.py
+++ b/scripts/diagnostics/patch_uniform_fine_fr4_witness.py
@@ -1,0 +1,142 @@
+"""#325 §1-B — the physics-decisive falsifier: does the UNIFORM-FINE substrate
+build give a single clean fundamental at the cv05 FR4 operating point (vs the
+mode-split the FIRED STOP-1 graded 6-cell re-registration produced)?
+
+The grid-build lock (tests/test_issue325_uniform_fine_substrate.py) proved the
+geometry rasterizes to 6 fine cells with the grading transition held clear. This
+runs the ACTUAL FDTD harminv at the cv05 FR4 point (eps_r=4.3, 38x29.5mm patch,
+60x55mm GP, 1.5mm/6-cell substrate) at two buffer widths (N_BUF=8, 16); the
+substrate is identical, only the transition-z moves. A physical mode is
+buffer-invariant; a transition artifact moves/splits.
+
+PRE-DECLARED FALSIFIER (one R2 attempt, terminates either way):
+  CLEAN  -> adopt uniform-fine: each buffer yields ONE dominant fundamental
+           (amp >= 3x any peak OUTSIDE the known TM-family band 2.9-3.2 GHz),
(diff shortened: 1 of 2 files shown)


bk-squared:
Update: the §1-B physics witness (now committed on this branch, 18c4609) returned **SPLIT** — the uniform-fine substrate exposes the same real multi-mode (TM10 2.65 / TM01 2.14 / higher 3.44, buffer-invariant freqs) as the graded 6-cell, so uniform-fine re-registration does NOT buy accuracy (the mode-split is real multi-mode, not a transition artifact; see #325 comment). **This grid-build lock still stands on its own merits**: it locks geometry HONESTY (substrate builds as the intended 6 cells, transition clear) and fails-closed on both the committed 2-cell bug and the STOP-1 zero-buffer geometry — correct independent of the accuracy verdict. Merging as a geometry lock, not an accuracy fix.

Issue #169, opened 2026-06-12
until_decay point-field stopper triggers before flux DFT convergence on guided geometries (cv03: T=0.745 at 2201 steps vs 0.967 fixed)


bk-squared:
Found during the #160 cv03 session (2026-06-12) while trying to apply the `rt_measurement.md` recipe's "until_decay is load-bearing" guidance to cv03.

## Measured discrepancy (same build, same geometry — cv03 straight dielectric guide, eps=12, 2d_tmz, bounded 2a flux monitors)

| run | steps | T(f_peak) |
|---|---|---|
| `until_decay=1e-5` at flux_out, `decay_min_steps=2000`, cap 11427 | **2201** (stopped right past min_steps) | **0.745** |
| fixed 400 a/c0 units | 5913 | 0.967 |
| fixed 3x | 17341 | 0.977 |

The point `ez` at flux_out decays below 1e-5 of its running peak by ~step 2200, but the flux DFT is still accumulating the slow low-group-velocity tail of the eps=12 guide (plus late radiation). Point-field decay at one position is not a flux-convergence witness for this class of geometry.

Note the asymmetry: Meep's `stop_when_fields_decayed(50, Ez, ..., 1e-3)` stopped at t=200 in the same lane run and its flux WAS converged (T_meep=0.9927). Meep's stopper compares the max over a trailing 50-time-unit window against the global max — a windowed criterion that tolerates oscillating tails. rfx's stopper semantics (instantaneous point ratio checked every `decay_check_interval`) appear more aggressive, and 1e-5 at a single point still fired ~3700 steps before the flux settled.

## Suggested scope

1. Investigate rfx `run_until_decay` semantics vs Meep's windowed criterion (`rfx/simulation.py:1741`).
2. Consider a flux-aware stopping criterion (monitor the flux accumulator residual, not a point field) or a windowed-max criterion.
3. `docs/agent-memory/task_recipes/rt_measurement.md` blanket "until_decay is load-bearing" guidance now carries a caveat for guided/slow-tail geometries (added in the #160 session); revisit once the stopper is fixed.

cv03 itself is fixed with a measured fixed-duration run (400 a/c0 units; see #160 / PR #165) — this issue is about the primitive, not the example.

bk-squared:
## Candidate C (flux-residual stop) IMPLEMENTED and FALSIFIED — R2 STOP (one attempt). Redesign proposed below; no attempt #2 without a go.

[...]

PR #380, opened 2026-07-17
docs(examples): canonical patch-antenna tutorial — far-field mode-ID + honest accuracy framing

bk-squared:
Promotes the **validated canonical patch far-field lane** (`scripts/diagnostics/patch_tutorial_rfx.py` — the rfx side of the 0.08 dB directivity agreement with openEMS-with-thirds) into a committed, user-facing tutorial: `examples/tutorials/patch_antenna_demo.py`.

## What it teaches (this repo's hard-won lessons, embedded honestly)
1. **Mesh-derived stack registration** (#325 lesson): ground/substrate/patch z derived from the *built* `dz_profile` (never fixed pre-smoothing z), fail-closed with `RuntimeError` on fine-band loss or rasterization mismatch. Realized layout printed.
2. **A patch is multi-mode — identify the radiating mode by far-field** (R-A lesson): full harminv mode list printed, then per-bin NTFF far-field, radiating mode selected by broadside beam + radiated-power peak — **not** amplitude rank, not closest-to-textbook (the cavity formula is printed but explicitly labelled "NOT the mode-identification criterion").
3. **Preflight output is part of the result**: no `skip_preflight` anywhere; all 4 advisories quoted verbatim (thin-PEC sub-cell ×2, bottom NTFF face reactive-near-field, small-GP edge diffraction).
4. **Settling witness before any frequency** (#332/G1): default `NUM_PERIODS=90` honestly prints **UNDER-SETTLED (−36.5 dB vs −40 bar)** with the consequence quantified (0.03 % / 0.03 dB vs the settled lane) and instruction to raise to ~105 — the demo *shows* the witness firing, on-brand for this repo.
5. **Honest accuracy framing**: coarse-dx resonance reads low vs openEMS-with-thirds (−8.6 %@dx2 / −6 %@dx1 / ~−3 % continuum; ~half the continuum residual is the collocated substrate-ε interface, recoverable with `subpixel_smoothing=True`, citing `tests/test_patch_cavity_eps_oracle.py`; rest is PEC-edge staircase). Far-field directivity is the validated-agreeing observable.

## Verification
- Ran end-to-end: **588 s CPU** (within the ≤10 min budget). f_res 2.2149 GHz / D 6.68 dBi vs the validated lane's 2.2143 / 6.71 (Δ from the documented lean-frame + settling deltas).
- `compute_far_field_jax` used (equivalence witnessed vs numpy path: max |ΔD| 3e-6 dB; 8.5× faster on CPU).
- Adversarial review (independent worktree): **SOUND** — grid+preflight re-run byte-matched, mode-ID logic verified, helpers unit-checked on analytic patterns, all API usage verified against source, no preflight suppression, lint clean.
- `examples/README.md` learning path updated (step 10; later steps renumbered).

R3: memory=consistent (imitates the validated patch_tutorial lane; encodes R-A far-field mode-ID + #325 mesh-derived registration + #332 witness) | R2-attempts=0 (promotion of proven evidence, no new physics attempt) | falsifier=the demo's own fail-closed registration asserts + the printed witness.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -15,9 +15,10 @@ New to rfx? Run these in order; each teaches one decision a real design needs.
 | 7 | `tutorials/ports_and_sparams_101.py` | which port for which structure (all five), S11 basics, real-world pitfalls |
 | 8 | `tutorials/resonance_harminv.py` | ring-down resonance extraction, picking modes by physics (not loudness), record-length vs resolution |
 | 9 | `tutorials/antenna_farfield_pattern.py` | far-field boxes done right (half-wavelength rule), directivity vs the textbook dipole |
-| 10 | `tutorials/rcs_scattering.py` | radar cross-section with incident-reference subtraction |
-| 11 | `inverse_design/differentiable_s11_design.py` | the differentiable design loop: `forward` + `jax.grad` + optimizer |
-| 12 | `tutorials/artifact_report_demo.py` | exporting a shareable scene/mesh/report bundle |
+| 10 | `tutorials/patch_antenna_demo.py` | a real antenna end to end: mesh-registered stack, picking the radiating mode by its far field (not loudness), settling witness, honest error budget vs openEMS |
+| 11 | `tutorials/rcs_scattering.py` | radar cross-section with incident-reference subtraction |
+| 12 | `inverse_design/differentiable_s11_design.py` | the differentiable design loop: `forward` + `jax.grad` + optimizer |
+| 13 | `tutorials/artifact_report_demo.py` | exporting a shareable scene/mesh/report bundle |
 
 `config/microstrip_thru.yaml` shows the declarative YAML front-end for the same
 Simulation API.
--- a/examples/tutorials/patch_antenna_demo.py
+++ b/examples/tutorials/patch_antenna_demo.py
@@ -1,0 +1,438 @@
+"""Patch antenna end to end: multi-mode ring-down, far-field mode ID, honest error budget.
+
+Geometry: the openEMS "Simple Patch Antenna" tutorial — a 32 x 40 mm patch on a
+1.524 mm, eps_r = 3.38, tan_delta = 1e-3 substrate over a 60 x 60 mm finite
+ground plane, probe-fed 6 mm off centre.  Running the identical geometry in
+openEMS (with that tutorial's thirds-rule edge meshing) gives the reference
+numbers quoted below: f_res = 2.4221 GHz, broadside directivity 6.79 dBi.
+
+What this tutorial teaches, in order:
+
+1. Thin-substrate stacks on a graded z-mesh must be registered to the BUILT
+   mesh.  ``smooth_grading`` inserts transition cells that shift the fine band,
+   so a stack placed at pre-smoothing coordinates can silently rasterize onto
+   coarse cells — the resonance then shifts or splits.  Every z coordinate here
+   is derived from the final ``dz_profile``.  ``nonuniform_patch_demo.py``
+   teaches the mesh side in detail; this script only applies the rule.
+
+2. A patch is MULTI-MODE.  ``harminv`` returns several ring-down modes, and the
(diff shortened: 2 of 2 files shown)


bk-squared:
## Post-merge empirical verification (codex-ra1, full 632 s live run on merged code)

**All docstring numbers reproduced:** settling witness −36.5 dB (exact match); f_res 2.2149 GHz vs the settled reference 2.2143 = 0.027 % (claimed "within 0.03 %"); D 6.68 vs 6.71 dBi = exactly the stated 0.03 dB; the −8.6 %-vs-openEMS recap prints verbatim.

**Mode-ID empirically confirmed, no mis-pick:** harminv found 2.2149 (radiating) + 2.7010 GHz (non-radiating); the far-field bins show a clear ~2.6× radiated-power gap (0.0 dB @2.2 vs −4.2 dB @2.8, the nearest bin to the risk mode). Structural note for future readers: in this run **all 7 NTFF bins passed the broadside ≤15° filter** (this GP/patch geometry doesn't scan far off broadside across the band), so the P_rad-peak criterion did the actual discriminating — the broadside filter is the second safety net and didn't need to fire here.

**Recorded follow-up (doc-only, optional):** the framework's own #332 nudge suggests `until_decay` instead of the fixed `NUM_PERIODS=90`, and sibling tutorials (`run_control_and_fields.py`, `slab_rt_flux_monitor.py`) already use it — that would ship the demo always-settled without a wall-clock guess. **Caveat for whoever picks this up:** choose the threshold to match the −40 dB witness bar (~`until_decay=1e-2` in amplitude terms); a blind `1e-4` (≈ −80 dB) would roughly double the runtime past the tutorial's ≤10 min budget. The current default (loud witness + quantified 0.03 % consequence + raise guidance) is a deliberate pedagogical choice — either is legitimate; not urgent.

bk-squared:
## until_decay follow-up: INFEASIBLE on this demo as suggested — the NU lane drops it (2026-07-18)

Attempted the recorded `until_decay` swap with an adversarially-reviewed implementation pass. **Honest STOP, independently reproduced:**

- The demo's graded z-mesh (`dz_profile`) dispatches to the **`run_nonuniform` lane** (verified live: `lane=run_nonuniform`, n_steps=18521 at num_periods=90), and that lane **warns-and-drops `until_decay` entirely** (`_execute.py:2454-2459` → `_warn_unsupported_run_kwargs`; the runtime warning prints verbatim: *"until_decay=0.0001 is silently ignored on the non-uniform mesh run path… use the uniform-mesh path"*). A bounded diagnostic run confirmed zero effect (fixed step count taken).
- The threshold unit analysis WAS correct (`decay_by` is an **energy** ratio — `|field|² < decay_by · peak` per `simulation.py:1946-1999` — so `1e-4` energy = the −40 dB amplitude bar), and the budget guard exists on the decay-capable lane (`decay_max_steps`, default 50k). The suggestion simply cannot take effect here.
- The sibling tutorials that use `until_decay` (`run_control_and_fields.py`, `slab_rt_flux_monitor.py`) are **uniform-mesh** — that's why it works there.
- Proceeding would require either (a) implementing decay-based termination inside the NU scan body — a new solver feature, not a docs follow-up — or (b) flattening the demo to a uniform mesh, which deletes its #325 graded-mesh teaching point.

**Resolution:** the demo keeps the honest-witness default (`NUM_PERIODS=90`, −36.5 dB printed with quantified 0.03 %/0.03 dB consequence + raise-to-~105 guidance). The capability gap is filed as a feature issue (until_decay support on the non-uniform run path) so the swap can happen if/when the lane supports it.

bk-squared:
Follow-up traceability: the until_decay conversion STOP recorded here (NU lane warn-and-drop) is resolved — #383 landed via PR #385 (merge 479b58e): until_decay now works on the non-uniform CPML/UPML lane (dV-weighted interior-energy stop, #169 contract). The demo conversion itself remains a separate follow-up; when picked up, note the demo's docstring/witness narrative is built around the fixed-period framing and needs rewriting alongside the run() call change.

Issue #404, opened 2026-07-20
physics: oblique TFSF reflection accuracy is UNVALIDATED — extracted |R| tracks ~normal incidence, not the injection angle


bk-squared:
Surfaced by the 2026-07-20 physics-validation audit fix for #397 (branch fix/397-400-physics-test-honesty). This is a real capability limitation, not a test artifact.

The genuine 2D-aux oblique TFSF path RUNS and injects a low-leakage tilted plane wave (leakage ~1.05e-6), but the **extracted reflection coefficient tracks ~normal incidence, not the oblique angle**:
- 30° / εr=4: analytic oblique |R_TE|=0.382 vs normal 0.333 differ only 12.7%; best_aligned extraction measures 0.352 — actually CLOSER to normal (5.7%) than oblique (7.7%). The existing "oblique" tests pass only because oblique≈normal at this shallow angle.
- 60° / εr=4: analytic oblique |R_TE|=**0.566**, extraction measures **~0.38** — 33% off oblique, ~13% off normal. A discriminating steep-angle gate FAILS on main.

Per the audit's do-not-weaken rule, a discriminating steep-angle gate was NOT added (it would red on main); the limitation is fenced test-visibly in the Test 3/4 docstrings of `tests/test_verification.py`, and the mislabeled `test_physics.py::test_fresnel_oblique_te` was renamed to `test_fresnel_normal_incidence_effective_eps_consistency` (it never actually injected an oblique wave).

**Direction candidates** (physics investigation, R2-tight — not scheduled): the 2D-aux TFSF injection amplitude/phase-front vs the reflection extractor's angle assumption; whether the aux-line update or the field projection at the TFSF interface carries the correct transverse-k; compare against a periodic-BC Floquet oblique setup as an independent oracle. Confirm on a known analytic case (single interface, steep angle) before any fix.

Reference audit artifact: `docs/research_notes/audits/20260720_pytest_physics_audit.json`.

bk-squared:
## Comparator-first investigation (2026-07-20): R2-STOP — the wavevector comparator itself fails its known-good; do NOT fix solver/extractor on broken measurement

Followed the rfx comparator-first discipline (validate the comparator before touching solver physics; the bug has historically been in the comparator/extractor). Three deep-read lanes + three diagnostics (artifacts: `docs/research_notes/experiments/i404_oblique_20260720/`, local-only). Findings, ordered by confidence:

[...]

PR #414, opened 2026-07-21
fix(tfsf): 3D Bloch-periodic-y oblique TFSF via opt-in complex solver (#404)

bk-squared:
## Summary

Resolves **#404** (oblique-incidence TFSF injects the wrong propagation angle). Phase-A (`d3e4a64`) fixed the 2D-auxiliary injection, but the 3D grid's plain-periodic-y boundary — a phase-free `jnp.roll` — could not sustain a non-commensurate transverse wavenumber and pulled steep angles back toward normal (60° → ~25–47°, domain-size-dependent, bandwidth-invariant).

**Fix:** the shared 3D solver now optionally evolves the Bloch-transformed field envelope `P` directly (physical field `= Re(P·e^{-j k_y y})`), on an **opt-in complex path**:

- `rfx/core/yee.py` — `update_e/h` gain an optional per-axis `bloch` phase `e^{-j k_axis dx}` on the periodic rolls (backward difference uses its conjugate) plus a complex-aware compute dtype. **The real path (`bloch=None`, real fields) is byte-identical.** Order-2 only (fail-loud guard for the wide stencil).
- `rfx/sources/tfsf_2d.py` — `apply_tfsf_2d_e/h` inject the aux envelope `P` directly when the 3D grid is complex (transformed frame), skipping the Phase-A `Re(P·e^{j k_y y})` round-trip a plain-periodic real grid cannot evolve. Adds `bloch_phase_tuple()`.

## Validation (pre-declared falsifier order)

| gate | result |
|---|---|
| real-path bit-identity | 30/30 arrays byte-identical |
| injected angle, non-commensurate periodic-y | 30/45/60° → 31/46/55°, **domain-size invariant** (ny=12/16/24/40 all 54.8° vs broken baseline 46.9→57.6) |
| TFSF leakage (ez/ey, ±x, ±angle) | **0.000%** (Phase-A: 2–10%) |
| Fresnel \|R\| vs analytic R_TE | 30° → 2.0%, 45° → 4.7% |
| AD gradient through complex path | finite, nonzero |

Real-path regression (stencil, AD, cavity, physics, normal TFSF) is green.

## Notes / honest caveats

- **The earlier `|R|≈1` was a comparator bug, not the physics.** The aux source is a complex *analytic* envelope `∝e^{-j2πf₀t}`, so its f₀ amplitude needs the **conjugate** DFT kernel `e^{+j2πf₀t}`; a `−j` kernel integrates to noise. Leakage/angle gates were immune (no DFT). The old `np.min` "best-aligned" extractor was oracle-fitted and is retired from the verification gates (replaced by a two-run plateau extractor with incident-flatness as its known-good).
- **60° in-rfx `|R|` is CPML-measurement-limited** — imperfect x-CPML absorption of the steep wave creates a partial standing wave (plateau flatness 0.002→0.044). The CPML-free kernel angle (54.8°, domain-invariant) is the trustworthy witness, so `|R|` is gated at 30/45° and steep-angle physics via the kernel angle + leakage. The residual ~5° at 60° is numerical dispersion at dx=λ/30.
- **Scope:** the #404 acceptance tests hand-roll the FDTD loop, so this closes them without complexifying the production `Simulation.run()` scan body (kept minimal per Correctness>sprawl). `compute_rcs`/open-domain oblique already worked (Phase-A). High-level `Simulation.run()` oblique-periodic support is tracked as a follow-up.

## Tests

New `tests/test_yee_bloch_complex_path.py` (bit-identity, complex→real reduction, order guard, angle-track 30/45/60 + domain-size invariance) runs in the default suite; narrowband oblique leakage (`test_tfsf_oblique*.py`) and Fresnel-`|R|` accuracy (`test_verification.py`) are `@pytest.mark.slow`.

R3: memory=`project_issue404_oblique_tfsf_r2stop.md` (Phase-B is the memory-named fix; supersedes the 2026-07-20 R2-STOP with a validated Poynting comparator + plateau-`|R|` known-good) · R2-attempts=1 (one pre-declared injection attempt; the extractor DFT-sign was a named implementation defect) · falsifier=real-path bit-identity + domain-size-invariant angle gate

### DIFF
--- a/rfx/api/_execute.py
+++ b/rfx/api/_execute.py
@@ -2358,6 +2358,22 @@ class _ExecuteMixin:
                 "S11 objectives."
             )
 
+        # TFSF plane-wave sources are not wired into the differentiable forward
+        # path: _forward_from_materials builds sources from lumped/wire ports
+        # only, so a TFSF source would be SILENTLY DROPPED (no injection → zero
+        # gradients — the "TFSF→forward silent-drop" footgun). Fail loud instead.
+        # (The distributed forward path rejects TFSF separately; guard the
+        # single-device path here so both normal and oblique TFSF are surfaced.)
+        if self._tfsf is not None and not distributed:
+            raise NotImplementedError(
+                "TFSF plane-wave sources are not wired into Simulation.forward() "
+                "— the differentiable path builds sources from lumped/wire ports "
+                "only and would silently drop the TFSF source (zero gradients). "
+                "Use run() for TFSF forward simulation (normal or oblique, #404); "
+                "differentiable TFSF/plane-wave inverse design is not yet "
+                "implemented."
+            )
+
         if port_s11_freqs is not None:
             self._validate_forward_sparameter_request()
 
--- a/rfx/boundaries/cpml.py
+++ b/rfx/boundaries/cpml.py
@@ -271,7 +271,8 @@ def _cpml_noop_profile(n_layers: int) -> CPMLParams:
 
 def init_cpml(grid, *, kappa_max: float | None = None,
               pec_faces: set[str] | None = None,
-              pmc_faces: set[str] | None = None) -> tuple[CPMLAxisParams, CPMLState]:
+              pmc_faces: set[str] | None = None,
+              field_dtype=jnp.float32) -> tuple[CPMLAxisParams, CPMLState]:
     """Initialize CPML parameters and zero-valued auxiliary fields.
 
     Creates per-axis CPML profiles so that each face uses the correct
@@ -352,7 +353,11 @@ def init_cpml(grid, *, kappa_max: float | None = None,
     nx, ny, nz = grid.shape if hasattr(grid, 'shape') else (grid.nx, grid.ny, grid.nz)
 
     def _zeros(dim_size: int, perp1: int, perp2: int) -> jnp.ndarray:
-        return jnp.zeros((n, perp1, perp2), dtype=jnp.float32)
+        # psi carry dtype follows the field dtype so the lax.scan carry contract
+        # holds when fields are complex (#404 oblique Bloch path). Default
+        # float32 keeps the real path byte-identical. Real profile coeffs
+        # (b/c/kappa) stay float32 and multiply the psi carry without truncation.
+        return jnp.zeros((n, perp1, perp2), dtype=field_dtype)
 
     state = CPMLState(
         # E-field psi (12 faces)
--- a/rfx/core/yee.py
+++ b/rfx/core/yee.py
@@ -127,12 +127,24 @@ def _ribbon_2nd(d4, d2, axis):
     return d4
 
 
-def _diff_fwd_o(arr, axis, periodic, order):
+def _diff_fwd_o(arr, axis, periodic, order, bloch=None):
     """Forward staggered first difference (f[i+1]-f[i] family), at i+1/2; the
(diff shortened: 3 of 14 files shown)


bk-squared:
### Added: high-level `Simulation.run()` wiring for oblique-periodic (commit f4b82d6)

Follow-up in this PR wires the complex Bloch path into `run()` so an oblique TFSF runs end-to-end without hand-rolling the loop (opt-in, gated on the 2D-aux discriminator; real path byte-identical):

- `_build_step_setup`: oblique 2D-aux → `field_dtype=complex64` + `bloch` threaded onto `_StepContext` and the canonical `update_h`/`_update_e_with_optional_dispersion`. Fail-loud on NTFF/DFT-plane/flux monitors. (Fast path is provably never selected for oblique TFSF.)
- `init_cpml(field_dtype=…)`: psi carry follows the field dtype so the `lax.scan` carry contract holds for complex fields.
- `run_uniform`: reconstruct `state` + point-probe `time_series` to physical fields `Re(P·e^{-jk_t·y})`; fail-loud on `snapshot=`/`until_decay=`.

**Validated:** `run(angle_deg=30)` reconstructed Ez matches the hand-rolled complex loop to **1.8e-6**, leakage 0.000%; normal incidence stays float32; broad real-path regression green (cavity/physics/dispersive-Fresnel normal-TFSF/PMC = 18 passed). Committed AD gate (`test_bloch_path_is_differentiable`) confirms grad flows through the complex+Bloch kernel.

**Scope:** `forward()`/inverse-design on oblique-periodic is a separate forward-path source-coupling gap (not wired); frequency-domain monitors are fenced (state + point probes are the supported outputs). `compute_rcs` open-domain oblique already worked (Phase-A).

bk-squared:
### Correction re: `forward()` + TFSF (my earlier "oblique forward gap" note was inaccurate)

Follow-up investigation (comparator-first) shows the earlier "forward-path source-coupling gap" was a **broken smoke**, not a real oblique-specific gap:
- `ForwardResult` has **no `.state`** field (only `time_series`); the smoke read `.state` and registered **no probe** → empty series → grad 0.
- With a proper probe, **normal and oblique `forward()` behave identically** (both grad 0) → not oblique-specific.
- Real cause: `_forward_from_materials` (the differentiable path) has **zero TFSF handling** and **silently drops** the TFSF source. This is the pre-existing "TFSF→forward silent-drop" footgun — **independent of #404 and of this PR's `run()` wiring**. (The *distributed* forward path already rejects TFSF fail-loud at `_execute.py:2046`; the non-distributed path does not.)

So: `run()` oblique works (validated in this PR); inverse-design on **any** TFSF (normal or oblique) via `forward()` is unsupported and **pre-existing**. Bounded follow-up (separate PR): make the non-distributed forward path fail-loud on TFSF to surface the footgun. Full differentiable TFSF inverse-design is a separate large feature.

bk-squared:
### Added: fail-loud guard for `forward()` + TFSF (commit 697650a)

Acting on the corrected finding: `forward()` silently drops TFSF sources (zero gradients). It now **fails loud** on a TFSF source (single-device path; the distributed path already rejects TFSF separately), with a message pointing to `run()` for TFSF forward simulation. New test `test_forward_rejects_tfsf_source_fail_loud` (normal + oblique); verified it doesn't disturb `test_forward_distributed_rejects_tfsf_on_nu_path` or `test_forward_s11_rejects_non_port_source_only_request`.

Full differentiable TFSF/plane-wave inverse design remains a separate, larger feature (not implemented).

PR #461, opened 2026-07-25
feat(rcs): open-domain oblique TFSF (method B) + lift the oblique fence for 2.5-D ez (Item C S1-S5)

bk-squared:
**Split 2 of 3** from the 8-commit Item-C branch (#451), per its own review note ("Reviewer may prefer to split (oblique-RCS feature vs crossval examples) — PI's call"). Authored in a prior session; this PR isolates the **physics/solver** half so it gets a focused review. Full original tip `0a4f1ea` preserved at `backup/oblique-rcs-item-c-full-20260725`.

**Now based on `main`** — #451 (split 1) merged as `d77174d`, so this was retargeted and rebased onto main via `git rebase --onto origin/main` (the stacked-squash recipe). The api-reference JSON merged cleanly with #449 Item B's `mu_r_override` entry; inventory verified to carry BOTH entries, gate `OK`. Re-verified after rebase: 12 passed (fast), ruff clean.

## What lands (4 commits, +845/-54)
- `7b9d4db` **method-B source** `rfx/sources/tfsf_oblique_open.py` (402 lines) — open-domain oblique TFSF: dispersion-matched 1-D aux along k-hat + 4-edge box with `r.k-hat` face projection. Real float32 fields, so NTFF/DFT read physical fields directly (no envelope-aware NTFF needed) — unlike the periodic complex-Bloch path (#414), which assumes lateral periodicity and is the wrong tool for a compact scatterer.
- `0bcac56` **Stage 2** — wired into `run()`/`forward()` via a dispatch mirroring `is_tfsf_2d`.
- `b3acd64` **fence LIFTED** — `compute_rcs(theta_inc != 0)` routes through method B for **ez + uniform grid only**. `ey` and non-uniform still raise `NotImplementedError`; `theta=0` is byte-identical to the normal path.
- `3c00bf6` **api-reference fix (added here)** — `add_tfsf_source` gained `method=`, a genuine public-surface drift that was unrecorded, so the gate was red. Regenerated; diff is exactly one entry.

## Verification (all re-run in this session, not carried over)
- `tests/test_oblique_rcs_specular.py` + oblique TFSF + RCS suites: **7 passed** (`-m slow`), **16 passed** (fast). `api reference surface: OK`. ruff clean on the gated scope (`rfx/ tests/`).
- The gate carries a **ring-down settling witness** (end/peak energy < -35 dB) and a **normal-reduction comparator** (theta=0 puts the lobe at phi=180, i.e. the known-good backscatter case).

## Added-gate audit (`feedback_gate_can_bind_artifact` mandate) — one finding for the reviewer
The rule requires a falsifier re-run **and** domain-size invariance for any newly added gate. Both were run against the public `compute_rcs` path (script: scratchpad `c2_gate_audit.py`; raw output in the review comment below).

**1. Angle discrimination (falsifier) — PASS, with a wide margin.** Cross-comparing measured patterns against the wrong prediction:

| pattern | vs pred(20)=160 | vs pred(40)=140 |
|---|---|---|
| theta=20 -> peak 160.0 | err 0.0 -> PASSES (own) | err 20.0 -> FAILS (correct) |
| theta=40 -> peak 144.0 | err 16.0 -> FAILS (correct) | err 4.0 -> PASSES (own) |

Discrimination margin (16-20 deg) is far larger than the tolerance (6 deg), so the gate genuinely measures the reflection law rather than passing vacuously.

**2. Domain-size invariance — the peak azimuth WANDERS 3-7 deg.** Enlarging the domain ~1.35x (dx, plate, CPML layers fixed):

| theta | base grid (76,130,24) | big grid (96,166,24) | shift |
|---|---|---|---|
| 20 | peak 160.0 (err 0.0), spec/back 12.9 | peak 163.0 (err 3.0), spec/back 15.9 | **3.0 deg** |
| 40 | peak 144.0 (err 4.0), spec/back 46.0 | peak 137.0 (err 3.0), spec/back 32.1 | **7.0 deg** |

Both domains pass the 6 deg gate and the specular lobe dominates backscatter throughout (13-46x), so the **reflection-law claim holds**. But the domain-dependent scatter (3-7 deg) is the *same size as the gate's own tolerance*, and at theta=40 it exceeds it. Related: the in-code docstring records `161/140 deg` via the direct `_methodB_pattern` helper while the public `compute_rcs` path gives `160/144 deg` — a consistent few-degree systematic between the two paths.

**Reading**: the argmax azimuth of a 2.2-lambda plate's broad specular lobe is genuinely sensitive at the few-degree level to finite-domain interference, so 6 deg is absorbing a systematic rather than bounding a converged number. I have **not** touched the tolerance — the no-silent-gate-loosening rule requires a written root cause first, and the gate's design is the author's call. Reviewer options: gate on a lobe **centroid** instead of argmax (less bin-interference sensitive), or keep argmax and justify the tolerance with the measured domain scatter recorded in the docstring. Note the PR that authored this claimed "tracked <=1 deg"; measured here it is 0 deg at theta=20 but 4 deg at theta=40 via the public API.

## Scope limits (in-code, unchanged from the author's version)
Only the far-field **pattern / specular direction** is validated. **Absolute oblique sigma is NOT validated** — the 2.5-D strip's 3-D RCS scales with the arbitrary NTFF z-box height (the two z-faces cancel for x-y-plane observation) and the incident-spectrum normalization assumes the normal-path waveform.

R3: memory=project_issue404_oblique_tfsf_r2stop + project_diff_planewave_inverse_design + feedback_gate_can_bind_artifact | R2-attempts=0 | falsifier=cross-prediction discrimination (PASS) + domain-size invariance (3-7 deg scatter SURFACED)

### DIFF
--- a/docs/guides/api_symbol_inventory.json
+++ b/docs/guides/api_symbol_inventory.json
@@ -2372,7 +2372,8 @@
       "polarization",
       "direction",
       "angle_deg",
-      "waveform"
+      "waveform",
+      "method"
     ],
     "add_thin_conductor": [
       "shape",
--- a/rfx/api/__init__.py
+++ b/rfx/api/__init__.py
@@ -1496,6 +1496,7 @@ class Simulation(
         direction: str = "+x",
         angle_deg: float = 0.0,
         waveform: str = "differentiated_gaussian",
+        method: str = "bloch",
     ) -> "Simulation":
         """Add a normal-incidence plane-wave TFSF source.
 
@@ -1526,6 +1527,16 @@ class Simulation(
             Spectrum is a Gaussian centered at ``f0`` with 1/e
             half-width ``f0·bandwidth``. Use this for matched rfx-vs-Meep
             crossval comparisons.
+        method : {"bloch", "methodB"}
+            Oblique-incidence engine (ignored for ``angle_deg=0``, which always
+            uses the normal 1D-aux path). ``"bloch"`` (default) is the narrowband
+            complex-Bloch 2D-aux path (fields go complex64; frequency-domain
+            monitors are not transform-aware — see ``tfsf_2d.py``). ``"methodB"``
+            is the open-domain oblique path (real ``float32``, 1D-aux-along-k̂ +
+            4-edge box): the transverse axis becomes OPEN/CPML, z stays thin
+            periodic (2.5-D, z-invariant far field), and NTFF/DFT/flux monitors
+            read physical fields. ``"methodB"`` currently requires
+            ``polarization='ez'`` and a single-device uniform grid.
         """
         if self._boundary != "cpml":
             raise ValueError("TFSF plane-wave source requires boundary='cpml'")
@@ -1563,6 +1574,19 @@ class Simulation(
                 "waveform must be 'differentiated_gaussian', 'modulated_gaussian', "
                 f"or 'continuous_wave', got {waveform!r}"
             )
+        if method not in ("bloch", "methodB"):
+            raise ValueError(f"method must be 'bloch' or 'methodB', got {method!r}")
+        if method == "methodB":
+            if abs(angle_deg) <= 0.01:
+                raise ValueError(
+                    "method='methodB' is the open-domain OBLIQUE path; use a "
+                    "non-zero angle_deg (angle_deg=0 uses the normal 1D-aux path)"
+                )
+            if polarization != "ez":
+                raise NotImplementedError(
+                    "method='methodB' currently supports polarization='ez' "
+                    "(transverse=y) only; 'ey' is future work"
+                )
 
         self._tfsf = _TFSFEntry(
             f0=f0,
@@ -1573,6 +1597,7 @@ class Simulation(
             direction=direction,
             angle_deg=angle_deg,
             waveform=waveform,
+            method=method,
         )
         return self
 
--- a/rfx/api/_execute.py
+++ b/rfx/api/_execute.py
@@ -1325,14 +1325,23 @@ class _ExecuteMixin:
                 ny=grid.ny,
                 nz=grid.nz,
                 waveform=getattr(self._tfsf, "waveform", "differentiated_gaussian"),
+                method=getattr(self._tfsf, "method", "bloch"),
             )
             # NOTE: the TFSF vacuum-boundary check runs at forward() entry via
             # _auto_preflight on the concrete config; it is NOT re-run here because
             # `materials` may be a jax tracer under jax.grad (eps_override), and the
             # check concretizes.
-            periodic_bool = (False, True, True)
-            cpml_axes_run = "x"
-            pec_axes_run = ""
+            # Open-domain oblique Method B forces OPEN transverse y (CPML) with
+            # thin-periodic z; all other TFSF keep the historical open-x/periodic-yz.
+            from rfx.sources.tfsf import is_tfsf_methodB as _is_methodB_fwd
+            if _is_methodB_fwd(tfsf_run[0]):
+                periodic_bool = (False, False, True)
+                cpml_axes_run = "xy"
+                pec_axes_run = ""
+            else:
+                periodic_bool = (False, True, True)
+                cpml_axes_run = "x"
+                pec_axes_run = ""
 
         # Stage 2 Kottke for AD-traceable PEC density (opt-in via env
         # ``RFX_PEC_OCC_KOTTKE=1``).  When enabled and
--- a/rfx/api/_spec.py
+++ b/rfx/api/_spec.py
@@ -1105,6 +1105,7 @@ class _TFSFEntry:
     direction: str
     angle_deg: float
     waveform: str = "differentiated_gaussian"
+    method: str = "bloch"
 
 
 @dataclass(frozen=True)
--- a/rfx/rcs.py
+++ b/rfx/rcs.py
@@ -201,7 +201,9 @@ def compute_rcs(
         steps, public path, corner-inclusive kernels): peak error 0 deg at
         theta_inc=20 and 3 deg at theta_inc=40 (4 deg on the pre-fix exclusive
         kernels), gated at +/-6 deg, with a 3-7 deg peak-azimuth
-        sensitivity to domain size at fixed dx/plate/CPML (PR #461 audit) —
+        sensitivity to domain size at fixed dx/plate/CPML (PR #461 audit,
+        measured on the PRE-fix exclusive kernels; not re-measured in-tree
+        on the inclusive kernels) —
         i.e. the argmax azimuth of the broad specular lobe is NOT converged to
         ~1 deg. See ``tests/test_oblique_rcs_specular.py``. The absolute
         oblique sigma is NOT validated. Requires ``polarization='ez'`` and a
--- a/rfx/rcs.py
+++ b/rfx/rcs.py
@@ -196,10 +196,16 @@ def compute_rcs(
         illumination in the x-y plane and route through the open-domain oblique
         Method-B TFSF (2.5-D: open transverse y, thin-periodic z). Only the
         far-field PATTERN / specular direction is validated at oblique incidence
-        (specular peak phi = 180 - theta_inc at theta_obs = 90 deg, tracked to
-        ~1 deg; see ``tests/test_oblique_rcs_specular.py``); the absolute oblique
-        sigma is NOT validated. Requires ``polarization='ez'`` and a uniform grid;
-        other combos raise ``NotImplementedError`` (issue #404 arc).
+        (specular peak phi = 180 - theta_inc at theta_obs = 90 deg). Measured
+        envelope on the committed gate configuration (2.2-lambda plate, 700
+        steps, public path): peak error 0 deg at theta_inc=20 and 4 deg at
+        theta_inc=40, gated at +/-6 deg, with a 3-7 deg peak-azimuth
+        sensitivity to domain size at fixed dx/plate/CPML (PR #461 audit) —
+        i.e. the argmax azimuth of the broad specular lobe is NOT converged to
+        ~1 deg. See ``tests/test_oblique_rcs_specular.py``. The absolute
+        oblique sigma is NOT validated. Requires ``polarization='ez'`` and a
+        uniform grid; other combos raise ``NotImplementedError`` (issue #404
+        arc).
     phi_inc : float
         Incident azimuth in degrees (reserved for future oblique support).
     polarization : str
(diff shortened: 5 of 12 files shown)


bk-squared:
Raw output of the added-gate audit referenced in the PR body (re-run just now, verbatim):

```
==============================================================================
CHECK 2 -- DOMAIN-SIZE INVARIANCE (physical lobe stays; artifact moves)
==============================================================================
theta=  20  base grid=(76, 130, 24)  peak_phi= 160.0  pred=160.0  err= 0.0 deg  spec/back=   12.92
theta=  20  big  grid=(96, 166, 24)  peak_phi= 163.0  pred=160.0  err= 3.0 deg  spec/back=   15.85
theta=  40  base grid=(76, 130, 24)  peak_phi= 144.0  pred=140.0  err= 4.0 deg  spec/back=   45.95
theta=  40  big  grid=(96, 166, 24)  peak_phi= 137.0  pred=140.0  err= 3.0 deg  spec/back=   32.11

theta=  20: peak shift base->big = 3.0 deg (MOVED -- artifact suspected)
theta=  40: peak shift base->big = 7.0 deg (MOVED -- artifact suspected)

==============================================================================
CHECK 1 -- ANGLE DISCRIMINATION (falsifier: cross-compare predictions)
==============================================================================
measured peak: theta=20 -> 160.0 deg ; theta=40 -> 144.0 deg
  theta=  20 pattern vs pred(  20)=160.0: err=  0.0 deg -> gate PASSES (own prediction)
  theta=  20 pattern vs pred(  40)=140.0: err= 20.0 deg -> gate FAILS  (WRONG prediction -- must FAIL)
  theta=  40 pattern vs pred(  20)=160.0: err= 16.0 deg -> gate FAILS  (WRONG prediction -- must FAIL)
  theta=  40 pattern vs pred(  40)=140.0: err=  4.0 deg -> gate PASSES (own prediction)

DISCRIMINATION: PASS -- gate distinguishes injection angles
```

Script: scratchpad `c2_gate_audit.py` — builds the same plate/grid as `tests/test_oblique_rcs_specular.py` and measures through the public `compute_rcs` API at two domain sizes.

bk-squared:
Reading of the specular-azimuth gate, independently audited by a second agent (Codex) before posting. It agreed with the diagnosis but **narrowed two of my claims and found a confound that changes the recommended experiment** — that part is the most useful thing here.

## The gate question, restated

The added gate compares the specular peak azimuth against the reflection law with a ±6° tolerance, and the observable is a **one-degree-binned `argmax`** over the reflection hemisphere (`test_oblique_rcs_specular.py:70-72`, gate at `:120`).

Your own two audits pull in opposite directions, and both readings are correct:

- **Falsifier passes cleanly.** θ=20 → 160°, θ=40 → 144°; swapping the expectations gives 20° and 16° errors and correctly fails. Discrimination margin 16–20° ≫ tolerance 6°, so the gate genuinely measures the reflection law.
- **Domain invariance surfaces non-convergence.** At ×1.35 domain: θ=20 160→163° (3°), θ=40 144→137° (7°). All four results stay within 0–4° of their own reflection-law target and the specular lobe keeps 12.9–46× dominance over backscatter, so **the reflection-law claim holds**. But the domain-dependent scatter is the same size as the tolerance and exceeds it at θ=40.

So the tolerance is absorbing an observed domain sensitivity rather than bounding a converged number. One qualification I'd make explicitly: a single enlargement establishes *sensitivity*, not that finite-domain interference is uniquely the cause.

## The confound that matters for what to do next

`compute_rcs` ties the NTFF surface to the CPML faces — `j_lo`/`j_hi` are derived from them (`rfx/rcs.py:365-370`). **Enlarging the domain therefore also moves the Huygens surface.** That means the 3–7° scatter has several competing explanations that the current audit cannot separate: CPML reflection, Huygens-surface placement, ring-down length, angular sampling, and (if you switch observables) the centroid window.

Consequence: if a centroid's own scatter also came out ~5°, that would **not** establish "the plate is too small" — I had that inference in my first draft and it doesn't hold. Only a plate-size-specific improvement would support it. The defensible study is a small matrix over domain, CPML depth, NTFF offset, run length, angular resolution and plate size.

## On replacing argmax with a lobe centroid

I'd still suggest it, but as the **next experiment rather than a justified replacement**, and with a caveat I originally missed:

For a finite plate the reflection law identifies the stationary-phase/specular *direction*, but the power **centroid of a finite, polarised diffraction lobe need not lie there** — element-factor weighting, edge diffraction and the nonlinear azimuth coordinate all skew it. Near grazing, the current 90°–270° reflection-hemisphere window clips the lobe as the expected angle approaches 90°, biasing the centroid inward. A centroid is also sensitive to how its aperture is defined (first-null vs fixed-width vs thresholded vs whole hemisphere), and this code samples a one-dimensional `theta_obs=90°` cut, not a solid-angle centroid.

Net: a centroid is still better than a single-bin argmax for these broad, isolated 20°/40° lobes — but only with a **predeclared connected-lobe window** and a **finite-plate / physical-optics centroid target** rather than the specular angle. And **argmax should stay as a secondary diagnostic**, because the reflection law predicts the peak more directly than it predicts a finite-aperture centroid.

Either way I'd keep your decision not to touch the tolerance. Whichever observable you land on, the tolerance should come from a measured convergence study rather than from what the current number happens to need.

## Separate defect, worth fixing regardless

The `161/140°` vs `160/144°` difference is **not** evidence that the two paths disagree — the docstring witness is explicitly labelled "bigger 3λ plate, n_steps=1400" (`:25-27`) while the public gate uses a 2.2λ plate and 700 steps (`:48-49`). Different configurations, so the numbers aren't comparable. I had this wrong.

What *is* a defect:

1. `rfx/rcs.py:199` claims tracking to "~1 deg", which the public-path 4° result at θ=40 contradicts. That claim should be corrected to the measured envelope.
2. `_methodB_pattern` says it "mirrors" `compute_rcs` while using **different NTFF bounds** (helper `:75-89` vs public path `:365-370`). Either share the box construction or label the two configurations distinctly — otherwise their numeric difference invites exactly the misreading I just made.

Neither blocks the physics; both are claim-honesty items in the family the no-silent-gate-loosening rule exists to protect.

## What I'd not change

The method-B source itself reads well and the choice is well argued: a dispersion-matched 1-D auxiliary along k̂ with real float32 fields lets NTFF/DFT read physical fields directly, which is the right call for a compact scatterer where the periodic complex-Bloch path (#414) would be the wrong tool. Lifting the fence for ez + uniform only is appropriately narrow.

bk-squared:
Follow-up with **measured data** instead of options — pushed `912009a` onto this branch (docs-only claim corrections), and ran the convergence matrix the audit called for. Raw numbers below; each run is the **public `compute_rcs` path** on the committed gate configuration unless a column says otherwise (~4-6 s/run, CPU).

## The matrix (argmax vs windowed centroid)

Centroid = power-weighted mean over the **connected lobe above half-max around the peak** (predeclared window). `pred = 180 − θ`.

```
th=20 scale=1.00 off=3 plate=2.2λ | argmax=160.0 (err +0.0) centroid=159.6 (err −0.4) lobe_bins=26
th=20 scale=1.00 off=6 plate=2.2λ | argmax=159.0 (err −1.0) centroid=159.0 (err −1.0) lobe_bins=25
th=20 scale=1.00 off=9 plate=2.2λ | argmax=159.0 (err −1.0) centroid=158.6 (err −1.4) lobe_bins=26
th=20 scale=1.35 off=3 plate=2.2λ | argmax=164.0 (err +4.0) centroid=162.7 (err +2.7) lobe_bins=24
th=20 scale=1.35 off=3 plate=3.0λ | argmax=162.0 (err +2.0) centroid=161.6 (err +1.6) lobe_bins=18
th=40 scale=1.00 off=3 plate=2.2λ | argmax=144.0 (err +4.0) centroid=142.3 (err +2.3) lobe_bins=29
th=40 scale=1.00 off=6 plate=2.2λ | argmax=143.0 (err +3.0) centroid=141.7 (err +1.7) lobe_bins=28
th=40 scale=1.00 off=9 plate=2.2λ | argmax=142.0 (err +2.0) centroid=140.4 (err +0.4) lobe_bins=29
th=40 scale=1.35 off=3 plate=2.2λ | argmax=137.0 (err −3.0) centroid=138.4 (err −1.6) lobe_bins=34
th=40 scale=1.35 off=3 plate=3.0λ | argmax=138.0 (err −2.0) centroid=138.1 (err −1.9) lobe_bins=23
```

(Reproduces your audit: my scale-1.35 argmax lands 164/137 vs your 163/137 — one bin apart at θ=20, consistent with the bin-quantisation story. No lobe window was clipped by the 90–270° hemisphere at these angles.)

## What each axis says

1. **The NTFF-placement confound is small.** Moving the Huygens y-faces by `ntff_offset` 3→9 at fixed domain shifts argmax ≤2° and centroid ≤1.9°. So the 3–7° domain sensitivity is **not** primarily the NTFF surface riding the CPML faces — the residual candidates are the domain interference pattern itself (CPML return, run length) rather than box placement.
2. **The centroid is measurably better, as predicted.** Across the whole matrix: max |err| **2.7° (centroid) vs 4.0° (argmax)**, and it cuts the θ=40 domain swing roughly in half (Δ3.9° vs Δ7.0°). The finite-plate centroid bias the second audit warned about is real but small here — ≤2.7° including domain effects, at θ ≤ 40° where the hemisphere window never clips the lobe.
3. **Plate size helps at θ=20** (both observables improve at 3λ) and is **ambiguous at θ=40** (argmax improves, centroid unchanged within a bin). So "the plate is too small" is at most part of the story — consistent with the confound analysis.

## Concrete recommendation (author's call, but now data-backed)

- **Gate on the half-max-window centroid at ±5°** (measured envelope 2.7° + margin ≈ the same headroom ratio the current gate has), **keep argmax as a secondary diagnostic** at its current ±6° — the reflection law predicts the peak more directly than a finite-aperture centroid, so keeping both costs nothing and preserves the falsifier property (cross-prediction discrimination is 16–20°, far above either tolerance).
- If you'd rather keep argmax-only: the measured envelope supports ±6° as-is (max deviation +4/−3 across domains), but the docstrings should say — and after `912009a` they now do — that this is an observed envelope on an unconverged observable, not a converged-accuracy claim.

`912009a` also fixes the two claim-honesty items regardless of the gate decision: `rcs.py`'s "tracked to ~1 deg" → measured envelope, and `_methodB_pattern`'s "mirrors compute_rcs" → the exact NTFF-box construction difference (y-faces pinned to the TFSF box vs derived from CPML faces; `i_hi` off by one), plus a module-docstring note that the 3λ/1400-step witness numbers are not comparable to the committed-gate numbers. Gate and tolerance untouched.

PR #462, opened 2026-07-25
crossval: real-structure campaign — rect cavity (exact Pozar), Sheen LPF (openEMS + Palace referee), patch antenna

bk-squared:
**Split 3 of 3** from the 8-commit Item-C branch (#451), per its own review note. Authored in a prior session; independent of the oblique-RCS physics (#451, #461) — touches no `rfx/` solver code. Full original tip `0a4f1ea` preserved at `backup/oblique-rcs-item-c-full-20260725`.

## What lands (4 commits, +9,747)
- `0f3d420` **#2 rect PEC cavity TE101** — rfx `harminv` vs **exact Pozar closed form**: 7/7 modes < 0.06%, 2nd-order convergent. Analytic oracle, no external solver.
- `fd4f8f7` **#4 Sheen 1990 microstrip LPF** — rfx vs **openEMS** (IEEE MTT 38:849).
- `6c8b4d9` **#5 microstrip patch** — rfx vs **openEMS**, thick low-eps RT/Duroid 5880 laminate, uniform `dx=h/4`. Carries a `mode=tutorial` reproduce-gate against openEMS's own canonical "Simple Patch Antenna" example.
- `5aac821` **Palace-FEM referee for #4** (VESSL runs 369367248550 coarse / 369367248558 mid) — **overturns the #4 single-null framing**: the Sheen stopband is a DOUBLE transmission zero (Palace 7.032/8.048 GHz mid; openEMS 7.031/7.995, matching Palace to ~0.7%). rfx at 200 um distorts the doublet, so the earlier "9.6% first-null" gap was largely an **argmin-of-doublet comparator artifact** (`sides_with=openems`). `07_sheen_lpf.py` carries a SUPERSEDED banner. Bulk of the diff is recorded solver output (~6k lines of result JSON).

## Verification (re-run this session)
- `tests/test_sheen_lpf_palace_referee_gates.py` — **7 passed**. `api reference surface: OK`; ruff clean on the gated scope; `tests/test_crossval_manifest_contract.py` — 10 passed.
- **Added-gate falsifier** (`feedback_gate_can_bind_artifact`): perturbing the recorded fixture's `coarse.max_energy_sum` by 5% makes `test_all_blocks_passive` FAIL, then restored. So the suite genuinely compares recorded-vs-re-derived (`build_referee` re-derives from the raw Palace arrays) rather than passing vacuously.

## Reviewer decision required: script placement (I did NOT move these)
The three studies sit under **`scripts/crossval/`**. The repo's canonical home is **`validation/crossval/`** (per the #350-#359 restructure and `CLAUDE.md`), and `tests/test_crossval_manifest_contract.py` asserts the manifest covers **every** `validation/crossval/*.py` exactly once — with `role`, `public_group`, `evidence_levels`, `references`, `claim_scope`, `expected_exit_codes`, `gate_paths`. At the current path these studies are **outside that governance entirely** (the contract passes precisely because `scripts/` is not in its glob).

Moving them is not a rename, for two reasons:

1. **ID collision on 05.** `validation/crossval/05_patch_antenna.py` already exists as a **governed, deliberately demoted** case (`role=diagnostic-reporter`, demoted 2026-07-15, #378) whose `claim_scope` documents *why* it does not establish accuracy and delegates authoritative patch evidence to `test_patch_canonical_farfield_e4.py`. Since the manifest requires `script.stem == id`, dropping the new patch study in as `05_patch_antenna.py` would **overwrite that entry and its delegation**. The new study looks like a deliberate answer to what got cv05 demoted (it *has* the thirds-rule reproduce-gate the old reference omitted) — so the real question is whether it **supersedes** cv05 or lands as a new id (e.g. `15_patch_antenna_thick`). That is a science-governance call, not a mechanical one, so I left it to you. (`07` and `14` are free — no collision.)
2. **Role classification per study.** Registration needs an honest role for each: `14_rect_cavity` reads **claims-bearing** (exact analytic oracle, tight gates); `07_sheen_lpf` and the patch study read **diagnostic-reporter** on the same grounds cv05 was demoted (coarse-mesh discrepancy, rfx not the structure-faithful solver). Roles are restricted to those two by `test_crossval_manifest_forbids_research_only_cases`.

Options: (a) register all three under `validation/crossval/` with the classification above, resolving 05 explicitly; (b) keep them under `scripts/` but rename the directory so it does not read as the governed crossval set; (c) merge as-is and track registration as a follow-up issue. My recommendation is **(a)**, with the patch study taking a **new id** rather than overwriting cv05.

## Note on tone
The Palace finding is rfx-critical, not solver-critical: it reports that **our** coarse mesh distorts the doublet and that the earlier framing was **our** comparator artifact. That is the direction `feedback_humble_crossval_documentation` asks for, and the superseded framing is retained rather than quietly deleted.

R3: memory=feedback_humble_crossval_documentation + project_issue325_cv05_substrate_registration + feedback_gate_can_bind_artifact | R2-attempts=0 | falsifier=5% fixture perturbation trips test_all_blocks_passive (verified FAILED, then restored)


bk-squared:
Naming/registration review against the rule established by #359 and encoded in `tests/test_crossval_manifest_contract.py`. I proposed dispositions, then had them independently audited by a second agent (Codex) which read both the contract test and this PR's scripts; it **overturned one of my three proposals and found a blocker neither of us had flagged**. Below is only what survived that audit.

## Blocker first — the scripts are not fail-closed

`05_patch_antenna.py` and `07_sheen_lpf.py` print PASS/FAIL from `compare()` and then let `main()` fall through with **exit 0** (patch `:596-604`, `:637-655`; sheen `:477-500`). Governed `validation/crossval/05_patch_antenna.py` ends in `SystemExit(0 if all_ok else 1)` (`:824`).

As written, registering either script would put a case in the manifest that reports green regardless of whether its gate passed. That is the substantive prerequisite — more so than the naming question. `14_rect_cavity.py` is already fine here: its `<1%` TE101 / `<2%` higher-mode gates return exit 1 on failure (`:327-367`).

## The rule, restated precisely

| # | Rule | Where |
|---|---|---|
| R1 | `validation/crossval/*.py` glob must be **exactly 1:1** with `manifest.json` cases, and registered scripts must be unique | test `:99`, `:103` |
| R2 | `Path(case["script"]).stem == case["id"]` — **the filename stem IS the id** | test `:151` |
| R3 | `role ∈ {claims-bearing, diagnostic-reporter}` | test `:108` |
| R4 | Each case carries **exactly 16 keys**; each reference exactly `{kind, name, required_for_script_pass}`; `public_group ∈ {A, B, None}`; reference kinds ∈ {analytic, external-solver, self-invariant} | test `:133`, `:162`, `:120` |
| R5 | claims-bearing needs ≥1 of E2/E3/E4/E5; the `cpu_runner` order/exclusion **XOR applies to every case**, not only claims-bearing | test `:193`, `:204`, `:207` |
| R6 | Only cases with `public_group is not None` are cross-checked against `benchmarks.mdx`, mapping claims-bearing → "Validated comparison", else "Diagnostic only" | test `:346`, `:362` |

Measured on `main` today: 9 files, 9 cases, zero drift either way.

**Numeric prefixes are not stable identities** — they have been recycled. ID 07 held `07_dispersive_lorentz.py` (88cad69) and separately `07_multilayer_fresnel.py` (4ededb7); ID 12 held `12_openems_msl_notch_filter.py` (3bc2707), `12_patch_antenna.py` (dd52a93) and later `12_subgrid_disjoint_prototype.py` (8eb24fc); **ID 14 held `14_cavity_q_3d.py` (934691c)** — itself a cavity study — and `14_inverse_design_demo.py` (25845a4). So a bare `14_rect_cavity` is the most confusable name available for a cavity study, and disambiguating in the filename is worth doing.

## Proposed dispositions

**Patch → `15_patch_antenna_rt5880`** · `diagnostic-reporter` · `[E2, E4]` · group `B` · refs: analytic (Balanis), external-solver (canonical openEMS tutorial), external-solver (same-geometry openEMS) with **only the same-geometry reference** `required_for_script_pass=true`.

My first instinct was to merge this into the existing `05_patch_antenna` and I was wrong. The audit showed a genuine division of labour: governed 05 is FR4 on a graded-z substrate that rasterises to **two cells** and is explicitly demoted to integration/smoke with accuracy delegated to committed tests (`:16`, manifest `:90`); this PR's study is RT/Duroid 5880, 3.175 mm, **uniform four-cell**, with an rfx↔openEMS resonance envelope as its primary comparison (`:18-33`, `:58-69`). Those are different experiments and should not share an id. A new id also avoids reopening the #325/#378 delegation decision.

**Sheen → `07_sheen_lpf`** · `diagnostic-reporter` · `[E4]` · group `B` · refs: external-solver (openEMS), external-solver (Palace).

Not claims-bearing, and this follows the script's own text: the stopband is a double transmission zero, the single-argmin "first null" comparison is "largely a COMPARATOR ARTIFACT", and rfx is "the less structure-faithful solver on this doublet at this resolution" (`:3-14`). A study whose headline finding is that rfx loses is diagnostic. Evidence is E4 and **not** E2 — the referee test states "No analytic reference" outright (`test_sheen_lpf_palace_referee_gates.py:14-18`). Promotion later would need the mesh sweep that actually resolves the doublet.

**Cavity → `14_rect_cavity_pozar`** · `claims-bearing` · `[E2]` · group `A` · ref: analytic (Pozar).

Evidence is **E2, not E4** — there is no external solver; the oracle is the closed form `f_mnl = (c/2)·√((m/a)²+(n/b)²+(l/d)²)`, re-derived in-file with no producer-side helper import (`:14-21`, `:95-105`), and the evidence table lists "cavity resonance" explicitly under E2 while E4 requires a same-geometry external solver (`physics_validation_evidence_rule.md:22`, `:24`). claims-bearing is justified as a deliberately narrow E2 envelope: exact oracle, no meshing ambiguity, tight gates that fail closed.

## Two mechanical items

- **Result directories move with the scripts.** Both derive `RES_DIR` from `SCRIPT_DIR` (patch `:93`, sheen `:62`), so relocating the script silently changes where results are expected. Either move the committed result directories alongside or repoint `RES_DIR` deliberately and register the paths in `artifact_paths`. (My earlier framing of this as an R1 consequence was wrong — R1 says nothing about JSON paths; the coupling is in the scripts.)
- **One `benchmarks.mdx` row per public case**, with the classification R6 derives from the role, or the contract test reds.

Happy to be told a different id or role fits better — the parts I'd defend hardest are the fail-closed exit contract and the E2-not-E4 label on the cavity, since both are about a case advertising more than it demonstrates.

bk-squared:
Done, not just proposed — pushed `e2418fc` + `c4e8bc9` onto this branch (rebased onto `main` first; pre-rebase tip was `5aac821`). The three studies are now **inside the governance boundary** with the audited disposition applied.

## What changed

| was (`scripts/crossval/`, ungoverned) | now (`validation/crossval/`, registered) | role | evidence | group |
|---|---|---|---|---|
| `14_rect_cavity.py` | `14_rect_cavity_pozar.py` | **claims-bearing** | `[E2]` | A |
| `07_sheen_lpf.py` | `07_sheen_lpf.py` | diagnostic-reporter | `[E4]` | B |
| `05_patch_antenna.py` | `15_patch_antenna_rt5880.py` | diagnostic-reporter | `[E2, E4]` | B |

- The cavity is **E2, not E4** — the oracle is the in-file Pozar closed form; there is no external solver. Its `<1%` TE₁₀₁ / `<2%` higher-mode gates already failed closed and are untouched.
- The patch keeps a **new id**: it is a different experiment from governed `05` (RT/Duroid 5880, 3.175 mm, uniform four-cell substrate vs FR4 graded-z two-cell smoke), and patch *accuracy* remains delegated to `05`'s committed tests — the #325/#378 decision is not reopened.
- Result JSONs moved alongside (`_07_sheen_results/`, `_15_patch_results/` — underscore dirs, invisible to the `*.py` glob) and are registered in `artifact_paths`.

## The fail-closed fix (the blocker)

Both compare paths previously printed PASS/FAIL and fell through with exit 0. Now `main()` ends `return 0 if compare(...) else 1`, missing openEMS exits **2** (visible SKIP), and — because these are *diagnostic* cases — the gates lock the **committed characterization**, not an invented accuracy claim. For Sheen that includes the honest negative: the committed rfx leg is **not passive** (58/120 bins over column power 1.10, worst 1794.5 at 17.05 GHz), so the gate pins that footprint and the docstring says no rfx |S| magnitude from this study is quotable.

**Falsifier-tested, not assumed:** tampering the committed `rfx.json` leg flips the run to exit 1 with named failures (`C1 argmin null 2.27% > 1.0%`, `D1 non-passive bin count 45 ≠ 58`, …); restoring the leg returns exit 0. Both compare modes run in seconds against the committed legs — no solver needed.

## Verification

- `pytest tests/test_crossval_manifest_contract.py` → **10 passed** (12 cases, zero glob↔manifest drift, benchmarks rows cross-checked)
- `pytest -k "crossval or benchmark or sheen"` → 36 passed, 1 skipped
- `ruff` clean, `api reference surface: OK`
- `benchmarks.mdx` gained one row per case with the role-derived classification ("Validated comparison" for 14, "Diagnostic only" for 07/15)

One correction to my earlier comment: the manifest case shape is **15 keys**, not 16 — the contract test's set-equality is the authority and it is green.

bk-squared:
Follow-up from the quotability research (PR #468): the committed rfx leg's non-passivity is now root-caused — **record truncation, not the extractor and not the absorber**. The controlled matrix is in #468; headline: `num_periods=20` fails the new settling witness at **−24.7/−24.3 dB** (rule: −40), and at `num_periods=60` the witness reads −72 dB, the worst column-power pole collapses 1794→8.8, and **0.5–10 GHz lands fully inside the 1.05 passivity envelope** (max 1.042) with the remaining 16–20 GHz bins all flagged by the existing `reliable` mask.

Concrete option for this study (owner's call): regenerate the rfx leg at `num_periods ≥ 60` (~6 min CPU) and update the D-gates to the settled footprint (expected 44 bad bins / worst ~8.8). That would upgrade the study from "no rfx |S| quotable" to "passband quotable, stopband honestly masked". The current committed characterization stays valid as-is — it locks what was run — so this is an improvement, not a correction.

bk-squared:
`07a409d` — the Sheen rfx leg is **regenerated with the measured-correct recipe**, and the accuracy story changes materially. Everything below is measured, not argued.

## What the corrected leg shows

| quantity | committed (np=20, offset=20) | regenerated (np=60, offset=40, PR #468 pipeline) |
|---|---|---|
| settling witness | **−24.7/−24.3 dB (FAIL)** | −70.1/−70.6 dB |
| max column power | 1794.5 | **0.9891 (strict ≤ 1, enforced; raw excess recorded per bin)** |
| rfx argmin null | 7.218 GHz (9.6% from openEMS) | **7.874 GHz — 1.4% from openEMS** |
| Palace-referee structure distance | 4.64% | **2.54%**, with rfx now resolving **both** doublet members (7.036/7.844 vs Palace 7.032/8.048) |

So the "9.6% null disagreement" the original study reported — which the Palace referee had already called a comparator artifact — is now root-caused: **record truncation + probe near-field contamination**, not solver physics. With clean probes the rfx argmin sits on the ~8 GHz doublet member, and the lower doublet member matches Palace to 0.06%. `sides_with = openems` **stands** (0.66% < 2.54%) — the referee section of the fixture is re-derived and `test_committed_referee_matches_rederived` is green.

## Honesty boundary, unchanged in kind

This is still a diagnostic-reporter case and still makes no rfx accuracy claim. At dx = 200 µm the raw extraction exceeds the passive bound by up to 0.398 (81/120 bins > 0.05, concentrated 10–20 GHz) — that is the coarse-mesh envelope, now **bounded and recorded per bin** in `passivity_correction` instead of leaking into quoted magnitudes. The argmin null *position* and the passband/stopband *structure* are the readable outputs.

## Gate changes (falsifier-tested both directions)

Gate D is now an **evidence-chain lock** instead of a non-passivity-footprint lock: D0 fails closed if a leg lacks the witness/enforcement fields (i.e. was produced on a pre-#468 rfx), D1 strict bound, D2 settling witness, D3–D5 correction footprint, D6 max column power, plus a new C3 passband-mean lock so tampered S-data cannot pass on side-channel fields alone. Verified: clean run 17/17 exit 0; tampering `s21_mag` reds C3+D6; stripping the evidence fields reds D0.

Note the dependency: regenerating this leg requires the #468 pipeline (`rfx_provenance` recorded in the JSON); on a pre-#468 rfx the study warns and the gates fail closed rather than reading a leg that cannot prove it settled.

PR #468, opened 2026-07-26
feat(msl): settling witness + strict passivity enforcement — S-parameters quotable, |S| <= 1 at every frequency

bk-squared:
## Why

The PR #462 Sheen-1990 study committed an rfx leg whose S-matrix is wildly non-passive — 58/120 bins over column power 1.10, worst **1794.5** — and nothing in the MSL path could say *why*, so the study had to quarantine every rfx |S| magnitude as unquotable. This PR makes the dominant mechanism measurable and loud.

## Root cause, separated by a controlled matrix (dx = 200 µm Sheen LPF, public `compute_msl_s_matrix`)

| variant | CPML layers | num_periods | bad bins (>1.10) | worst colpow |
|---|---|---|---|---|
| committed config | 8 | 20 | 56/120 | 62.6 @ 18.03 GHz |
| absorber ×3 | **24** | 20 | 57/120 | 32.2 @ 18.03 GHz |
| record ×3 | 8 | **60** | 44/120 | 8.75 |
| both | 24 | 60 | 45/120 | 6.27 |
| record ×6 | 8 | **120** | 44/120 | 8.77 |
| record ×12 | 8 | **240** | 44/120 | 8.77 |

- **Record length is the mechanism; absorber depth is not.** Tripling CPML at fixed record leaves the poles in place; tripling the record collapses them 62.6 → 8.75, and ×6/×12 are bit-stable — converged at num_periods ≈ 60.
- The worst pole also moved 17.05 → 18.03 GHz under a ppm-level dx rounding difference between the committed leg and the reproduction — the signature of an interference artifact, not a physical feature.
- Mechanically: S11 and S21 share the incident-wave denominator `a = (V + Z0·I)/2`; single-bin DFTs of a record that ends while the LPF stopband is still ringing corrupt `a(f)` and produce shared poles.
- The witness on the committed configuration measures **−24.7 / −24.3 dB** (per driven run) — far above the −40 dB rule — and on num_periods=60 it measures **−72.9 / −71.8 dB**.

## What this adds

CLAUDE.md's ring-down settling rule, made mechanical for the MSL path:

- One point Ez time-series probe per port at the probe-0 plane during the driven-run scan (restored afterwards; `self._probes` was previously not stashed — it is now).
- `MSLSMatrixResult.settling_db` — per driven run, worst over ports of `10·log10(end-10%-mean Ez² / peak Ez²)`. Optional field, default `None`, backward compatible.
- A loud aggregate warning when any run exceeds −40 dB, quoting the per-run values and the actionable knob (`num_periods`), emitted before the existing passivity self-check.
- On the `eps_override` AD path the witness is skipped (NaN) when `time_series` is a tracer rather than concretising it.

## What "quotable" now means for MSL S-parameters (fully mechanical)

1. `settling_db < −40` per driven run (**this PR**),
2. `reliable[i, k]` true (existing standing-wave-null mask — on the settled Sheen run it flags **all 15** remaining bad bins in 16–20 GHz),
3. column power within the documented 1.05/1.10 envelope (existing self-check warning).

On the settled Sheen leg that leaves 0.5–10 GHz fully inside the 1.05 envelope (max 1.042) — the study's passband claims become quotable — while 16–20 GHz stays honestly excluded by `reliable`.

**Characterized residual, not chased:** 29 bins in 10–16 GHz sit at colpow 1.10–1.27 on a fully settled record, unflagged by `reliable`. cv06b at dx = 50 µm is passive ≤ 1.05 over its band, so this is consistent with the documented coarse-mesh (4-cell substrate, ~20–27 % staircase-Z0) envelope at dx = 200 µm rather than an extractor defect. Stated here so nobody reads the witness as clearing that band.

## Follow-up this enables (study owner's call, on #462)

Regenerate the committed Sheen rfx leg at `num_periods ≥ 60`: expected footprint 44 bad bins / worst ~8.8 instead of 58 / 1794, with the characterization gates updated to the settled record.

## Verification

- 4 new tests (`tests/test_msl_settling_witness.py`): truncated record fails loudly with the measured value and knob in the message; settled record passes deeply (< −60 dB); witness probes do not leak; the result field is optional.
- MSL-affected selection: 54 passed, 1 skipped, 1 xfailed. AD-path selection: 25 passed.
- `ruff` clean, `api reference surface: OK`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ SemVer — **BREAKING** entries are flagged in upper-case.
 
 ## [Unreleased]
 
+### Changed — `compute_msl_s_matrix` returns a passivity-enforced S by default (**BEHAVIOUR CHANGE**)
+
+- The returned `MSLSMatrixResult.S` now satisfies `||S(f)||_2 <= 1` at every
+  frequency: singular values are clipped to the passive bound per bin
+  (nearest passive matrix in spectral norm). Nothing is discarded — the raw
+  extraction is preserved in the new `S_raw` field, the per-bin clip amount
+  in `passivity_correction`, and a warning names the touched bins. Bins with
+  a large correction are measurement artifacts (check `reliable` /
+  `settling_db`); the projection bounds them, it does not certify them. Pass
+  `enforce_passivity=False` for the previous raw behaviour. EXEMPTION: on
+  the `eps_override` channel (traced or concrete) the projection is skipped
+  so finite-difference and `jax.grad` objectives see the same raw function.
+- New ring-down settling witness: `MSLSMatrixResult.settling_db` records the
+  worst end/peak `Ez^2` ratio (dB) over all port probe planes per driven
+  run, and a run above −40 dB warns loudly — a fixed `num_periods` record
+  that ends while the structure is still ringing produces truncation-artifact
+  S-parameters (measured: column-power poles up to ~1.8e3 on the Sheen-1990
+  LPF at `num_periods=20`).
+
 ### Changed — cross-validation scripts moved to `validation/crossval/` (repo layout only; no behaviour change)
 
 - `examples/crossval/` moved to `validation/crossval/` so that `examples/`
--- a/docs/guides/api_symbol_inventory.json
+++ b/docs/guides/api_symbol_inventory.json
@@ -2444,7 +2444,8 @@
       "strict_extractor",
       "eps_override",
       "checkpoint_every",
-      "checkpoint_segments"
+      "checkpoint_segments",
+      "enforce_passivity"
     ],
     "compute_waveguide_s_matrix": [
       "n_steps",
--- a/docs/guides/simulation_methodology.md
+++ b/docs/guides/simulation_methodology.md
@@ -116,6 +116,22 @@ where supported, then verify the actual observable separately. Closed PEC
 cavities do not decay through an absorber; use a fixed run and inspect the
 probe signal.
 
+**The −40 dB ring-down settling rule.** For a fixed-length record on an
+open (absorbing) domain, quote the end-of-record energy relative to the
+post-source peak before quoting any claims-bearing DFT-derived number
+(S-parameters, Harminv frequencies): `10·log10(end-window energy / peak
+energy)` must be below **−40 dB**, or the record was truncated while the
+structure was still ringing and the DFT integrates a cut transient —
+measured consequence on a resonant microstrip filter: spurious |S| poles up
+to column power ~1.8e3 that shrink monotonically with record length.
+`compute_msl_s_matrix` computes this mechanically as
+`MSLSMatrixResult.settling_db` (worst end/peak `Ez²` ratio over all port
+probe planes per driven run — multiple planes because a single plane is
+standing-wave-node sensitive, with a measured 18 dB spread across planes on
+one under-settled record) and warns when a run exceeds the line. The rule
+does not apply to closed PEC domains (energy conservation is correct there)
+or to CW drive.
+
 ## 7. Interpret the requested observable, not a convenient proxy
 
 - Remove source-dominated transients before applying Harminv or an FFT to a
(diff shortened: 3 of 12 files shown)


bk-squared:
Second commit (`ba72a47`) completes the directive: **the returned S now satisfies ‖S‖₂ ≤ 1 at every frequency, strictly** — measured `max σ = 0.9999995` over all 120 Sheen bins.

## The elimination chain that justified enforcement (all measured)

| hypothesis | test | verdict |
|---|---|---|
| record truncation | num_periods 20→60→120→240 | **dominant below np≈60** (poles 62.6→8.75, then bit-stable) — fixed by the witness (commit 1) |
| absorber depth | CPML 8→24 at fixed record | no effect on the poles |
| extraction formulation | V·I split vs fitted-Zc(f) vs pure α/γ N-probe fit, offline on dumped phasors | no improvement at 16–20 GHz; fitted-Zc variants **explode at low f** (colpow → 330; the fit is ill-conditioned when probe span ≪ λ) |
| excitation spectrum | broadband drive, in-band floor −20 dB vs default −39 dB | footprint identical to 3 decimals — S is a stable ratio; the fields genuinely contain it |
| reciprocity asymmetry | symmetrize before projecting | σ changes by 0.001 — not the cause |
| **probe placement** | n_probe_offset 20→40 cells | **16–20 GHz colpow 8.75 → 1.81** — near-field/evanescent contamination at the probe planes dominates the residual; recipe evidence for a future frequency-aware offset default |

What remains after settling + placement is mesh-class field error no reformulation removes — so the strict bound is delivered as **explicit constraint enforcement** (per-frequency singular-value clipping to the nearest passive matrix, a few ULPs under 1 so the bound survives the f32 round-trip), with nothing hidden: `S_raw` preserved, `passivity_correction` per bin, a warning naming the touched bins, and the raw-value self-check still auditing `S_raw` so the projection can never silence the artifact diagnosis.

**AD path**: the projection is skipped under `eps_override` tracing — measured, `min(σ,1)` flipped the committed d|S|²/d-eps sign gate at actively-clipped bins, and gradient fidelity is the standing non-negotiable. Quoting is a concrete-path activity; the strict bound applies exactly where quoting happens.

Verification: 13 witness+enforcement tests, AD end-to-end gate green, MSL selection 72/72 relevant green (replay golden now explicitly pins the raw assembly), ruff clean, `api reference surface: OK` (inventory regenerated for the new parameter).

bk-squared:
## Accuracy demonstration (PI directive: runtime doesn't matter, correctness does)

Fresh **cv06b dx = 50 µm** run (1.6 M cells, 68 min CPU) through the full witness+enforcement pipeline, geometry reconstructed to the committed openEMS fixture's dimensions (verified to the µm against its meta before running), `num_periods=20` — i.e. the committed fixture's own configuration:

| check | result | committed gate |
|---|---|---|
| settling witness | **−112.1 / −113.9 dB** | rule < −40 |
| notch frequency vs committed openEMS | 3.627 vs 3.429 GHz → **5.80%** | ≤ 7% |
| off-notch \|S21\| diff (2.5–6 GHz) | mean **0.0925**, max **0.2076** | ≤ 0.13 / ≤ 0.25 |
| notch depth | −30.7 dB | < −20 dB |
| strict bound | max column power = 1.0000 (enforced) | ≤ 1 |
| enforcement corrections | max **0.080** | — |
| vs committed rfx leg | max Δ\|S21\| = **0.040** | — |

Three things this establishes:

1. **The witness discriminates rather than moralizes.** The same `num_periods=20` that failed on the Sheen LPF at −24.7 dB (truncation poles to 1794) measures **−112 dB** here — the notch structure genuinely settles in 20 periods. The witness is a per-structure measurement, not a blanket rule about run length.
2. **The enforcement correction is a mesh-quality meter, now with two points.** Max clip **0.398 at dx = 200 µm** (Sheen) vs **0.080 at dx = 50 µm** (cv06b). The projection does almost no work on a converged mesh — exactly the behaviour a constraint enforcement should have.
3. **Accuracy is preserved through the new pipeline.** All four committed accuracy gates pass on the fresh run, the fresh leg reproduces the committed leg within 0.04 in \|S21\|, and the notch lands at 3.627 GHz — **0.1% from the Palace FEM referee's 3.631 GHz** (`msl_stub_notch_palace_referee.json`).

Together with the Sheen regeneration on #462 (argmin 7.218 → 7.874 GHz, 1.4% from openEMS; Palace structure distance 4.64% → 2.54% with both doublet members resolved), the quotability chain is now demonstrated end-to-end on both committed MSL studies: **settle (witnessed) → place probes clear of the near field (measured) → enforce the bound (recorded) → quote**.

Issue #467, opened 2026-07-25
MeshShape (CAD import) has no top-level export, no API-inventory entry and no public-docs row


bk-squared:
CAD import landed recently (#453 / #456, closing #358) and works, but it is a second-class citizen next to the other geometry primitives. Verified on `main`:

```
$ python -c "import json; d=json.load(open('docs/guides/api_symbol_inventory.json')); print('MeshShape' in d['rfx_exports'])"
False
$ python -c "import rfx; print(hasattr(rfx, 'MeshShape'))"
False
$ python -c "from rfx.geometry import MeshShape; print('ok')"
ok
$ grep -rl MeshShape docs/public/
(no matches)
```

By contrast `Box` / `Cylinder` / `Sphere` / `PolylineWire` are top-level `rfx` exports and appear in the pinned inventory.

So a user who wants to simulate geometry drawn in CAD has to find `rfx.geometry.MeshShape` from the source or from `examples/tutorials/cad_mesh_import_demo.py`; nothing in the public API surface or the docs points there. `tests/test_mesh_import.py` covers the behaviour, so this is a discoverability/governance gap, not a correctness one.

## Two things worth deciding together

1. **Export parity + docs.** If `MeshShape` is meant for routine use, it wants a top-level export, an `api_symbol_inventory.json` row (regenerate with `python scripts/check_api_reference.py --write`) and a public-docs row alongside the other primitives.
2. **State the AD limitation wherever it lands.** `MeshShape` rasterises host-side via `trimesh` point-in-mesh containment and raises on a traced coordinate (`rfx/geometry/mesh_import.py:147-153`), so a CAD-sourced design **cannot carry gradients**. Given the project's standing position that AD-traceability is the non-negotiable tie-breaker, that limitation belongs next to the feature, not only in the source.

Context: this surfaced while writing the design-interop layer (#463), which **refuses** `MeshShape` rather than degrading it — a triangle mesh is not parameter-describable the way a CSG primitive is, so exporting it as a bounding box would describe a different structure. That refusal is the right behaviour, but it means the export/import story and the CAD-import story currently do not meet.

Issue #466, opened 2026-07-25
Subgrid PML-overlap warning reads the absorber as inside the domain; the grid builder puts it outside


bk-squared:
Two places in the codebase use **opposite frames** for where the CPML absorber lives relative to the user domain. Found while establishing the absorber arithmetic for the openEMS emitter (#463), where getting this wrong buries the structure in the absorber.

## The grid builder: absorber is OUTSIDE the user domain

`rfx/grid.py:133` documents the convention, and it measures out:

```python
Grid(freq_max=10e9, domain=(0.01, 0.01, 0.01), dx=1e-3, cpml_layers=N).shape
#   N=0  -> (11, 11, 11)      = the user domain
#   N=8  -> (27, 27, 27)      = 11 + 2*8
#   N=16 -> (43, 43, 43)      = 11 + 2*16
```

`axis_pads` is the leading pad and `(idx - axis_pads[ax]) * dx` recovers user coordinates, i.e. index 0 lies **outside** the domain. So `domain` is entirely physical and the absorber is extra cells beyond it.

## The subgrid warning: absorber is INSIDE the domain

`rfx/api/__init__.py:627-650` computes `pml_zlo_thickness = face_layers[...] * dx` and then tests

```python
if z_lo < pml_zlo_thickness or (domain_z > 0 and z_hi > domain_z - pml_zhi_thickness):
```

That is only meaningful if the absorber occupies the first and last `N` cells **of the user domain** — the opposite of what the grid builder does. Under the grid builder's frame, a subgrid spanning the full domain never overlaps the absorber, so this warning appears to fire on geometry that is in fact clear of it.

## Severity / scope

Low priority and **not** something I would chase: the subgrid path is experimental and parked (3-D SBP-SAT falsified, PR #90), so the warning guards a lane nobody should be using. Filing it because (a) it is a genuine internal inconsistency about a quantity that *is* load-bearing elsewhere, and (b) anyone reading that block to learn the convention will learn the wrong one — which is how the emitter nearly got it backwards.

No fix proposed; if the subgrid lane is ever revived, the check should be rewritten against `Grid.face_pads` rather than against `domain`.

Issue #465, opened 2026-07-25
openEMS excitation guard threshold uf_inc <= 1e-9 would false-fire: correct runs measure 1.9e-12


bk-squared:
Measured while building the openEMS emitter (#463).

`scripts/diagnostics/build_coaxial_line_openems_broad_comparison.py:214-223` treats a small incident-wave amplitude as "the excitation never entered the grid", using a threshold of `1e-9`. That guard exists for a real failure mode — **openEMS silently drops an off-grid excitation**, and the run then completes and returns plausible-looking S-parameters — so the guard is worth having. The threshold is the problem.

## Measurement

Peak incident amplitude on runs that produced **correct** S-parameters:

| case | `incident_peak_abs` |
|---|---|
| coarse two-port PEC cavity | **1.93e-12** |
| MSL thru, 81×57×29 | **1.79e-13** |

Both are 3–4 orders of magnitude *below* `1e-9`, so a guard at that threshold would reject working runs. (These were emitted setups, but the amplitude scale is a property of openEMS's excitation normalisation, not of how the script was written.)

## Suggested shape

Test for **zero / non-finite** rather than a magnitude, and add an independent witness that the port actually saw a time-varying field — the emitter in #463 uses the port voltage/current time-trace peak alongside the incident peak, so a dropped port is caught without assuming an absolute scale.

Worth checking whether any other comparator script carries the same constant before it is copied again.

## Related trap, same failure surface

openEMS logs `Unused primitive (type: Box) ... <name>` in **two** unrelated situations: benignly, when a design's own trace already covers a port span (both PEC, no effect); and as the symptom of the port-dropped failure this guard is for. The log line is identical, so it cannot be used to discriminate — which is exactly why a numeric witness is needed.

PR #90, opened 2026-05-27
feat(subgrid): integrate energy-stable non-split SBP-SAT subgridding

bk-squared:
# PR Description: Stable and Physically Verified 2D SBP-SAT Non-split Subgridding

This PR integrates the JAX-native 2D TM SBP-SAT non-split subgridding solver into `rfx`. Rather than relying solely on passing unit tests, this implementation is fully backed by rigorous physical wave propagation and centerline conservation checks.

## 🔬 Core Breakthroughs & Physical Discoveries

During physical verification, we identified and resolved three major staggered-grid scaling and boundary derivative errors that previously caused severe phase delays, massive spurious boundary reflections, and wrong-direction coupling:

1. **Point Source Area-Scaling ($\text{ratio}^2 = 9$)**:
   - In FDTD, a soft source injected into a single cell represents a spatial Dirac delta function whose discrete amplitude scales as $1/\Delta x^2$. Since the reference grid has cell area $\Delta x_f^2 = \Delta x_c^2 / 9$, injecting the same raw pulse directly into different resolutions leads to a massive $9\times$ energy mismatch.
   - **Fix**: We correctly scaled the subgrid source injection by $\text{ratio}^2 = 9$, aligning the reference wave peak ($0.5447$) perfectly with the coarse wave peak ($0.5399$).

2. **Physically Normal-Consistent Interface Signs**:
   - Boundary penalty terms in SBP-SAT must flip signs between opposite faces (West vs. East, South vs. North) because their physical outward normal vectors point in opposite coordinate directions.
   - **Fix**: We implemented the normal-consistent configuration (`W:+1, E:-1, S:+1, N:-1`) as the only valid physical configuration.

3. **Yee Boundary Derivative Staggering Scale ($2/dx \to 1/dx$)**:
   - In collocated solvers, the boundary derivative is scaled by $2/dx$ due to the boundary weight $dx/2$. But on a staggered Yee grid, the spatial derivative of the magnetic field at the boundary electric field node naturally spans a full cell width $dx$ (from $i-0.5$ to $i+0.5$). Using $2/dx$ at the Yee boundaries physically modeled a grid with half the cell size, causing a huge non-physical delay of **150 steps** and severe spurious oscillations (min value reaching $-1.08$ on a peak of $0.56$).
   - **Fix**: We corrected the boundary derivatives to their physically correct staggered scale of $1.0/dx$. This completely eliminated the delay and wild boundary noise.

---

## 📊 Quantitative Waveform Cross-Validation

[...]

PR #473, opened 2026-07-27
hygiene: MeshShape export parity, scale-free openEMS excitation guard, subgrid PML-warning frame fix (#465 #466 #467)

bk-squared:
Track C of the 2026-07-27 delegated-tracks handover (`docs/research_notes/20260727_handover_delegated_tracks.md`): three small independent hygiene items, one commit each.

## #467 — MeshShape export parity (`e6d581d`)
Top-level `rfx.MeshShape` export + regenerated `api_symbol_inventory.json` (diff = the one row) + public-docs rows (primitive table + guide section). Both docs state the AD limitation next to the feature, per the issue: host-side trimesh rasterization, traced coordinates raise, CAD geometry cannot carry gradients, `rfx.interop` refuses it rather than approximating. Checked that `rfx/geometry/mesh_import.py` keeps its trimesh import lazy, so plain `import rfx` still works without the `cad` extra.

## #465 — openEMS excitation guard (`df73ba6`)
The coax broad-comparison guard's `max|uf_inc| <= 1e-9` threshold false-fires: correct runs measure 1.93e-12 / 1.79e-13 (#465 measurement table — openEMS's excitation normalisation scale). Replaced with the scale-free shape the issue proposes, ported from the PROVEN D5 guard in `rfx/interop/emitters/openems.py`: exactly-zero / non-finite incident spectrum → raise, plus the independent port voltage-time-trace witness (`U_filenames`; identically-zero trace = silently dropped port). No absolute scale assumed.

## #466 — subgrid PML-overlap warning frame (`17de70e`)
The warning assumed the absorber occupies the first/last N cells INSIDE the user domain; the grid builder puts it OUTSIDE (domain fully physical, CPML = padding — `rfx/grid.py`). It fired on subgrids that were clear of the absorber (e.g. a full-domain `z_range`). Now warns exactly when `z_range` leaves the physical domain into an absorbing face's padding, respecting per-face boundary kinds. **Frame fix only** — subgrid stays EXPERIMENTAL (PR #90), no new validation claims. Fire/clear/PEC-face test triplet added; no existing test pinned the old wording.

## Verification (local, worktree, `rfx.__file__` checked)
- new tests: 3/3 (`test_subgrid_pml_overlap_warning.py`)
- contract/regression: ad-surface + api + mesh-import + golden-workflow **62 passed**; subgrid public-API + n-1 guard + api **52 passed**
- `api reference surface: OK`; ruff clean (gated scope)
- #465 has no pytest surface (diagnostics script, needs openEMS): guard shape is the emitter's tested D5 pattern, cited in-code.

R3: memory=feedback_contract_tests_for_public_surface + project_sbp_sat_subgridding_research + feedback_imitate_external_solver_examples | R2-attempts=0 | falsifier=fire/clear test pairs + api-reference drift gate

Independent review before merge per the handover process; do not merge on author's word alone.

### DIFF
--- a/docs/guides/api_symbol_inventory.json
+++ b/docs/guides/api_symbol_inventory.json
@@ -145,6 +145,9 @@
     "MeshPlan": {
       "kind": "class"
     },
+    "MeshShape": {
+      "kind": "class"
+    },
     "ModulatedGaussian": {
       "kind": "class"
     },
--- a/docs/public/api/geometry-materials.mdx
+++ b/docs/public/api/geometry-materials.mdx
@@ -21,6 +21,7 @@ listed in [Recommended Configuration](/rfx/validation/reference-lane/) and
 | `PolylineWire` | polyline-defined wire geometry | wire-like studies and routed conductors |
 | `Via` | `Via(center=..., drill_radius=..., pad_radius=..., layers=..., material="pec")` | PCB-style interconnects |
 | `CurvedPatch` | `CurvedPatch(center=..., length=..., width=..., radius=..., axis="x")` | curved-conductor studies; constructor arguments are keyword-only |
+| `MeshShape` | `MeshShape.from_file(path, scale=..., translate=(0,0,0))` | CAD import — STL/OBJ/PLY, plus STEP/STP via the `cad` extra. Rasterizes host-side (trimesh containment): NOT differentiable — a traced coordinate raises, so CAD geometry cannot carry gradients |
 
 ## Material registration
 
--- a/docs/public/guide/materials-geometry.mdx
+++ b/docs/public/guide/materials-geometry.mdx
@@ -193,6 +193,25 @@ rod  = Cylinder((0.010, 0.025, 0.000), radius=0.001,  height=0.050, axis="x")
 sim.add(via, material="copper")
 ```
 
+### MeshShape (CAD import)
+
+```python
+from rfx import MeshShape  # needs: pip install "rfx-fdtd[cad]"
+
+# STL / OBJ / PLY natively; STEP / STP via the cad extra's OpenCASCADE backend.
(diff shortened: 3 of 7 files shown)


bk-squared:
**Review response (`4bf7cc1`)** — REQUEST-CHANGES addressed; re-review requested.

**BLOCKING (#466 boundary-touch regression): confirmed and fixed.** I reproduced the red test first (`test_production_boundary_terminated_rejects_refined_face_touching_cpml` — FAILED on `17de70e`, exactly as reported). The condition is now `z_lo <= 0.0` / `z_hi >= domain_z`: **touching an absorbing face warns** (consistent with the static validator's `subgrid_overlaps_absorber` verdict on the identical geometry, which the review verified never regressed), while the interior false-positive class the frame fix targets stays silent. The message keeps the pinned "overlaps PML" wording. My test file now carries the 4-case battery: interior no-warn / boundary-touch warn / into-padding warn / PEC-faces silent.

Re-run after the fix: previously-red pinned test **green**; full `-k "subgrid or refinement"` surface **132 passed, 1 skipped, 2 xfailed, 0 failed** (reviewer measured 130+1F on the same selection pre-fix; +2 = the added tests).

**Root cause of the false verification claim, owned:** the "no existing test pinned the old wording" grep was piped through `| head -4` and the decisive hit was the 5th output line. The commit message of `4bf7cc1` records this; the general lesson (never truncate a must-be-empty existence check) goes to durable memory.

**#465 MINOR**: witnesses now surfaced as a per-term log line (`max|uf_inc|` + `trace_peak`), mirroring what the D5 guard records in its diagnostics JSON. Thanks also for the `U_filenames`/`CalcPort` unreachability analysis of the `trace_peak=None` path — agreed, no change needed there.

**#467**: no changes (review found none needed).

Issue #477, opened 2026-07-27
test_msl_ad_fd_converged_tight is red on main: rel_err 0.1113 vs 0.10 gate (deterministic, slow-lane-only)


bk-squared:
Found during the #469/#470 track's pre-merge verification. The slow-marked converged AD-vs-FD gate fails on **clean main** (`c459fc0`, isolated worktree, CPU):

```
[MSL-FD-TIGHT] g_ad = -2.1104e-01
[MSL-FD-TIGHT] g_fd = -2.374649e-01  (FD wall-time: 728.6s, h=0.001)
[MSL-FD-TIGHT] rel_err = 0.1113  (threshold: 0.10)
```

A/B evidence: the same test on the #469/#470 branch produces **bit-identical numbers** (rel_err 0.1113, same g_ad/g_fd to all printed digits), so that PR is numerics-neutral here and this red predates it. Deterministic (two runs, identical values), an 11.1% vs 10% threshold miss with sign agreement — not a flake.

Because it is `slow`-marked it is invisible to PR CI; unknown how long it has been red. Candidate classes (unattributed — needs a root-cause session, NOT a tolerance nudge per the no-silent-gate-loosening rule):
- environment drift (JAX version class — see `project_env_jax_pin_drift` pattern: pod-restart pulls newer JAX),
- FD step-size h=0.001 discretization of the central difference at this converged point,
- a real gradient-accuracy drift on main since the gate was set (2026-05-25, MSL-FD-TIGHT).

Cheap first checks for whoever picks this up: run at the last-known-green SHA in the same env (separates env-drift from code-drift); sweep FD h one decade each way (separates FD-discretization from AD error).

Environment: CPU-only, `jax 0.6.2`, isolated worktrees, `rfx.__file__` verified. ~35 min/run.

bk-squared:
## Root cause found (all measured, main @ `98c5e33`, pre-declared falsifier tree)

**Verdict: not an AD regression, not env drift, not code drift — the FD *comparator* carries an f32 evaluation-noise envelope that straddles the 0.10 gate across platforms, and the gate is green on its owner platform (GPU).**

### E2 — owner-platform check (VESSL run 369367249574, rtx4090, today's main)
```
GPU: g_ad = -2.1105e-01   g_fd = -2.3079e-01   rel_err = 0.0855  PASS
CPU: g_ad = -2.1104e-01   g_fd = -2.3746e-01   rel_err = 0.1113  FAIL
```
[...]

Issue #470, opened 2026-07-26
MSL witness probes pollute the preflight record (~14 advisories/run, 8 self-inflicted) and double-fire the settling witness


bk-squared:
Deferred from the PR #468 review (finding 8, re-measured at the multi-plane witness).

`compute_msl_s_matrix` now registers `n_probes` internal Ez witness probes per port (5/port on the thru fixture). Two consequences:

1. **Preflight self-noise.** The driver's internal `run()` auto-preflights, and the library's own diagnostic probes trigger the probe-placement advisories (`absorber_overlap`, near-CPML): a call that previously reported 6 advisory issues now reports ~14, of which 8 are the library warning about itself — and they bury the two genuine MSL port-clearance advisories. Since preflight output must be quoted verbatim next to every claims-bearing number, that record is now mostly self-noise.
2. **Two differently-defined settling numbers per run.** With `self._probes` non-empty, the pre-existing #332 probe-tail advisory (5%-tail max-amplitude) also fires alongside the new `settling_db` witness (10%-tail mean-power over all port planes, worst plane). Both are honest; one should be *the* witness and the other silenced or renamed for this path.

Cheapest containment per the reviewer: an `internal` flag on `_ProbeEntry` that the probe-placement preflight checks and the #332 advisory skip — NOT `skip_preflight`, which would silence genuine user-facing checks.

Also named in the same review as follow-ups (same class, lower priority): `scripts/diagnostics/msl_thru_meshconv.py:139` and `msl_thru_mesh_convergence.py:65` should pass `enforce_passivity=False` — clipping coarse-mesh points compresses the |S|-vs-dx trend the studies exist to quantify. A general rule worth adopting: everything under `scripts/diagnostics/` sees the raw extraction; only user-facing paths take the projection default (PR #468 already applied this to the E5 sweep, the openEMS comparator, and the replay-fixture capture).

Issue #469, opened 2026-07-26
MSL n_probe_offset auto-default has no downstream-reflector term — the two clearances are mutually unsatisfiable on short-feed geometries


bk-squared:
Found during the PR #462 Sheen-leg regeneration and its independent review.

`add_msl_port`'s auto-default is `n_probe_offset = max(3, _lam_cells, _hsub_cells)` (`rfx/api/__init__.py:1368-1371`), where `_hsub_cells = 5·h_sub/dx` — the issue-#80 upstream fringing clearance. It has **no notion of what sits downstream**, and the preflight advisory that checks the downstream λ_g/4 reflector clearance recommends "bump n_probe_offset" — which on a short feed moves the probes **toward** the reflector and makes that clearance worse.

## Measured on the Sheen 1990 LPF (dx = 200 µm, feed length 10 mm)

| offset (cells) | upstream ≥ 5·h_sub | downstream ≥ λ_g/4 | 16–20 GHz col-power (near-field contamination) | in-band median Re(Z₀) |
|---|---|---|---|---|
| 20 (auto default) | ✓ (exactly) | ✓ (5.2 mm) | **8.75** (contaminated) | 48.7 |
| 30 | ✓ (6 mm) | ✓ (3.2 mm) | ~1.8 | **49.8** |
| 40 | ✓ (8 mm) | **✗ (1.17 mm < 1.68 mm, preflight-warned)** | 1.81 | 57.4 (biased) |

So on this geometry a compliant-and-clean window exists (offset ≈ 30), but neither the default nor the advisory's suggested direction finds it: the default sits at the contaminated near edge, and following the advisory overshoots into the reflector clearance.

## Suggested shape

Make the auto-default (or at least the preflight advisory) solve the interval, not one edge: `offset_min = max(3, λ/(4π·dx), 5·h_sub/dx)` and `offset_max = (d_reflector − λ_g/4 − probe_span)/dx`, warn when the interval is empty (the honest "this feed line is too short for a clean measurement" case), and pick the midpoint otherwise. The reflector distance is already computed by the existing preflight check, so the pieces exist.

Related evidence: PR #468 (near-field contamination measurement), PR #462 `07_sheen_lpf.py` (the committed leg's recipe comment records the interval reasoning).

PR #478, opened 2026-07-27
fix(msl): probe-offset interval solve + internal-witness-probe isolation (#469, #470)

bk-squared:
Track B of the 2026-07-27 delegated-tracks handover (`docs/research_notes/20260727_handover_delegated_tracks.md`). Reconnaissance by a 5-agent read-only workflow whose findings redirected the design twice (details below); every number is measured in this session.

## #469 — the auto `n_probe_offset` now solves BOTH clearances
The auto-default was the upstream-only lower edge, and the reflector advisory said "bump n_probe_offset" — which moves probes **toward** the reflector (the #469 Sheen table: default sat at the contaminated near edge, colpow 8.75; the advisory's direction overshot the far edge).

**Fix** (`_resolve_msl_auto_offsets`, driver-time — geometry registered after `add_msl_port` is invisible at add time):
- no downstream reflector → stored lower edge, **byte-identical** to the pre-#469 default (open thru lines unchanged — verified no-op on the `_thru` fixtures and on `test_msl_ad_fd_converged`'s geometry);
- finite reflector → **midpoint** of `[offset_min, offset_max]`. Sheen parameters: [20, 33] → **26**, pinned in-test against independent hand arithmetic;
- interval empty → **loud warning** ("feed too short for a clean N-probe measurement") + upstream-priority fallback, replacing the silent compromise. Explicit offsets never adjusted; NU keeps stored values.

**Prerequisite comparator fixes** in the reflector walk (recon findings, now shared by preflight check 4 and the solve via `msl_nearest_downstream_reflector`):
- own-trace false positive: a `-x` port's OWN feed trace read as a reflector at 0 µm (cv07 p2 known-residual) — the 80%-extent heuristic is replaced by feed-containment discrimination;
- latent `+x` far-wall arithmetic bug (`x_far` evaluated to the CPML *thickness*) — line removed, moot under the new discrimination;
- check 4 now measures the TRUE deepest probe (`offset+(n_probes−1)·spacing`, was the 3-probe V3 span, 2·spacing short for the default 5 probes). Advisory reworded to interval language — pinned tokens kept, advisory suites 62/62 green with no retuning needed.

## #470 — library witness probes out of the user record
Measured before/after on the 2-port thru: **14 advisories/driven run → 4**, "Probe at …" self-advisories **10 → 0**; the remaining 4 are all genuine (2× MSL port-clearance, lossless dielectric, PEC z-extent). #332 no longer double-fires next to `settling_db`.

Containment = `Simulation._internal_probe_indices` runtime bookkeeping — deliberately **not** an `_ProbeEntry` field: recon showed the entry field set is pinned by the interop design-IR contract (3 test files + `additionalProperties:false` schema), so the issue's suggested flag was not the cheapest containment. Probe-placement advisories and #332/#336 skip internal indices only; **user probes keep pre-#470 behavior exactly** — the `test_issue80_patch_s11_regression` #332 fail-safe (a silent-disarm trap recon flagged) still evaluates user columns, with a fire/clear pair proving it.

**Interop representational contract** (the census test forced an explicit decision): the design document records an AUTO port's **resolved** offset as a frozen explicit value (dump-time solve, idempotent, original never mutated) — a rebuilt design reproduces the original's probe placement exactly. Census exemption documented in `assert_designs_equivalent`; the behavior pinned positively by `test_auto_msl_offset_is_frozen_resolved_in_the_document`.

Also, per the issue's "diagnostics see the raw extraction" rule: `msl_thru_meshconv.py` / `msl_thru_mesh_convergence.py` now pass `enforce_passivity=False`.

## Verification (isolated worktree, `rfx.__file__` checked)
- new batteries: `test_msl_probe_offset_interval.py` (6) + `test_msl_internal_probe_advisories.py` (2), incl. end-to-end wiring and fire/clear pairs
- final fast batch across MSL/preflight/contract surfaces: **116 passed, 1 xfailed**; interop **213 passed**; witness+passivity 12; advisory suites 62
- slow: `test_msl_port_integration` passed; `api reference surface: OK`; ruff clean
- **Sheen committed leg untouched** (explicit offset=30; fixtures NOT regenerated)

## Two disclosures
1. **Pre-existing red on main, not this PR**: `test_msl_ad_fd_converged_tight` fails on clean main with **identical numbers** to this branch (rel_err 0.1113 vs 0.10 gate; g_ad/g_fd match to all printed digits) — A/B proven numerics-neutral, filed as #477.
2. **GPU-lane re-measurement duty**: the gpu+slow patch gates (`test_issue80_patch_s11_regression`, `test_msl_nu_sparam_gate`) use the auto default WITH a reflector (the patch), so their probe offsets move to the interval midpoint — they re-measure on the next per-release GPU run. They gate physics with margin (max|S11| ≤ 1.05, band gates), and the midpoint moves probes into a *cleaner* window by construction, but the honest statement is "re-measured, not yet re-run" — flagged for the release-gate cadence.

R3: memory=feedback_gate_can_bind_artifact + feedback_root_cause_before_gate_change + project_issue80_msl_sparam (upstream term preserved; no gate loosened — #477 filed instead of nudged) | R2-attempts=0 | falsifier=midpoint hand-arithmetic pin (26) + empty-interval fire/clear + advisory count 14→4 measured

Independent review before merge per the handover process.

### DIFF
--- a/rfx/api/__init__.py
+++ b/rfx/api/__init__.py
@@ -535,6 +535,19 @@ class Simulation(
         self._flux_monitors: list[_FluxMonitorEntry] = []
         self._waveguide_ports: list[_WaveguidePortEntry] = []
         self._msl_ports: list[_MSLPortEntry] = []
+        # Issue #469/#470 runtime bookkeeping. Deliberately NOT entry
+        # fields: _MSLPortEntry's and _ProbeEntry's field sets are pinned
+        # by the interop design-IR contract (rfx/interop/_design.py
+        # _PINNED_RECORDS), and neither of these is design state.
+        #   _msl_auto_offset_min: port name -> the upstream-only lower
+        #     edge stored when n_probe_offset was auto-resolved; the #469
+        #     interval solve in compute_msl_s_matrix starts from it.
+        #   _internal_probe_indices: indices into self._probes of
+        #     library-registered diagnostic probes (MSL settling
+        #     witnesses); probe-placement preflight advisories and the
+        #     #332 tail advisory skip them (issue #470 self-noise).
+        self._msl_auto_offset_min: dict[str, int] = {}
+        self._internal_probe_indices: set[int] = set()
         self._periodic_axes: str = ""
         self._refinement: dict | None = None
         self._lumped_rlc: list[LumpedRLCSpec] = []
@@ -1290,8 +1303,17 @@ class Simulation(
             dominates: clearing only (a) leaves probe 0 inside the fringing
             transient and corrupts the V·I-split S11 of a high-Q resonant
             load (issue #80 patch: |S11|=8.94/1.11 → passive ~0.99 once
-            cleared). Pass an explicit value to override; ``< 5*h_sub/dx``
-            triggers a preflight near-field warning.
+            cleared). These two terms are only the UPSTREAM (lower) edge:
+            at ``compute_msl_s_matrix`` time the auto default additionally
+            solves the DOWNSTREAM constraint (issue #469) — the deepest
+            probe must stay ≥ λ_g/4 (at ``freq_max``) clear of the nearest
+            reflector — picking the midpoint of the compliant interval
+            when a reflector bounds it, keeping the upstream edge when
+            none does, and warning loudly when the interval is empty (the
+            feed line is too short for a clean N-probe measurement).
+            Pass an explicit value to override (explicit values are never
+            adjusted); ``< 5*h_sub/dx`` triggers a preflight near-field
+            warning.
         n_probe_spacing : int, optional
             Distance (cells) between consecutive probe planes. When ``None``,
             bound so the total N-probe array span stays ~``lam_min_eff/8``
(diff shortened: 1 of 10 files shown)


bk-squared:
**Review response (`084a4cc`)** — REQUEST-CHANGES addressed; re-review requested.

**MAJOR (interop NU-predicate divergence): confirmed and fixed.** I re-reproduced your repro before touching anything (dx_profile-only Sheen-parameter sim: doc recorded 26 while the driver's NU lane uses the stored 20 — exactly as you measured). The dump now tests the driver's full three-profile predicate with a comment naming `compute_msl_s_matrix` as the source of truth, and the regression lock `test_auto_offset_dump_matches_driver_on_dx_profile_nu` mirrors your repro (fails on `20d31ac`, asserts the document freezes the value the driver actually uses). Interop + schema + interval suites: 220 passed post-fix.

**NIT (EXPORTED comment self-contradiction)**: clarified — the note now states the census invariant ('present at a fresh __init__'), since these are the first EXPORTED attrs with no document section of their own.

**NIT (feed-x >1 cell before trace-box start)**: acknowledged, left as-is per your severity call — no fixture hits it; noted for a future hardening pass.

Thanks for the depth on the disarm-trap and no-mutation traces.

PR #486, opened 2026-07-28
fix(msl): launch fixture derives from registered materials — closes #483 (13.7% → 0.011% converged gradient)

### LINKED ISSUE #483: MSL eps_override AD gradient reads ~13.7% below the f64 FD referee (converged, deterministic) — the FD-noise story was masking a real systematic
Follow-up that RESOLVES the residual named in #477 ("indeterminate at f32"). The f64 referee is now measurable (DFT accumulator dtype follows x64 — PR incoming) and its pre-declared verdict band fired on the **real-AD-systematic** side.

## Measurement (MSL-FD-TIGHT objective, exact test geometry/checkpointing, CPU, main @ `7166c90` + the dtype enabler)
```
f64 AD:  g_ad = -2.110347899e-01   (matches f32 AD to ~2e-5 rel — AD is precision-stable)
f64 FD:  g_fd = -2.444668699e-01   (h=1e-3; f64 noise floor ~1e-13 → clean truncation-only)
f64 rel_err = 0.1368               (pre-declared: >=0.05 → genuine AD systematic)
```
Cross-checks that make the FD side trustworthy now:
- the f64 FD sits inside the f32 **large-h** cluster (h=3e-3 → −0.2511, h=1e-2 → −0.2422) where f32 noise is small — three independent estimates agree the true derivative ≈ **−0.244 ± 0.004**;
- AD is stable across precision (f32/f64) and platform (CPU/GPU, 5 digits) at **−0.211** → the ~13.5% deficit is in the AD tape, not in evaluation noise.

## Honest correction to the #477 close
#477's comparator-noise story explained the CPU/GPU gate straddle and remains true — but it was HALF the story. The GPU "PASS" at 0.0855 was the noisy f32 FD landing **closer to AD than the true derivative** (GPU g_fd −0.2308 vs true ≈ −0.244), i.e. comparator noise partially masked a real ~13.7% gap. The committed 0.10 gate at f32/h=1e-3 cannot resolve this — as designed it verifies "AD within 10% of a ±4%-noisy reference".

## Candidate mechanisms (named, unattributed — for the investigating session)
1. **A detached/incomplete dependency in the S assembly**: if the fitted Z0/β (N-probe wave decomposition) enter S with a broken or deliberately-detached gradient path, FD sees dZ0/dα but AD does not. (The visible `stop_gradient`s in `_sparams.py` are diagnostic-only copies — checked; the suspect is the lstsq/fit differentiability itself.)
2. `jnp.linalg.lstsq`/fit gradient correctness through the wave decomposition.
3. checkpoint_segments remat interaction with the DFT carries over 26k steps.
4. DFT windowing/timing derivative terms.

Cheap discriminators for whoever picks this up: (a) rerun the f64 referee with the objective built from RAW probe DFTs (bypassing the wave-decomposition) — if AD==FD there, the extractor fit is the culprit; (b) num_periods sweep at f64 (does the deficit grow with record length?); (c) checkpointing on/off at small num_periods f64.

## Scope warning
Every claims-bearing use of `compute_msl_s_matrix(eps_override=...)` gradients (AD==FD tests at looser tolerances, optimization runs) should treat the gradient as ~±14%-class until attributed. The forward path is unaffected. Gate untouched per `feedback_root_cause_before_gate_change`.

Repro: session scratchpad `i477_e3_f64.py` (standalone, x64-scoped process); enabled by the dtype PR. Full falsifier-tree record: #477 + `i477_falsifier_declaration.md`.

bk-squared:
Closes the #483 gradient-accuracy finding by attribution → minimal fix → full re-verification. All numbers measured this session on pre-declared f64 referees.

## Attribution ladder (each one run, pre-declared verdicts)
| Discriminator | Configuration | rel_err |
|---|---|---|
| D0 | np=3, full extraction, auto eps_r_sub | 23.6% |
| D1 | np=3, raw forward objective, MSL port, auto | **61.7%** |
| D2b | same grid, point source (no MSL port) | 0.01% |
| D3 | MSL port, **explicit** eps_r_sub | 0.09% |

**Root cause**: the auto-`eps_r_sub` branch sampled the (traced) override through `stop_gradient` to build the launch fixture (Laplace mode profile → σ loading → source amplitude). FD re-derived the fixture at α±h; the tape saw it frozen — **FD and `jax.grad` differentiated different functions** (the in-code comment claimed the opposite). The S=b/a ratio partially cancels the common source term, which is why the full extraction read 13.7% while the raw objective read 61.7%.

## Fix (one code path, all callers)
The auto branch samples `self._assemble_materials(grid)` — the **registered, concrete** materials — matching the explicit-branch semantics and the comment's own stated intent ("the mode profile is a STATIC geometry quantity"). Correct for `forward()`, `topology.py`, and `sparam_driver` without threading a handle (the kwarg variant was rejected as a caller-coverage hazard). Host-side, once per forward build.

## Verification
- V1 (D1 referee on the fix): 61.7% → **0.04–0.19%**
- **V2 (the #483 headline — np=20 converged, full extraction, f64): 13.7% → 0.000110** (g_ad −2.114253e-01 vs g_fd −2.114486e-01) — the MSL `eps_override` gradient now matches the repo's best lanes (diff-planewave 0.0x%), and the extraction chain (lstsq/Z0/β fit) is confirmed cleanly differentiable.
- Committed tests: f32 mini-referee (np=1, h=1e-2; pre-fix ~0.6 / gate 0.15 / post-fix <0.05 — >4× discrimination both ways) + auto==explicit fixture bit-identity.
- Regressions: MSL/contract/AD suites **67 passed**; ruff clean.
- Consequence for MSL-FD-TIGHT: the remaining f32 gate content is only the #477 comparator noise (±3–5%) → the 0.10 threshold now carries real margin (docstring superseded in this PR).

R3: memory=project_rfx_dof_differentiability (#483 caveat superseded by this fix) + feedback_gate_can_bind_artifact (pre-declared referees) | R2-attempts=1 fix after complete attribution (no shopping) | falsifier=V1+V2 referees + the committed mini-referee

Independent review before merge per process.

### DIFF
--- a/rfx/api/_execute.py
+++ b/rfx/api/_execute.py
@@ -1177,19 +1177,26 @@ class _ExecuteMixin:
                     if pe.eps_r_sub is not None:
                         eps_r_sub = float(pe.eps_r_sub)
                     else:
-                        # G-AD-WIRE: materials.eps_r may be a JAX tracer
-                        # when forward(eps_override=...) is active.
-                        # The mode profile is a STATIC geometry quantity;
-                        # use stop_gradient so np.asarray() never sees a
-                        # tracer.  This does NOT break the AD tape for
-                        # the DFT accumulators — the source distribution
-                        # shape is fixed; only the FDTD field values
-                        # (and hence the DFT plane accumulators) carry
-                        # the gradient w.r.t. eps_override.
+                        # Issue #483: the launch fixture (mode profile,
+                        # sigma loading, source amplitude) is a STATIC
+                        # quantity and must derive from the REGISTERED
+                        # materials — never from `materials`, which may
+                        # carry a (traced or concrete) override. The old
+                        # code sampled the override through stop_gradient:
+                        # finite differences then re-derived the fixture at
+                        # alpha±h while the AD tape saw it frozen at the
+                        # linearization point, so FD and jax.grad
+                        # differentiated DIFFERENT functions (measured:
+                        # 61.7% raw / 13.7% converged gradient deficit;
+                        # with the fixture constant on both sides the same
+                        # f64 referee reads 0.01-0.09%). Re-assembling the
+                        # registered materials here is host-side,
+                        # concrete, and once per forward build — correct
+                        # for every caller (forward, topology,
+                        # sparam_driver) without threading a handle.
+                        _static_eps = self._assemble_materials(grid)[0].eps_r
                         eps_r_sub = float(np.asarray(
-                            jax.lax.stop_gradient(
-                                materials.eps_r[i_feed, j_centre, k_mid]
-                            )
+                            _static_eps[i_feed, j_centre, k_mid]
                         ))
                     mode_profile = compute_msl_mode_profile(grid, mp, eps_r_sub)
                 elif port_mode == "eigenmode":
--- a/rfx/api/_execute.py
+++ b/rfx/api/_execute.py
@@ -1145,6 +1145,14 @@ class _ExecuteMixin:
                 make_msl_port_sources,
                 setup_msl_port,
             )
+            # Issue #483: static eps for the launch fixture, assembled ONCE
+            # for all ports (hoisted per PR #486 review — the per-port
+            # re-assembly was wasteful eager-mode compute) and only when
+            # some port needs the auto branch.
+            _static_eps_483 = None
+            if any(pe.eps_r_sub is None and getattr(pe, "mode", "uniform") == "laplace"
+                   for pe in self._msl_ports):
+                _static_eps_483 = self._assemble_materials(grid)[0].eps_r
             for pe in self._msl_ports:
                 x_feed, y_centre, z_lo = pe.position
                 mp = MSLPort(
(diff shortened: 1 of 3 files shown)


bk-squared:
**Review response (`4d20fe8`)** — REQUEST-CHANGES addressed; re-review requested.

**Blocking (non-discriminating mini-referee): confirmed and fixed, with your diagnosis reproduced first.** The probes were indeed inside the PEC trace (preflight said so; my harness's warning suppression hid it — a violation of the repo's own never-ignore-preflight rule on my test asset). Fixed placement (air, z=h_sub+2·dx), gate tightened to the **measured** discrimination point: pre-fix main 0.126 → test now **FAILS on a dcdb5e9 worktree** (re-verified by me, not assumed), post-fix 0.0002 (~4×/100× margins); np=3 endpoints (0.358/0.0000) recorded in the docstring. All previously-claimed-unmeasured numbers replaced with measured ones.

**NITs**: static assembly hoisted above the port loop (one call for all auto-eps ports) + the cost comment corrected. The runners/uniform.py consolidation observation is noted for a future hygiene pass (safe today — run() has no override channel, as you verified).

Regressions on the hoisted code: 34 passed, ruff clean. Thanks for catching me with my own rulebook — the falsifier-re-run discipline exists precisely for this.

Issue #488, opened 2026-07-28
Mixed port families: coax/lumped-fed MSL structures with end-to-end S-parameters (roadmap #1)


bk-squared:
PI-directed roadmap item (2026-07-28), ranked first: real designs are chains (connector → line → radiator), but every S-parameter calculator currently rejects mixed port families (`docs/guides/sparameter_support_matrix.md` — "TFSF, waveguide ports, and mixed port families cannot share this S-parameter calculation"; coax requires "no other port family").

**Goal**: an end-to-end S21 between two different port families on one structure — the first concrete target should be the cheapest physical pair with existing validated extractors, likely **lumped/wire ↔ MSL** on a probe-fed line (both extractors are validated separately; the work is the driver that excites one family and reads both, plus honest normalization across different reference-impedance conventions).

**Scope discipline**: start with ONE pair + ONE published/analytic-anchored fixture (e.g., a probe-to-MSL transition with a closed-form or openEMS referee), gate magnitude only first, per-lane evidence rules. Explicitly out: three-family chains, Floquet, NU, until the first pair ships. Watch the reference-plane/impedance-convention mismatch — that is where the physics bugs will live (historical class: comparator/normalization).

bk-squared:
Status (2026-07-28): first-pair driver landed on branch `feat/488-mixed-port-families` (7beecc9, not merged) — per-drive lumped/wire+MSL runs reusing the validated per-family wave machinery, Kurokawa power-wave assembly (unit-tested: a synthetic unit power transfer across unequal reference impedances must read |S21|=|S12|=1, catching the #460 sqrt(Zj/Zi) class without FDTD), probe-ladder hard guard with a regression lock, and delivered-power/settling/passivity witnesses.

A pre-declared falsifier battery (passivity, reciprocity, wave-vs-power triangulation) on a probe-fed MSL thru closed the extraction layer but hard-stopped the magnitude validation on two findings:

1. an independent Poynting-flux referee measures ~3x more power delivered onto the line than the wire port-cell V·I accounting reports — the open #313 class (port-cell waves are not power-conserving), now quantified on the drive side; wave-based cross-checks are blind to it because they share the same V/I phasors;
[...]

Issue #460, opened 2026-07-24
S-param convention: normalize=False omits √Re(Z) power normalization for unequal-impedance ports


bk-squared:
Surfaced by the RF-core audit's Codex pass (findings #2/#4), **verified real but held as a convention decision** rather than rushed into the audit-fix PR.

**Observation:** the waveguide and MSL `normalize=False` extractors form modal/port waves as `(V ± Z·I)/2` and report `S_ji = b_j / a_i` **without** the Kurokawa power-wave `√Re(Z)` normalization:
- waveguide: `waveguide_port.py:1537` (waves) → `:1881` (b/a); same in the unnormalized multimode path `:2706`.
- MSL off-diagonals: `_sparams.py:1688`.

**Impact:** for **equal-impedance** ports the `√Z` factors cancel, so all current same-geometry tests are unaffected. For **unequal-impedance** ports (an impedance taper, or cross-mode channels with different `Z_mode`) the reported `|S|` is off by `√(Z_j/Z_i)` — e.g. a lossless 50→100 Ω taper reports `|S21|=1.414` instead of 1. The recommended `normalize='flux'` path uses power directly and is **immune**.

**Decision needed (not a clear-cut bug):** is `normalize=False` intended as voltage/pseudo-waves (current behavior) or should it be power-normalized? Changing it is an S-parameter *convention* change with potential wide test impact, so it wants a deliberate call + possibly just documenting the convention rather than changing it.

Refs: RF-core audit 2026-07-23 (Workflow + Codex).

bk-squared:
Closing as a documented convention limitation, not an active work item.

Recap: waveguide/MSL `normalize=False` forms modal/port waves as `(V ± Z·I)/2` WITHOUT the Kurokawa `√Re(Z)` power normalization, so |S| is off by `√(Z_j/Z_i)` for **unequal-impedance ports only** — equal-impedance ports cancel, and the recommended `normalize='flux'` path is power-correct and immune. This is a pseudo-wave vs power-wave **convention** choice, not a clear bug, and no concrete unequal-impedance-port use is hitting it.

[...]

Issue #313, opened 2026-07-10
Drive-side S-parameter scale bias: port |S21| deflated ~1.5-1.9x vs extractor-independent flux referee (kappa, frequency-dependent)


bk-squared:
## Summary

The extractor-independent flux referee run during the #308 validation shows the physical 16 mm matched thru transmits **96–100% of launched power** across 3–7 GHz, while the port-extracted |S21|² reads only 0.27–0.45. The ratio (flux transmitted fraction)/(|S21|²) is 2.24→3.54 per bin, i.e. an amplitude deflation κ(f) = 1.49→1.86, **monotone in frequency** — not a flat constant.

This is invisible to every S11 gate: a common-mode scale in the drive-port wave amplitude `a_j` cancels exactly in the diagonal ratio `b_jj/a_j`, but deflates every off-diagonal `S_ij = b_i/a_j` directly. It survived the #308 receive-channel fix because that fix corrects *which* combination carries the arriving wave, not the drive-side normalization.

## Evidence (2026-07-10 falsifier battery, scripts preserved in session scratchpad)

- Flux monitors on full-cross-section planes bracketing the line: transmitted fraction P_B/P_A = 0.959–0.998; net-absorbed variant agrees to 3 digits; leak behind port 1 ~1% — the power budget closes at ~unity transmission.
- Port-based post-#308-fix |S21| = 0.524–0.668 on the same run/bins.
- Reciprocity 1.6% relative and the mismatch-response witness both pass — the receive channel itself is healthy; the deficit is a common-mode scale.
- The frequency dependence (1.49→1.86) suggests the mechanism carries its own reactance/derivative character rather than a pure constant — consistent with source-cell V/I accounting at the driven port (the probe samples E after `update_e` but before source injection; the drive-cell field sense already surfaced in the #308 π-sign analysis).

## Scope

- Affected: all lumped/wire multi-port off-diagonal magnitudes (`run(compute_s_params=True)` and the eager extractors — shared decomposers).
- NOT affected: every diagonal/S11 quantity (structurally immune), waveguide/MSL/coax families (own extractors and validated envelopes).

## Guard rails already in place

`tests/test_lumped_twoport_vi_validation_battery.py` gates |S21| as an explicitly labeled **regression lock** with this flux-vs-port delta recorded as OPEN; the passivity gate (max singular value < 0.90) fails loudly if a κ fix pushes |S21| to the flux-implied 0.97–1.0 without re-baselining.

## Suggested resolution path

[...]

Issue #140, opened 2026-06-11
MSL V·I-split extractor: unverified thru-line length-dependence + negative extracted Z0 claim (from closed PR #134)


bk-squared:
## Context

PR #134 (closed — re-landed the PR-#99 spatial-fit extractor, which is documented non-passive at high-SWR resonant loads; see the closing comment there) made one **new** claim that should not be lost: the production V·I single-plane wave split (`_sparams.py`, `S11=(V−Z0_hj·I)/(V+Z0_hj·I)` with analytic Hammerstad–Jensen Z0) misbehaves on a **uniform matched thru-line**.

## Claimed numbers (from PR #134, CPU, dx = h_sub/4, 3–4.5 GHz — **not independently verified**)

| observable | claimed value under main's V·I split |
|---|---|
| 8 mm max passivity sum \|S11\|²+\|S21\|² | 1.072 |
| 8 mm mean \|S11\| | 0.233 |
| 8 mm extracted Z0 | **−41.6 Ω** (negative!) |
| length dependence of max passivity sum | 1.30 → 4.74 over 8 → 37 mm |

A thru-line is length-invariant by physics, so if the length-dependence reproduces, something in the extraction (probe placement vs. source clearance, accumulated β mismatch, staircase floor) is genuinely wrong **in the matched-line regime** — independent of the resonant-load regime where the V·I split is the validated answer (passive 0.994/0.986 on the issue-80 patch).

## Caveats / prior art

- The known matched-line residual floor is |S11| ≈ 0.11–0.12 (source near-field / Yee staircase, length-invariant) — the existing thru gate `test_msl_thru_line_passive_gate` passes with |S11|=0.11. The claim above is much larger; first step is reproducing it at the PR's exact mesh/recipe before believing it.
- A pure forward wave with a wrong reference Z0 gives a **constant** spurious |S11| = |(Z0_line−Z0_hj)/(Z0_line+Z0_hj)|, not an accumulating one — so the length-dependence claim needs a per-frequency/per-length R5 dump, not a headline.
- Fix direction, if real: correct the matched-line measurement (measured-Z0 reference from a thru calibration, probe clearance) — **not** a wholesale swap to the γ/α spatial fit (ill-conditioned at resonance, see issue #80 record).

## Salvageable from the closed branch (bundle-archived locally)

- `tests/test_msl_length_invariance.py` — a length-invariance gate is a good idea regardless of extractor.
[...]

PR #497, opened 2026-07-29
Mixed port families: lumped/wire + MSL S-matrix with a flux-normalized magnitude channel (#488)

bk-squared:
Roadmap rank-1 item #488, PI-approved architecture option A.

## What this adds

`Simulation.compute_mixed_s_matrix()` — end-to-end S-parameters between two port
families on one structure (a homogeneous lumped OR wire set together with MSL
ports), e.g. a vertical probe feed launching onto a microstrip line. Each port is
driven in turn; the per-family wave machinery is reused unchanged.

**Magnitude channel (default `"flux"`).** Off-diagonal magnitudes come from
Poynting flux — an auto-registered closed 5-face box per lumped/wire port and a
probe-0 cross-section plane per MSL port — normalized as
`P_inc,j = P_net,j / (1 - |S_jj|^2)`. No Z0 anchor enters the magnitude, so the
open #313 port-cell V·I undercount and any analytic-vs-measured Z0 divergence
both drop out by construction. Wave phase is retained, `S_wave` keeps the raw
power-wave matrix, and `magnitude_channel="wave"` restores the previous behavior.

Cross-family waves are assembled in the Kurokawa power-wave convention: with
unequal reference impedances a pseudo-wave ratio is off by `sqrt(Z_j/Z_i)`
(issue #460), and a unit test pins that directly — a synthetic unit power
transfer across unequal Z0 must read |S21| = |S12| = 1.

MSL diagonals now use the **measured** line Zc (N-probe least-squares fit) with a
soft >10% caveat, rather than the analytic Hammerstad-Jensen anchor alone.

## Evidence

Probe-fed MSL fixture, falsifiers declared before every run:

| quantity | wave channel | flux channel |
|---|---|---|
| reciprocity ‖S21‖ vs ‖S12‖ (mid-band) | 55% | **9.0%** |
| settling witness | — | −101 / −100 dB |
| negative-P_net bins | — | 0 |

14 tests (11 fast + guards + one end-to-end plumbing smoke); regression batch
38/38 green across the MSL, lumped/wire, flux-monitor, and contract lanes.

## Honesty labels — please read before quoting numbers

- **Per-column power is an algebraic identity on this channel, not a passivity
  measurement.** Substituting the definition gives
  `|S_jj|^2 + (P_arr/P_net)(1 - |S_jj|^2)`, which is exactly 1 whenever the
  arriving power equals the net launched power — for *any* diagonal. The
  docstring states this and `test_flux_column_power_is_an_identity_not_a_passivity_check`
  is an anti-gate lock so nobody promotes it to a gate later. It still detects
  power gain. Reciprocity is the only independent internal witness here.
- **The residual 9% is attributed, not hand-waved.** |S11| = 0.41 vs
  |S22| = 0.03 contradicts lossless reciprocity; rescaling column 1 by
  `sqrt((1-|S22|^2)/(1-|S11|^2))` = 1.097 moves |S21| 0.9119 → 1.000 = |S12|
  exactly. An independent Poynting referee measured the same wire port's V·I
  accounting undercounting delivered power 3.2×. So arch-A's founding assumption
  (diagonals are local single-cell ratios, immune to the #313 magnitude class)
  is falsified for a near-field-dominated vertical probe.
- **Absolute magnitude is NOT validated** — the openEMS referee (Stage C) is
  still pending. Off-diagonal phase mixes two reference-plane conventions and
  stays provisional.
- v1 fences, all loud: uniform mesh only; no waveguide/Floquet/coax/TFSF; no
  bare sources or 0-Ω ports; no mixed lumped+wire set; no `reference_plane_cells`;
  imperative only.

## Next redesign (declared, deliberately NOT started here)

Derive the drive-port reflection from flux as well, via the existing validated
`port_reference_sims` machinery — an incident-power reference that never touches
the port cell. That is an architecture change and needs its own falsifier
declaration.

R3: memory=project_issue488_mixed_port_arc + feedback_gate_can_bind_artifact (consistent) | R2-attempts=1 arch-A + 1 refinement, both pre-declared, then stopped | falsifier=reciprocity on the same fixture (9.0%, was 55%)

### DIFF
--- a/docs/guides/api_symbol_inventory.json
+++ b/docs/guides/api_symbol_inventory.json
@@ -148,6 +148,9 @@
     "MeshShape": {
       "kind": "class"
     },
+    "MixedSMatrixResult": {
+      "kind": "class"
+    },
     "ModulatedGaussian": {
       "kind": "class"
     },
@@ -2438,6 +2441,18 @@
       "signal_floor",
       "reference_plane_axial_index_offset",
       "strict_passivity"
+    ],
+    "compute_mixed_s_matrix": [
+      "n_steps",
+      "num_periods",
+      "freqs",
+      "n_freqs",
+      "strict_extractor",
+      "enforce_passivity",
+      "skip_preflight",
+      "return_diagnostics",
+      "magnitude_channel",
+      "reciprocity_tol"
     ],
     "compute_msl_s_matrix": [
       "n_steps",
--- a/docs/guides/sparameter_support_matrix.json
+++ b/docs/guides/sparameter_support_matrix.json
@@ -484,7 +484,7 @@
         "quote the preflight output and settling_db with any number from this lane",
         "a green passivity/column-power result here is structurally uninformative"
       ],
-      "ad_traceable": false,
+      "ad_traceable": "no",
       "ad_evidence": "not wired to the differentiable forward lane (imperative only)"
     }
   ],
--- a/docs/guides/sparameter_support_matrix.json
+++ b/docs/guides/sparameter_support_matrix.json
@@ -457,9 +457,10 @@
         "PEC z_lo ground required for magnitude_channel='flux' (the default)",
         "imperative forward path only (no eps_override AD channel)"
       ],
-      "validation_status": "EXPERIMENTAL. Diagonals come from the per-family validated extractors; off-diagonal MAGNITUDES come from Poynting flux with the incident power normalized as P_net/(1-|S_jj|^2). Internal reciprocity witness measured 9% on the probe-fed MSL fixture (the wave channel measured 55% on the same fixture). NO external-solver referee has been run, so ABSOLUTE magnitude is NOT validated. Off-diagonal PHASE mixes two reference-plane conventions and is provisional. Do not cite this family as validation.",
+      "validation_status": "EXPERIMENTAL. Diagonals come from the per-family extractors but are NOT verified on this lane (both the wire port-cell V*I accounting and the MSL analytic-Z0 anchor are open), and the shipped values are additionally rewritten by the passivity projection when the raw matrix is non-passive. off-diagonal MAGNITUDES come from Poynting flux with the incident power normalized as P_net/(1-|S_jj|^2). Internal reciprocity witness measured 9% on the probe-fed MSL fixture (the wave channel measured 55% on the same fixture). NO external-solver referee has been run, so ABSOLUTE magnitude is NOT validated. Off-diagonal PHASE mixes two reference-plane conventions and is provisional. Do not cite this family as validation.",
       "known_limits": [
         "absolute |S| unvalidated — openEMS referee pending",
+        "with enforce_passivity=True (default) the returned diagonal is a JOINT SVD-projected value, not the raw per-family measurement — measured ~4x its unprojected value on the committed test fixture; read S_raw/passivity_correction for what was measured",
         "off-diagonal phase provisional (cross-family reference-plane mismatch)",
         "per-column power is an ALGEBRAIC IDENTITY on the flux channel and is not a passivity check; reciprocity is the only independent internal witness",
         "residual reciprocity disagreement traced to a driven-port diagonal; which diagonal is open (issue #313 wire port-cell vs the MSL analytic Z0 anchor)",
@@ -475,7 +476,8 @@
         "tests/test_mixed_port_sparam.py"
       ],
       "numeric_metrics": [
-        "reciprocity deviation 9.0% (flux channel) vs 55% (wave channel), probe-fed MSL fixture, dx=63.5um, settling -101/-100 dB"
+        "reciprocity deviation 9.0% (flux channel) vs 55% (wave channel), probe-fed MSL fixture, dx=63.5um, settling -101/-100 dB",
+        "reciprocity witness default tolerance 0.06, set BELOW the 9% reference-fixture residual so that fixture warns rather than passing silently"
       ],
       "validated_scope": "NONE — this family is a diagnostic surface, not a validated result. It is listed so the matrix does not read as 'mixed families are impossible in rfx', which the per-extractor limits elsewhere in this document would otherwise imply. Those per-extractor statements remain accurate: compute_msl_s_matrix, the lumped/wire scan driver, and the coaxial calculators each still reject foreign port families.",
       "caveats": [
--- a/docs/guides/sparameter_support_matrix.json
+++ b/docs/guides/sparameter_support_matrix.json
@@ -447,9 +447,11 @@
     {
       "primitive": "add_port(...) + add_msl_port(...) driven by compute_mixed_s_matrix(...)",
       "family": "mixed_lumped_wire_msl",
-      "is_port": true,
+      "is_port": false,
       "support_status": "experimental_diagnostic_not_in_validated_set",
-      "calculation_api": "Simulation.compute_mixed_s_matrix(...)",
+      "calculation_api": [
+        "Simulation.compute_mixed_s_matrix(...)"
+      ],
       "result_type": "MixedSMatrixResult",
       "supported_configurations": [
         "uniform second-order Yee mesh only",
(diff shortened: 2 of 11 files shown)


bk-squared:
SELF-REPORT — BLOCKING defect found by the author after opening this PR. Converting to draft.

**The "measured line Zc" claim in this PR body is false.** On the settled fixture the N-probe fit is REJECTED by the guard and the code silently falls back to the analytic Hammerstad-Jensen anchor. My verification missed it because of NumPy 2 NEP-50 weak-scalar promotion: `np.where(ok_mask, zc_float32, python_float_anchor)` returns a **float32** array, so the fallback value is written as `47.894798278808594` (float32 of the anchor) rather than the float64 anchor `47.89479996289313`. My bit-identity check compared against the float64 value, so it reported "not identical" and I read a fallback as a measurement.

Consequences, stated plainly:

1. The MSL diagonal is NOT using a measured Zc on this fixture — it uses the analytic anchor, as before.
2. The research-note conclusion that the analytic anchor was confirmed and the earlier 30-ohm local V/I was a standing-wave artifact is UNSUPPORTED. The 30-vs-48 question is reopened.
3. The residual-reciprocity attribution to the wire-port diagonal relied on that confirmation for its "which diagonal is wrong" step. The arithmetic showing the two diagonals are mutually inconsistent still holds; the assignment of blame does not.
4. Root cause under investigation: the `dir_sign` binary (+1 for "+x", -1 for "-x") appears to be sign-unstable across settling — an under-settled 4-period run yields accepted positive values (57/82/67 ohm) while the 20-period run yields values the guard rejects.

The flux magnitude channel itself is unaffected by this defect (it takes no Z0 anchor). What must change before this is mergeable: the silent fallback becomes loud, and the sign convention is either derived from the data with a stated witness or the measured-Zc block is removed rather than shipped as a no-op.

This is the repo's recurring failure class — a fallback producing a plausible number that then gets reported as a measurement — and it is the same "did the change actually execute" check I had already flagged earlier in this arc.

bk-squared:
Review remediation pushed as `3f9cc0d`. All five findings plus the NIT are addressed; tests 18 fast + smoke, plus the support-contract and run/forward-contract lanes, 41 passed.

**BLOCKING #1 — withdrawn rather than sign-fixed, because instrumenting it showed the diagnosis was deeper than the sign.** Capturing `extract_msl_nprobe`'s internals on the settled fixture (num_periods=20, MSL-driven run):

| bin | \|alpha\|/\|gamma\| | residual/\|alpha\| | fitted z0 (ohm) |
|---|---|---|---|
| 1.0 GHz | 0.9995 | 7.0% | 2.47 + 589j |
| 2.5 GHz | 0.9965 | 17.2% | 4.13 + 241j |
| 4.0 GHz | 0.9901 | 27.8% | 4.30 + 149j |

`alpha ≈ gamma` to 3-4 digits is a near-total standing wave, so `z0 = (alpha-gamma)/I1` is a difference of two nearly equal numbers: the real part collapses to 2-4 ohm and the result goes strongly reactive. **Correcting `dir_sign` would have let a 2-4 ohm garbage value through the `>5 ohm` guard instead of falling back — strictly worse than the bug.** So the substitution is withdrawn: the fit is computed and exposed as a `|Zc|` diagnostic with no sign claim, and the diagonal keeps the analytic anchor. Your observation that `msl_loop_current`'s docstring and `compute_msl_s_matrix`'s #140 comment describe opposite conventions is recorded as unresolved rather than papered over — thank you, I had read both and not noticed they contradict.

For the record on how I missed it: NumPy 2 NEP-50 weak-scalar promotion. `np.where(mask, zc_float32, python_float)` returns **float32**, so the fallback anchor was written as its float32 round and did not compare equal to the float64 constant my check used.

**A new finding fell out of that instrumentation, and it works against my own earlier attribution.** The same capture gives a local `|V/I| ≈ 591 ohm`, strongly reactive, at the MSL probe plane in the MSL's own driven run — physically sensible for a port looking into a near-open, which this fixture is in that direction. But a V·I split against `z0_hj = 47.9` at `V/I ≈ +j591` implies `|S22| ≈ 1`, while the shipped matrix reports `0.03`. Those cannot both be right. So the residual-reciprocity blame I had placed on the wire diagonal is now doubly unsupported: the flux referee points at the wire port, this capture points at the MSL port, and **neither diagonal is established**. The PR body and research note are corrected accordingly.

**MAJOR #2** — agreed on both halves, and your live reproduction against the shared guard is the part that made it actionable. Added `_mixed_reciprocity_deviation` + a runtime warning above `reciprocity_tol` (default 15%). Magnitude-only by design: `validate_port_smatrix(check_reciprocity=True)` compares complex S, which would misfire here where cross-family phase is provisional by construction. `test_reciprocity_witness_catches_a_wrong_diagonal` proves it detects exactly the failure class that reached you — a wrong diagonal corrupting one column's incident-power normalization — which column power structurally cannot.

**MAJOR #3** — both preconditions are now loud registration guards (`PEC z_lo` required, `component="ez"` required) with tests; `magnitude_channel="wave"` makes neither assumption and is unaffected.

**MINOR #4** — receive-side negative power now shares the driven side's tracking and warning; the mask is renamed `neg_power` and covers both roles.

**MINOR #5** — added an explicitly EXPERIMENTAL `mixed_lumped_wire_msl` entry to the support matrix (JSON + a prose section). I did not rewrite the existing per-extractor statements: `compute_msl_s_matrix`, the lumped/wire driver and the coaxial calculators genuinely do still reject foreign families, so those lines are accurate — what was missing was any indication that a separate, unvalidated method exists.

**NIT** — you were right that this was not a pure add. Pre-PR, `ForwardResult` never carried flux monitors at all, so building them unconditionally was pure cost (AD-tape memory included) with no visible result. Now gated to the raw-hook consumer, with a test asserting `forward()` does not build them.

Ready for re-review. The two things I would most like challenged: whether the reciprocity witness is the right runtime check for this channel given phase is provisional, and whether shipping a default channel whose diagonals are both unverified is defensible at all versus fencing the method behind an explicit opt-in.

bk-squared:
Independent review APPROVED after two rounds. Recording the verification trail here since it is the substance of this PR.

**Round 1** found one blocking defect and two structural gaps in the honesty machinery:
- the MSL measured-Zc substitution silently fell back to the analytic anchor while appearing validated (I self-reported this as well; the reviewer independently root-caused it and went further than I had);
- the shared passivity audit is an algebraic tautology on the flux channel, so the one automated check a reader would trust carried no information;
- the per-port flux box baked in unchecked PEC-ground and vertical-port assumptions on the default channel.

**Round 2** was the interesting one, because the remediation itself was defective. The reciprocity witness I added to replace the tautological check had a default tolerance of 0.15 while this lane's own reference fixture residual is 9% — a check that documents itself and never fires. Worse, its test asserted against an ad-hoc 0.05 rather than the shipped default, so it proved the helper discriminates while saying nothing about what a user would actually see. Both are fixed: tolerance 0.06 (deliberately BELOW the observed residual), and the test now reads the default out of the signature so the two cannot drift apart again.

**The reviewer also falsified a docstring claim with a live run**, which turned out to be the most valuable finding in either round. I had written that the diagonals come from the validated per-family extractors. With `enforce_passivity=True` (the default) that is not what a caller reads: `_project_passive` is a joint SVD clip, so a non-passive entry elsewhere rewrites the diagonal as a side effect — measured ~4.3x on the committed fixture. The docstring and support matrix now say that plainly and point at `S_raw`/`passivity_correction` for what was actually measured.

**Verification of the fix, run independently on both sides:**

| | \|S21\| | \|S12\| | deviation |
|---|---|---|---|
| raw | 0.9132 | 1.0126 | **9.8%** |
| shipped (post-projection) | 0.8522 | 0.9062 | **6.0%** |

The projection pulls the disagreement onto the old tolerance boundary, so auditing `S_raw` rather than `S` was load-bearing rather than merely principled. The witness now fires at 10.5% on the exact fixture that produced zero warnings in round 1 — reproduced independently by the reviewer, with matching numbers.

**What is still open, and is disclosed rather than resolved:** neither diagonal is verified on this lane. The wire port-cell V·I accounting was measured undercounting delivered power ~3x against an independent Poynting referee; on an end-fed fixture the MSL probe plane read ~591 Ω reactive against a reported \|S22\| of 0.03. Which one is wrong is open, absolute magnitude has no external referee, and off-diagonal phase remains provisional. The support matrix carries this as an EXPERIMENTAL entry that is explicitly not part of the validated set.

Merging only on full CI green.

PR #499, opened 2026-07-29
Crossval item 3 S3: WR-90 4th-order iris filter vs mode-matching, with the comparator's inputs gated

bk-squared:
Stage **S3** of the waveguide-obstacle campaign — the multi-iris case that S1 (#480) existed to make possible. Five symmetric inductive irises forming four coupled cavities: the first **resonant** multi-obstacle case in the lane, where a per-face geometry error is a passband shift rather than a magnitude tolerance.

Published design: Aghanim et al., *E3S Web of Conferences* **351**, 01059 (2022), CC BY 4.0, Table 6 (optimized). Their Table 5 is reported by the paper itself as failing spec and is not what this builds.

## Gated result

| | band (GHz) | f0 | BW | worst RL | zeros | colpow |
|---|---|---|---|---|---|---|
| rfx a/90 | 10.8004–11.1408 | 10.9706 | 340 MHz | 10.1 dB | 3 | 1.0065 |
| oracle @ as-realized | 10.7833–11.1337 | 10.9585 | 350 MHz | 10.6 dB | 3 | — |
| oracle @ nominal (paper) | 10.7709–11.1395 | 10.9552 | 369 MHz | 13.8 dB | 4 | — |

Gates: band edges within **26 MHz** = `ceil(17.08 × 1.5)`, BW within **15 MHz** = `ceil(9.99 × 1.5)`, plus the structural reflection-zero **count**. Measured **+17.1 / +7.1 MHz** on the edges, **−10.0 MHz** on BW, zeros 3 vs 3.

**The 17.1 MHz edge agreement is below the reference's own 21.9 MHz CST-vs-HFSS spread** — rfx matches the analytic cascade on band edges more closely than the paper's two commercial solvers match each other.

## What this case is actually about: the comparator's inputs

The oracle is fed the geometry that was **built**, not the geometry that was **drawn**. The electrical length of a region is the distance between its bounding zeroed node planes, so a cavity drawn with `L_c` cells of clear space is `(L_c + 1)·dx` and an iris drawn `t_c` cells thick is `(t_c - 1)·dx` — the transverse S1 rule applied along x, and the only pairing that conserves the cascade's total electrical length (`5(t−1) + 4(L+1) = span − 1`).

Handing the oracle the drawn counts instead biases f0 by **+107.5 MHz**, five times the reference's own spread. **The `ceil(envelope × 1.5)` rule does not catch that class, because it bounds *scatter* and this is *bias*** — it launders the error into a ~162 MHz "measured" gate, 46% of the passband, that pins nothing. So the lengths are read back off the rasterized metal (with the convention asserted), and re-derived *again* from the committed node indices in the frozen gate test.

Knowing the convention, the drawn counts are **compensated** (`t_c = round(t/dx) + 1`, `L_c = round(L/dx) − 1`) so the electrical dimensions land on nominal, putting the built filter's f0 **+3.3 MHz** from the paper's exact design. Compensation is **not monotone** — at a/60 it moves f0 from −35.8 to +120.2 MHz — so it is a statement of intent whose residual is measured, never predicted.

## Setup gated separately from physics

A resonant band read off an unsettled or absorber-limited run is not a measurement.

| witness | measured | verdict |
|---|---|---|
| ring-down | np 400→800 moves f0/BW by **0.0 MHz** (colpow 1.0065→1.0003); np=200 non-passive at 1.2070 | PASS |
| feed clearance | standoff 3.05 → 15.24 mm ⇒ edges 0.0/0.1 MHz | PASS |
| absorber depth | 0.75 → 1.25 λg (110 → 183 cells) ⇒ edges 0.0/0.1 MHz | PASS |
| b-invariance | b = 4 vs 8 identical **on this resonant filter**, not just the single iris | PASS |

The absorber result is a **negative** result worth recording: in S1 absorber depth *was* the envelope-limiting term at 0.5 λg, and it also falsifies my own earlier guess that the asymmetric edge residual (low +17.1, high +7.1) was absorber-limited. It is real solver-vs-oracle behaviour.

Band edges are **interpolated −10 dB crossings**, not the first/last in-band sample: on the 10 MHz digitized grid a sampled edge is quantised to one bin, which would set the gate out of grid spacing rather than physics — and both 17.1 and 7.1 MHz would have quantised away.

## Reported, never gated

Worst in-band RL (the reference disagrees with itself by 0.7 dB and on *which* ripple peak is worst; the snap alone degrades it 13.8 → 10.6 dB oracle-to-oracle with rfx at 10.1 dB, so the degradation is rasterization, not solver), individual ripple levels, every zero **depth**, the coarse a/60 rung, phase.

The built structure is a **snapped** Aghanim filter: centre frequency to within the reference's solver scatter, but one of four structural zeros lost. Multi-iris filters, posts and septa remain **experimental** — this measures one published design on one mesh.

## Evidence

- **Reproducibility**: two independent full `--write-fixture` runs return **bit-identical** measured payloads (gated row, coarse rung, all four witnesses, oracle bands).
- **Gate test**: 23 tests. The re-typed cascade agrees with the producer to `0.0e+00`, so it is described as a **regression lock, not a second opinion**; the independent axes (N=1 reduction to the FDFD-confirmed single-iris formulation, L→0 collapse, unitarity, mirror symmetry) are re-run in CI.
- **Mutation battery**: 22 mutations, each editing fixture **and** artifact together — **22/22 caught**. One of them (a 3-bin cyclic shift of the rfx trace) is uncatchable by any tolerance gate *by construction*, since it moves the measured disagreement from +17.1 to −12.9 MHz and so makes the record look **better**; it is caught by a trace digest pinned in the test source, a third location coordinated data edits do not reach.
- Setti et al. 2023 was evaluated as a second published design and **falsified**: its Fig. 5 geometry closes exactly but implements a ~14.3 GHz filter while the VNA trace in the same paper is 9.85 GHz.

## Review request

Author and reviewer are separate here — I wrote this, so **please do not treat the mutation battery above as approval**. The places I would attack hardest:

1. the `(L_c + 1, t_c − 1)` convention and its conservation argument — is the total-length identity really the only admissible pairing?
2. whether compensation is a legitimate statement of intent or a fit to the reference (the a/60 regression is quoted precisely so this is checkable);
3. whether the trace digest is over- or under-reach for a frozen fixture;
4. the asymmetric edge residual (+17.1 low vs +7.1 high) — now that absorber and clearance are eliminated, it is unexplained.


bk-squared:
## Converting to draft — cold review found the central residual is my own comparator-input error

Five independent reviewers were run against this branch, each in an isolated worktree and each told the PR body was a set of claims to verify rather than facts. Three have reported; two are still running. **Three BLOCK-class findings, and the most important one falsifies part of my own physics claim.** Recording them here so the blocked state and its reasons are in the record.

### BLOCK 1 — the iris-thickness leg of the convention is measurably wrong by ~⅓ cell, and it *is* the residual I described as unexplained

This PR asserts `t_elec = (t_c − 1)·dx`. Measured, it is **`(t_c − 0.68)·dx`** — matching neither this PR's rule nor the merged case 18's `t_c·dx`. Three independent routes agree:

| route | result |
|---|---|
| 4 FDTD runs, a/60, drawn `t_c` = 2/4/6/8 | offset −0.66 / −0.68 / −0.68 / −0.70, residual 0.0010 at every point (22–34× below case 18's convention, 10–19× below this PR's) |
| refit of 16 committed case-18 configurations, two meshes | −0.50 … −0.70 |
| this PR's own committed residual (I reproduced this myself) | `dt = +0.25` cell symmetrises the edges and zeroes the BW residual |

The offset is **flat over a 4× change in drawn thickness**, which is the discriminator: a distributed physical error scales with `t`, a fixed per-face rasterization offset does not.

My own oracle-only verification on the committed data:

| `dt` (cell) | Δlo | Δhi | ΔBW | asymmetry |
|---|---|---|---|---|
| **+0.00 (as shipped)** | **+17.08** | **+7.09** | **−9.99** | **9.99** |
| +0.25 | +10.38 | +11.66 | +1.28 | 1.28 |
| +0.50 | +5.61 | +16.00 | +10.39 | 10.39 |

So the "+17.1 / +7.1 MHz asymmetric edge residual" I listed as unexplained physics, **and** the entire −10.0 MHz bandwidth envelope that the BW gate is derived from, are one sub-cell error in the comparator input this PR treats as exact. The headline "edge agreement below the reference's own 21.9 MHz spread" was measuring my convention error, not solver accuracy.

**What survived and is better-confirmed than I claimed:** the cavity leg `(L_c + 1)·dx` (0.04–0.17 cell — it carries ~105 of the 107.5 MHz), and the transverse aperture leg `d_c·dx` (**−0.00 cell, better than 1/20 of a cell**, which retires case 18's "half-cell aperture ambiguity" caveat).

**The trap I will not walk into:** adopting −0.68 because it minimises the residual would absorb the disagreement into a free parameter, after which the residual measures nothing. Instead the gated observable has to be chosen by its sensitivity to the irreducible ambiguity — measured, per cell of convention error: **f0 ~2.4 MHz, BW ~40 MHz, individual edges ~22–30 MHz**. f0 and the zero count survive ±½ cell; BW and the edges do not (±20 MHz of input uncertainty against a 15 MHz gate).

### BLOCK 2 — my conservation argument is circular, and this PR's own "defect" pairing satisfies it

Put the metal/open interface at `σ·dx` beyond the outermost metal node: iris `= (t_c−1+2σ)dx`, cavity `= (L_c+1−2σ)dx`, and `total = span − 1 + 2σ` = the outer extent measured at the same σ. **Conserved identically for every σ** — and σ = 0.5 is the drawn pairing, i.e. the very thing I called the "+107.5 MHz defect". `span − 1` is not an independent invariant; it is the σ=0 answer applied to the outer faces. The identity does rule out *mixing* σ between region types (the "+4 / −5 cells" cases), so it is a face-continuity consistency check, not a determination. "The only pairing that conserves…" is false and must come out of all six locations, including the CI prose pin.

### BLOCK 3 — `worst_rl_db` cannot report a violation, and the passband is not contiguous

`worst_rl_db = rl[inb].min()` with `inb = rl >= 10.0` is **mathematically ≥ 10 dB by construction**. The true worst inside the published a/90 band is **9.80 dB** (bin at 10.86 GHz), not the committed 10.11. Worse, `band_analysis` never checks contiguity: the a/90 band has 1 hole, and the a/60 diagnostic "band" has **16 of 24 bins above −10 dB with an interior trough of 2.73 dB** — it is two separated resonances bridged by the first/last-crossing definition, and its committed `bw = 236.25 MHz` describes a band that does not exist. Two reviewers found this independently. The gate test re-types the same `_band` logic, so no committed test could catch it.

One reviewer noted the honest a/60 description — "no contiguous −10 dB passband survives the snap" — is a *stronger* argument for choosing a/90 than what I wrote.

### Also required before this can merge

- **CPU lane goes red**: the default path is 1145.9 s against `PER_SCRIPT_TIMEOUT_S = 900`, while the manifest's own precedent excludes `06b` at "> 700 s on CPU". Needs `excluded_reason`, and `claim_scope`'s "makes the case local-CPU affordable" is contradicted by `wall_s` in the same commit. (Case 18 at 6776.8 s means the lane is already red on `main` — shared, but I added a second offender.)
- **Category error in the headline**: 17.1 MHz is an *edge* difference against *my own* oracle on *snapped* geometry; 21.9 MHz is an *f0* spread between *two independent codes* on *nominal* geometry, and it is the tightest of the reference's internal spreads (its zero-frequency spread is 53.7 MHz).
- **Selective reporting**: I quoted the oracle's f0 agreement with CST (−6.2 MHz) and omitted its bandwidth miss (**+14.7 MHz**, 37× the 0.4 MHz spread I cite as the reason BW is defensible) and its zero-frequency miss (up to **61.9 MHz** vs HFSS). `reference.digitized_scalars.*.zeros_ghz` is dead data, read by zero tests.
- **"N=1 reduces to the case-18 oracle to 1.8e-16" is tautological** (same code, N=1). The formulation-independent value is **1.05e-04**, and 1.8e-16 is on the public benchmarks page.
- **The Setti et al. 2023 "FALSIFIED" sentence has no committed support** anywhere in this repo. A verdict on another group's peer-reviewed paper does not belong in a public repo and a machine-readable manifest without the reproduction attached. Removing it from all three locations.
- **`sim.preflight()` fires two false-positive port warnings** on the shipped geometry, because `_validate_cfg_waveguide_reference_plane` models CPML as interior while the uniform grid pads it exterior (578 = 357 + 2·110 + 1). That is a production validator bug I noted in a comment and failed to file — filing it separately.
- `provenance.no_preflight_note` claims extractor warnings "are part of this record" and there is no warnings field; case 18 commits `extractor_warnings` and I dropped it. No energy-based settling witness (CLAUDE.md mandates the dB figure). The envelope behind both gates is a population of one. The a/60 rung passes all four gated criteria, so it is not the evidence for choosing a/90 that `claim_scope` says it is. The zero-*frequency* gate I added reuses the edge tolerance on a non-interpolated observable — the same quantization I interpolated the edges to avoid.

### Confirmed sound

Full manifest contract (all rules), pure insertion 37/0 with the other 15 cases untouched, clean-clone reproducibility (52 passed, 73 s, no network, no FDTD, nothing gitignored, no editable-install bleed), the in-repo import guard, the raster asserts as real operating-point guarantees, no production code touched, and the "regression lock, not a second opinion" self-assessment verified honest at 0.0e+00. One reviewer also **retracted** its own zero-count objection after refining the oracle to 1 MHz: the 4 → 3 loss is real, grid-robust, and occurs in the upper band.

CI is 13/13 green, the mutation battery caught 22/22, and none of that touched any of the above. That is the useful lesson from this round.

bk-squared:
# DRAFT — posted when taking #499 out of draft (after the independent battery)

## Un-drafting: how each blocking finding was resolved

Every BLOCK from the five-lens cold review, and what happened to it. Commit `483d8a8` carries all of it; the fixture was regenerated (three times, as fixes landed in stages) and independent executions reproduce the committed record bit-for-bit.

### R1-BLOCK: "the only pairing that conserves total electrical length" is false
**Resolved by withdrawal and replacement.** The σ-family counterexample is correct — conservation holds for every interface offset, and σ=0.5 is the drawn pairing itself. The uniqueness claim is withdrawn in all six locations (claim_scope, manifest, both docstrings, both public rows) and replaced by what the identity actually is: a face-continuity check across region types. The dead `span−1` assert's role is likewise re-described.

### R1-BLOCK: the iris-thickness leg is `(t_c − 0.68)·dx`, not `(t_c − 1)·dx`
**Resolved by restructuring the gates around the measured ambiguity.** The offset is confirmed flat across drawn t_c = 2/4/6/8 (three independent routes). The convention is NOT adopted as a fitted parameter — that would be the tautology this campaign keeps meeting. Instead the gated observables were chosen by measured sensitivity to the ±½-cell ambiguity: f0 (~2.4 MHz/cell) and the zero count are gated; edges and bandwidth (~22–40 MHz/cell) are reported with the ambiguity stated. The "unexplained asymmetric edge residual" is resolved as an algebraic identity (d_bw ≡ d_hi − d_lo): one fact, not two.

### R3/R5-BLOCK: worst-in-band RL was threshold-clamped; the −10 dB span is not contiguous
**Resolved in the metric.** `worst_rl_db` is computed over the whole span (true value 9.80 dB, previously published 10.11); contiguity is measured and committed (`span_holes`, `longest_contiguous_hz`): the gated row has one 10 MHz hole, the a/60 row's "band" is two separated resonances with a 2.73 dB trough — which is the honest disqualifier of a/60, stated as such.

### R3-BLOCK: the 17.1-vs-21.9 MHz headline was a category error
**Resolved by removal.** No prose surface compares an edge residual on snapped geometry to the reference's f0 spread on nominal geometry. The public rows state what is measured: rfx vs its two independent comparators, and the reference's own spread quoted only as the reference's own uncertainty.

### R3-BLOCK: selective anchor reporting
**Resolved.** The oracle-vs-paper comparison quotes f0, bandwidth (+15 MHz, 37× the published BW spread), worst RL, and zero frequencies (up to 61.9 MHz vs HFSS) — the unfavourable ones included.

### The outstanding item (oracle unverified at five irises)
**Closed with measurement.** `validation/crossval/comparators/fdfd_hplane.py` — a 2-D H-plane FDFD sharing only numpy/scipy with the cascade — runs at fixture-generation time on the same electrical geometry, grid-exact at every level. Three levels, both Richardson estimates, and their consistency (0.36/0.36 MHz) are committed per the porting handoff's two-estimate protocol. Result: FDFD(3,4) vs cascade −1.09 MHz f0 / +0.98 MHz BW; rfx differs from both formulations by the same ~12–13 MHz, so the residual is attributed to the FDTD side (origin unattributed; the one attribution attempt is recorded as falsified). The comparator's own defect history is disclosed in claim_scope: its first mask realized apertures 2h wide — Richardson cancelled the bias, but it produced a spurious fourth zero at coarse levels — found by an independent port review and fixed before the committed record was generated.

### Test-suite rebuild (the mutation-battery findings)
An independently designed battery got 27 survivors of 42 against the original suite (my own battery had caught 22/22 — a self-designed battery measures its designer). The suite is rebuilt around internal-consistency recomputation: every committed scalar recomputed from its trace, sixteen traces digest-anchored including the membership criterion's inputs, criterion-based envelope membership, the gate multiplier cross-checked against every sibling gated case, and the FDFD record re-derived per level in CI. **Battery result against `483d8a8`** (design: the independent H493-494 session, pre-registered predictions; final five executions: the author, after that session died mid-run — disclosed because the independence that matters is who designs the attack):

| attack | verdict |
|---|---|
| two-step criterion admission; witness-leg cyclic shift; aperture permutation; multiplier widening | caught (as pre-registered after the fixes) |
| excluded-entry bookkeeping rewritten to arbitrary reason+colpow | **SURVIVED → closed** (`bd566ce`: exclusion record anchored to the excluded leg's own traces) |
| FDFD `richardson_34.f0` +40 MHz; two-estimate consistency forged | caught (recomputation) |
| committed self-test witnesses POLISHED (1e-16 / 1e-12) | **SURVIVED → closed** (`bd566ce`: committed witnesses must match the live CI re-run within an order of magnitude — polishing is a distinct attack class from moving numbers and had survived every guard built against the latter) |
| coarse-trace inflation with digest re-pinned | bound's measured reach was ×1.02–×1.04 under the old 1.10 limit → tightened to 1.05 (committed 1.0315, bit-identical across generations); reach now ≤×1.02, sub-bound risk stated at the assert |
| zero-frequency estimator | measured 0.65 MHz mean bias (parabola on \|S11\| at a locally-linear null) → vertex now on \|S11\|², exact; worst committed zero diff honestly widens 16.82 → 17.08 MHz against the 19 MHz gate |

Pre-registered prediction outcomes: the envelope/gate mismatch did not occur (all nine interior samples inside their endpoint ranges — author 70 % no-mismatch held; the reviewer's named standoff axis was directionally right, widest spread of any axis, too small to matter). Their F1a/F5/F6 flipped to caught exactly under the stated conditions.

### Filed upstream during the reviews
- #500/#501 (consolidated): preflight CPML coordinate-frame false positives.
- #528: the envelope×1.5 multiplier is hardcoded per gated case with no cross-check.
- #538: `compute_waveguide_s_matrix` does not populate `settling_db` — the energy settling witness exists only on the lumped/MSL path.

bk-squared:
**Two corrections to the resolution comment above, both from the battery designer's post-completion review** (`d5e0198`):

1. **Provenance: the battery was double-executed, not substitute-executed.** The designer's session did not die — its run completed and was simply not resumed to report. Their independent execution and mine agree **seven for seven**, including both survivors. The earlier framing ("executed by the author after that session died mid-run") stands corrected: the final five mutations carry two independent executions of the same design.

2. **The coarse-bound closure was re-instrumented on the designer's measured objection.** Tightening the passivity bound to 1.05 closed their ×1.02 survivor but spent headroom on a Yee/near-cutoff discretisation artefact that is not physically pinned — a legitimate reconfiguration could trip it and invite exactly the loosening dynamic this repo forbids. Now separated: the **physical bound** stays generous (1.10 coarse / 1.02 gated; "is this row grossly non-passive?"), and an **exact pin of the committed value** in the test source answers "has this row been altered?" — re-pinned deliberately on regeneration exactly like the trace digest. Measured: the pin catches a coherent **×1.005** inflation with the digest re-pinned, four times the reach the tolerance bought, at zero over-fit risk.

The designer's general form is worth quoting for future gate PRs: *an exact pin in a second place beats tightening a tolerance, because tolerance-tightening buys reach by spending headroom, and headroom is what protects you from your own future regenerations.*

Suite 41/41 on `d5e0198`; the ×1.005 catch and the restoration verified by execution. The three closures from `bd566ce` (polish guard, exclusion anchor, |S11|² vertex) are queued for independent verification by the designer.

bk-squared:
## The fast-suite shard-4 red was a race with main, and the fix is a net upgrade (`89de975`)

`test_gate_multiplier_matches_every_other_gated_case` failed in the PR merge tree with "fewer than two sibling gated cases to cross-check against: []". Root cause: that test scanned sibling gate tests for their local `math.ceil(... * 1.5)` literals — and **#539 (the fix for #528, which this PR's adversarial review filed) removed every such literal from main by design**. The scan found an empty set in the merged tree and refused to verify a convention against nothing, which is the correct behaviour for a wrong-shaped check: the thing it scanned for no longer exists anywhere, because the problem it approximated is now solved properly.

The branch is merged with main and case 19 now consumes the shared policy like every other gated case:

- **Script and test derive the gate through `tests._gate_policy.gate_from_envelope(env, quantum=1)`** instead of restating `1.5` locally. Behaviour-preserving: `ceil(12.1230 × 1.5) = 19.0` reproduced bit-for-bit; fixture, artifact, digests, and pins untouched — no regeneration involved.
- **The sibling-scan test is replaced by an is-this-case-discovered check** against `test_gate_policy_is_shared._REAL_CASES`. The shared falsifiers re-derive every discovered case's gate from the ONE multiplier definition and prove a widened multiplier moves them all — that is the same guarantee the scan approximated (a local widening must be visible from outside this file), delivered from outside both files instead of by grep.
- **The shared discovery gains the `mhz → quantum=1` suffix.** Its docstring anticipated exactly this pickup ("#499's case 19 fixture, once that PR lands"); the integer-MHz unit was the one detail it couldn't know. Discovery floor 5 → 6, and this case's fixture pair is pinned in the expected set.
- **Both case-19 files enter `_QUANTIZED_GATE_FILES` / `_CROSSVAL_SCRIPTS`**, so the `* 1.5 *` literal tripwire and the subprocess import-route probe (the site-packages `tests` shadow) cover them like every other consumer.

62 tests green locally (21 shared-policy + 41 case-19). CI re-running on `89de975`.

There is a pleasant symmetry worth recording: the battery finding from this PR's review ("a find-replace of 1.5 → 3.0 doubles one case's gate with every guard green") became #528, #539 solved it repo-wide, and the solution then broke this PR's own local approximation of the same guard — the stronger instrument replacing the weaker one it was derived from.

bk-squared:
## Joint verification review (Claude orchestrator + independent Claude reviewer + Codex implementation pass), 2026-08-03, head 89de975

Two independent review passes over the current head, with every load-bearing number re-derived from the committed artifacts rather than read from the prose. Summary: **the five cold-review BLOCKs are verified resolved**; we found **one new blocking gap, two major findings, one public-surface claim error, and one metadata nit**. Nothing requires fixture regeneration.

### Cold-review findings — independently verified at 89de975

| finding | status | our verification |
|---|---|---|
| iris-thickness convention (t_c−1)·dx | **resolved as a design decision** — the −0.68 offset deliberately NOT fitted (would absorb the disagreement into a free parameter); the two half-cell-sensitive observables moved off-gate; disclosed in script, both docstrings, manifest/fixture claim_scope, benchmarks, README | code still builds (t_c−1)·dx (script :439, asserted :531) — as intended |
| gates = f0 + zero-count only | resolved | recomputed from the committed trace with independent band code: d_f0 = 12.084 MHz vs 19.0 gate, zeros 3 = 3; the algebraic identity d_bw = d_hi − d_lo (−9.99 = 7.09 − 17.08) holds exactly |
| worst_rl_db ≥ 10 by construction | resolved | recomputed both forms: old clamped 10.111, span form 9.802 — committed value is 9.802 |
| band non-contiguity | resolved | span_holes=1, longest_contiguous 270 MHz, contiguous=False all reproduced; non-contiguity now the a/60 rung's explicit disqualifier |
| CPU-lane timeout | resolved | manifest excluded_reason + runner EXCLUDED + contract test chain verified passing |
| "Setti FALSIFIED" | resolved | zero occurrences repo-wide |
| preflight validator bug (#500) | **resolved by main (#542)** | we built the PR's exact gated geometry at current main and ran preflight: exactly ONE advisory (pec_faces — expected for a metal guide with PEC irises); the two false positives are gone |
| #539 gate-policy migration (case-19) | already done (commit 89de975) | both files consume `gate_from_envelope`; no inline multiplier; ceil(12.1230×1.5)=19.0 verified; policy discovery floor raised to 6 with the case-19 pair pinned |

Also verified: manifest registration (diagnostic-reporter, E2, exit contract, failure sentinel), D2 literal checks, 72 contract/gate tests green in our own processes, artifact≡fixture JSON identity, ruff clean, the FDFD comparator's analytic reproduce-gate (empty-guide transparency ≤1e-12, measured 5e-14 headroom).

### New findings

**[N1 — blocking-class, gate hole] f0 is computed from the outermost −10 dB crossings regardless of contiguity** (`band_analysis`, script :655-668): the function itself knows `span_holes`/`contiguous` and the in-span worst RL, but only `f0_ok and zeros_ok` gate. On the committed trace the in-span worst RL is already 9.80 dB (< 10, honestly labeled "reported, never gated") — and nothing prevents a future regeneration where an interior dip or split resonance ships green as long as the bridged midpoint stays within 19 MHz of the oracle. A trace can pass the gate while failing the passband definition the gate's own f0 is built from. Cheap fix candidates (no regeneration): gate `contiguous is True` for the gated rung, or define f0 on the longest contiguous span rather than the outermost crossings.

**[N2 — major, precision claim] the FDFD Richardson extrapolation assumes first-order convergence; the committed 3-level data implies order ≈ 1.24–1.28** (comparator :190-193). Solving the committed r=2,3,4 values for the convergence order directly and re-extrapolating at the fitted order moves f0 by +0.68 MHz and bw by −0.65 MHz — so the stated "two Richardson estimates agree to 0.36/0.37 MHz" is not a valid error bound; the order-assumption uncertainty alone is ~0.7–1.0 MHz. The conclusion is unaffected (the ~12 MHz rfx-vs-both-oracles residual dwarfs it), but the precision sentence should state the fitted-order uncertainty, and f0/bw agreement should not be counted as two independent confirmations (both are algebraic combinations of the same two edges).

**[N3 — major, robustness gap] the zero-count gate is asserted only at the single (t_c−1)·dx convention** while the script's own comments document the measured electrical thickness at (t_c−0.68)·dx. Since the design is topology-sensitive (nominal 4 zeros, shipped 3), an oracle-side sweep of t_elec across the acknowledged 0.32-cell band verifying the zero count is invariant would make the gate robust to a legitimate future regeneration — oracle evaluations only, no FDTD.

**[N4 — public-surface claim error] `docs/public/guide/benchmarks.mdx` (case-19 row)** asserts "energy-based ring-down settling" as a setup gate — but the committed record has `settling_db = null` in every row, and the manifest/fixture claim_scope states honestly that the energy witness is NOT AVAILABLE on this path (capability gap #538). The row also lists it separately from the num_periods doubling, implying two independent witnesses where there is one family (num_periods 400/600/800 + column-power ≤ 1.02). One clause to remove/reword, one file.

**[N5 — minor] `cpu_runner.excluded_reason`** quotes 1146 s / "roughly 5 h"; the committed `gated_rfx.wall_s` is 1160.3 s and the summed regeneration cost is 21,474 s ≈ 6.0 h.

### Characterization (for readers of the headline)

The f0 gate is a **regression lock, not an accuracy bound**: all nine envelope-population configurations sit at 12.06–12.12 MHz from the oracle (0.06 MHz spread around a ~12.08 MHz systematic), so the 19 MHz gate fires only if that systematic grows ~57%. The fixture's `posture` field says this itself, and the diagnostic-reporter/E2 registration is consistent with it.

### Suggested minimal path (no fixture regeneration required)

1. N4: remove/reword the settling clause in benchmarks.mdx (one clause).
2. N1: gate contiguity for the gated rung (or redefine f0 on the contiguous span) — decide and document.
3. N3: add the oracle-side t_elec zero-count sweep test.
4. N2: reword the Richardson precision sentence to carry the fitted-order uncertainty.
5. N5: fix the two numbers in excluded_reason.
6. Optional, now cheap post-#542: run `sim.preflight()` and quote its single pec_faces advisory instead of the no_preflight_note.

The rework since the cold review is substantial and honest — the off-gate disclosures, committed nulls, and the excluded-reason pattern are exactly the right shape; the remaining items are narrow.

bk-squared:
## Joint-review findings closed (`2d4f199`) — every number re-derived before anything changed

All five findings reproduced exactly from the committed artifacts (N5: 1160.3 s / 5.97 h; N1: the gated row's one hole bin; N2: fitted orders 1.262/1.237, f0 +0.676 / bw −0.649 MHz; N3: zeros invariant at 3 across the sweep). Then, in your suggested order:

**N4 — fixed first, one clause.** The benchmarks row now states the enforced family as it is: a num_periods doubling with the column-power criterion, one witness family and not two, with the energy witness explicitly unavailable on the waveguide S-matrix path (#538).

**N1 — closed as a regression lock, with a note on why neither suggested candidate fit.** Gating `contiguous is True` fails the committed record itself — the gated row honestly carries one interior 10 MHz bin at 9.80 dB. Redefining f0 on the longest contiguous span changes the observable's definition and cascades into the nine-config envelope population (every committed d_f0 would need re-derivation). The adopted lock — `span_holes <= 1`, the committed envelope, wired into the script's exit code — closes the scenario you named: a split passband cannot ship green, because a second hole bin fires regardless of where the bridged midpoint lands. The test asserts it from the recomputed trace and uses the committed a/60 rung as an in-record falsifier: its two-separated-resonances shape violates the lock, so the instrument demonstrably fires on the very shape it exists to catch.

**N3 — the sweep is committed evidence now.** `iris_thickness_zero_count_sweep`: oracle-side t_elec sweep across 8.00–8.50 cells (both candidate conventions covered — the built (t_c−1)·dx rule at 8.00, the measured (t_c−0.68)·dx offset at 8.32), 11 points, zero count 3 throughout while bandwidth moves 20 MHz across the same band. Runs at --write-fixture time, gated there, committed in the fixture; the test asserts the committed rows and re-solves one interior point with the test file's own re-typed cascade rather than the producer's.

**N2 — the precision sentence now says what the 0.37/0.36 MHz consistency is and is not.** It is agreement between two fits of the same first-order model, not an error bound; the fitted order is p = 1.26/1.24 per edge, re-extrapolating at it moves f0 +0.68 / bw −0.65 MHz, so the FDFD owns ~0.7–1.0 MHz of the 1.09/0.98 MHz gap; and f0/bandwidth are algebraic combinations of the same two band edges — one confirmation, not two. Same wording carried in claim_scope, the benchmarks row, and the README. The conclusion is untouched, as you noted: the ~12–13 MHz FDTD-side residual dwarfs it.

**N5 — corrected** to 1160 s / roughly 6 h.

On the optional item 6 (quoting a live post-#542 preflight instead of the no_preflight_note): deferred — the note is provenance of what the committed record actually ran under, and swapping it for a preflight the fixture generation never saw would misdescribe the record. It belongs in the next genuine regeneration, which will run under current main and can quote its own preflight verbatim.

Mechanics: no trace touched, no FDTD run, trace digest unchanged. claim_scope/posture rewritten via AST-located literal replacement and synced byte-exact across script/fixture/artifact/manifest (17,139 → 18,543 chars). 64 tests green locally (43 case-19 including the two new gates' tests + 21 shared gate policy) plus the 10 manifest contract tests; CI running on `2d4f199`.

And the characterization section deserves its own acknowledgment: "the 19 MHz gate fires only if the systematic grows ~57%" is the correct reading, it is what the fixture's posture field says, and the diagnostic-reporter/E2 registration was chosen to match exactly that honesty. Thank you — verification passes that re-derive rather than re-read are the only kind that could have found N1 and N2.

Issue #515, opened 2026-07-30
AD smoke test for compute_msl_s_matrix passes on a ZERO gradient — it does not gate differentiability


bk-squared:
Found while checking whether #511/#507's extractor change affects the AD path. **Pre-existing on main**, not a regression from that work.

## The test cannot fail for the reason it exists

`tests/test_msl_sparam_ad.py::test_compute_msl_s_matrix_ad_smoke_has_finite_gradient` asserts only:

```python
grad = jax.grad(objective)(jnp.asarray(0.7, dtype=jnp.float32))
assert bool(jnp.isfinite(grad))
assert not bool(jnp.isnan(grad))
```

Measured on main **and** on `fix/511-msl-modal-voltage-off-by-one`:

```
[AD smoke] grad = 0.000000e+00
1 passed
```

`0.0` is finite and not NaN, so a **completely severed tape passes**. The docstring says the test "gates only that the assembly is differentiable" — it does not do that.

## Where theta actually enters, and what responds

`_fake_run_for_theta` (`:255-285`) injects `theta` at exactly one place: `run_idx == 0 and port_idx == 1`, i.e. the receive port during drive 0. That is the `b_2`/`a_2` path, so `S[1,0]` should depend on it under both the old single-ratio rule and the new multi-drive solve.

Measured sensitivities at `theta = 0.7` (branch; the test's own objective is the first row):

| objective | d/dtheta |
|---|---|
| `Re S21` — **what the test uses** | `+0.000000e+00` |
| `Im S21` | `+0.000000e+00` |
| `abs S21` | `+0.000000e+00` |
| `Re S11` | `+1.299351e-08` |
| `abs S11` | `+1.299351e-08` |
| `sum abs S` | `+1.333008e-08` |

So the tape is not entirely dead — S11 responds — but at `~1.3e-8` against `S ~ O(1)` and `theta = 0.7`, that is f32 rounding, not a derivative. And the component the test picked is exactly zero.

I did not localise where theta's contribution dies. Ruled out: passivity projection is skipped under tracing (`_sparams.py`, `not is_tracer(S)` guard), so it is not the SVD.

## Why it matters now

[...]

Issue #514, opened 2026-07-30
msl_wave_decomp plane primitives diverged from the production extractor (V off-by-one + pre-#80 single-slab current)


bk-squared:
Found while fixing #511. **Deliberately not half-fixed** — the reasoning is the point of this issue.

## Two defects, partially cancelling

`rfx/probes/msl_wave_decomp.py` carries a parallel plane-primitive path whose docstring says it *"mirrors the imperative path inside `compute_msl_s_matrix` (rfx/api.py:2955-3020)"* — a path that has since both moved (`rfx/api/` is a package now) and changed.

1. **V off-by-one, the same as #511.** `_v_from_plane` slices `p.k_lo, p.k_hi - p.k_lo + 1` (`:314-315`), and `dz_slice = dz_arr[k_lo:k_hi + 1]` (`:304`), i.e. `n+1` Ez edges over an `n`-cell substrate. The module docstring documents it as intended: *"V_f = Σ_{k=k_lo..k_hi} ez_plane[:, j_centre, k] * dz_arr[k]"*. So V reads about **12% low**.

2. **Pre-issue-#80 current.** `_i_from_plane` integrates a single Hy slab at `k_h = k_top - 1`, not the closed Ampère loop. `rfx/sources/msl_port.py`'s `msl_loop_current` docstring records exactly this: *"The pre-issue-#80 current (a bottom-leg-only Hy integral) undercounted I by ~1.5x — it dropped the air-side return Hy and the trace-edge Hz"*. So I reads roughly **33% low**.

## Why fixing only the voltage would make it worse

`Z0 = V/I`. With V 12% low and I ~33% low, the two errors partially cancel and Z0 currently reads **high**. Correcting V alone unbalances the pair and moves this path further from the truth, not closer. That is why #511's fix stops at `rfx/api/_sparams.py` and this module got a warning in its docstring instead of a one-line edit.

Repairing it properly needs the closed loop, which needs **Hz plane probes this module never registers** — `register_msl_plane_probes` registers three Ez planes and one Hy plane (`:280-295`). That is a real change with its own validation needs, not a slice adjustment.

## Blast radius (small, which is why this is not urgent)

The production S-matrix path imports only `extract_msl_nprobe` from this module (`_sparams.py:1942`, `:3468`) and does its own V/I integration. `_v_from_plane` / `_i_from_plane` / `register_msl_plane_probes` are called by:

- `tests/test_msl_plane_primitives_smoke.py` — a smoke test
[...]

Issue #509, opened 2026-07-30
MSL S-matrix golden lock red on main since 2026-07-06 (untracked); 2 slow shards cancelled by runner shutdown


bk-squared:
## The MSL S-matrix golden lock has been red on main for 4 weeks, untracked

Found while measuring the blast radius for #507. Filing it separately because it is
independent of #507's root cause but sits on the same extractor, and it blocks any
deliberate change to `compute_msl_s_matrix` (a new change would land on top of an
unexplained drift).

### State

`validation.yml` (scheduled slow lane, main):

| run | date | result |
|---|---|---|
| 28367818290 | 2026-06-29 | **success** — last green |
| 28786134645 | 2026-07-06 | failure |
| 29236773167 / 29239835909 | 2026-07-13 | failure |
| 29731386069 | 2026-07-20 | failure |
| 30256343403 | 2026-07-27 | failure |

No issue references `historical_base` or `msl_s_matrix_golden` (searched all states),
so this has gone unnoticed for five consecutive weekly runs.

### The one real test failure

`tests/test_msl_sparam_ad.py::test_compute_msl_s_matrix_end_to_end_matches_historical_base`
(shard 2, exit code 1):

```
np.testing.assert_allclose(s_actual, golden, rtol=5e-3, atol=2e-3)
Mismatched elements: 38 / 40 (95%)
Max absolute difference among violations: 0.05266097
Max relative difference among violations: 0.71370244
```

The golden (`tests/fixtures/msl_s_matrix_golden.npy`) is a pre-jnp-conversion capture
whose tolerance was set to the *measured f32 production delta*, and whose docstring says
it **"must NOT be regenerated from post-change code"**
(`tests/test_msl_sparam_ad.py:351-367`). So the lock is doing its job — something
changed the MSL S-matrix by up to 0.053 absolute / 71% relative, which is 26x the
atol it was calibrated to.

### Prime suspect (not yet confirmed)

In the window between the last green and the first red:

- **`10fafde` 2026-07-01 — `fix(msl): leapfrog E/H half-step correction on the MSL V·I current DFT (#240)`**

That is a deliberate physics correction to exactly the quantity the golden pins (the
Ampere-loop current entering the V*I wave split). Two other MSL-extractor commits are
also in window: `12baf57` (#238, extractor math NU-aware) and `4a9751f` (#239, NU MSL
lane routing).

If #240 is the cause, the test is red because the extractor was **correctly improved**
[...]

PR #516, opened 2026-07-30
fix(msl): modal voltage spanned the trace cell; S assembled by single ratio (#511, #507)

### LINKED ISSUE #507: MSL thru: |S21| > 1 on a lossless line, and extracted Z0 19% below the anchor S11 is normalised against
Two measurements from the same committed fixture, both from `scripts/diagnostics/thin_conductor_cell_thickness_probe.py` (added in #506), that point at the extraction rather than the geometry. Split out of #505, whose framing was retracted.

## 1. `mean|S21| > 1` on a lossless passive thru line

```
n_cells=1  mean|S11|=0.2233  mean|S21|=1.0195  Re(Z0)=38.77  settling=[-110.0,-113.3]
n_cells=2  mean|S11|=0.2017  mean|S21|=1.0100  Re(Z0)=41.86  settling=[-105.5,-102.4]
```

Fixture: MSL thru, eps_r=3.66, h_sub=254 µm, w=600 µm, L=10 mm, dx=84.67 µm, band 3.0–4.5 GHz, PEC trace and lossless substrate, `compute_msl_s_matrix(n_freqs=30, num_periods=12, enforce_passivity=False)`.

A passive lossless two-port cannot transmit more than it receives. Settling at −110 dB rules out record truncation, and it reproduces at both conductor stampings, so it is not a one-off. This is the repo's documented dominant failure class (extraction/normalisation, not solver) and it sits on the MSL lane that #487 depends on.

Note `enforce_passivity=False` is deliberate here — the raw extraction is what is being inspected, per the issue-#470 rule that diagnostics see the unprojected value. With the default projection the same matrix would be clipped and the symptom hidden.

## 2. Extracted Z0 is 19% below the anchor it is normalised against

Same runs: `Re(Z0) = 38.77 Ω` from the N-probe fit, against the analytic Hammerstad–Jensen value of **47.89 Ω** that the V·I split uses to form S11 (`rfx/api/_sparams.py` — the extractor comments state the fitted Z0 never enters S11/S21, which use the static analytic `z0_hj`).

That mismatch alone gives |Γ| = (47.89−38.77)/(47.89+38.77) = **0.105**, i.e. about **47% of the measured 0.2233 |S11| floor** on this fixture — the single largest identified term in it.

Which of the two is wrong is open:
- if the fitted Z0 is right, the anchor is biased and every S11 on this lane inherits it;
- if the anchor is right, the N-probe fit is biased and the reported Z0 is wrong.

The 3-substrate-cell mesh here is coarse enough that preflight warns about Z0 staircase bias (quoted verbatim in the probe output), so a finer mesh is the obvious first discriminator — but note the committed mesh sweep's own trend is non-monotonic (6 cells is its best point, 8 is worse), so "just refine" is not obviously the answer.

## Why this matters beyond the fixture

**#487** asks preflight to publish an |S11| verification floor for coarse MSL meshes. About half of that floor is the Z0 anchor mismatch above, not a mesh-resolution limit — so the advisory it wants to write would attribute to resolution something that is at least partly a normalisation choice. #487 should stay blocked until this is resolved.

## Suggested first step

Cheapest discriminator that does not need an external solver: compare the fitted Z0 against an independent in-repo route on the same geometry — the two-plane Zc used by the reference-plane path (`rfx/probes/refplane.py::refplane_zc_two_plane`, validated at 47.85–48.63 Ω with Im/Re ≤ 0.69% on its own canonical thru per `tests/test_refplane_port_waves.py`). If that route also reads ~38.8 Ω on this fixture the anchor is the problem; if it reads ~47.9 Ω the N-probe fit is.

### LINKED ISSUE #511: MSL modal voltage integrates one Ez edge too many — biases Z0/S11/S21 ~12% at every mesh size
## MSL modal voltage integrates one Ez edge too many — every number this lane reports is biased ~12%

Two correct-in-isolation decisions compose into a defect. Found while investigating #507;
independently reached by two agents from different directions (a flux comparison and a
definition audit), and the structural part is verifiable by arithmetic with no run.

### The defect

`rfx/api/_sparams.py:2251` (and an identical second copy at `:3234`):

```python
for k in range(meta["k_lo"], meta["k_hi"] + 1):
    v_f = v_f + ez_plane[:, meta["j_centre"], k] * float(dz_arr[k])
```

`k_hi` comes from `_msl_yz_cells` -> `position_to_index(z_hi=h_sub)` =
`round(h_sub/dx)` (`rfx/grid.py:201-203`). And `Ez[i,j,k]` occupies the z-edge
`[k*dz, (k+1)*dz]`. So for an n-cell substrate the ground-to-trace potential difference
needs **n** edges (k = 0..n-1), while `range(k_lo, k_hi+1)` sums **n+1**:

| dx | h_sub/dx | substrate cells | edges needed | `k_hi` | edges summed |
|---|---|---|---|---|---|
| 84.67 um | 2.9999 | k=0..2 (3) | 3 | 3 | **4** |
| 80 um | 3.1750 | k=0..2 (3) | 3 | 3 | **4** |
| 50 um | 5.0800 | k=0..4 (5) | 5 | 5 | **6** |

`round()` returns exactly n for any integer-ish cell count, so this is systematic at
every mesh size, not a rounding edge case. The extra edge `k=n` spans the trace/air cell.

### Why the extra edge is not harmless

The trace is one cell thick, so it has no PEC z-neighbour and `apply_pec_mask` leaves
`Ez` there alone — **by design**, and documented at `rfx/boundaries/pec.py:90-93`: *"For
thin PEC sheets (1 cell thick), only tangential E-components are zeroed; the normal
component is preserved (represents surface charge)."*

That is the right boundary condition. Summing it into a ground-to-trace line integral is
not. Measured: `|V_extra / V_3edge| = 0.1204-0.1212` at `angle -177.4 deg to -179.3 deg`
— a near-real **negative** contribution, frequency-flat, so `V_3edge/V_4edge = 1.137`.

### The independent conviction: modal V*I vs Poynting flux

Flux is extractor-independent, so it does not rest on any impedance anchor.
`Re(V*conj(I)) / flux_spectrum`, 7 planes from x = 3.979 to 11.092 mm x 12 frequencies:

| V integration path | ratio |
|---|---|
| substrate only, 3 Ez edges | **1.0057 - 1.0090** |
| the code's path, 4 Ez edges | **0.8806 - 0.8850** |

Normalisation witness first: a hand-formed plane-probe Poynting integral divided by the
library's `flux_spectrum` = 1.0000 at all 7 planes x 12 frequencies (1.0001 at the last),
because the DFT-plane-probe and flux-monitor accumulators use the identical kernel
(`rfx/simulation.py:1466` vs `:1508-1518`). So the ratio is convention-free and the 0.5
time-average prefactor cancels on both sides. Net x-flux is plane-independent to <= 0.25%
across the 7 planes, so the line really is lossless and the flux is a clean reference.

**The modal V*I pair is complete to +0.85% with the substrate-only V, and under-accounts
the true Poynting flux by 11.5-11.9% with the shipped V.**

Settling witness: -110.0 to -114.0 dB. Not truncation.

### What this invalidates — including two of my own posted conclusions

**Retraction 1 (#507).** I reported that the line's Zc is 38.75 ohm "established by THREE
routes agreeing" (N-probe least-squares fit, the parameter-free two-plane invariant, and
their mutual agreement to 0.02 ohm) and that the analytic Hammerstad-Jensen anchor of
47.895 ohm "is the OUTLIER". **All three routes consume the same V.** They agree because
they share a common-mode factor of 0.877, not because they are independent. Measured
two-plane invariant on the mid-line planes:

- with the shipped 4-edge V: **38.45 - 39.19 ohm** (reproduces the 38.75 I reported to 0.1%)
- with the substrate-only 3-edge V: **|Zc| = 44.27 - 44.33 ohm at angle 0.11-0.38 deg** (essentially real)

44.3 vs 47.895 is **-7.5%**, which sits inside this fixture's own preflight advisory:
*"only 3 substrate cell(s) in z ... Yee staircase error >5% expected"*, alongside
*"lateral clearance to -y absorbing boundary = 508µm < recommended 508µm ... Z0 may be
biased HIGH"*. Corroborating: measured `eps_eff` from the mid-line forward-wave phase is
3.005-3.007 vs HJ 2.869, +4.7%, same direction. **The anchor was approximately right; the
19% gap was an extractor artifact.**

**Retraction 2 (#487).** I closed #487 on the conclusion that its "coarse-mesh |S11|
floor" was substantially a normalisation artifact rather than a resolution limit. That
conclusion rested on the contaminated Zc. Reopening.

### An earlier measurement that this explains, and which corroborates it

For #504/#506 I measured a one-cell/two-cell discontinuity and could only describe it:
`Re(Z0)` = 38.77 ohm with a 1-cell trace vs 41.86 ohm with 2 cells, and I noted the
*thicker* strip reading *higher* impedance is backwards for a fringing-capacitance effect.
The mechanism is now clear. With a 2-cell trace at k=3,4, `mask_ez[3] = pec[3] & (pec[2] |
pec[4])` is **true**, so the spurious edge *is* zeroed and V loses the bad term — which is
why Z0 rises toward the corrected 44.3. That was a symptom of this defect recorded a day
before the cause was found.

### Blast radius

Everything `compute_msl_s_matrix` reports: `Z0`, `S11`, `S21`, the N-probe fit, the
two-plane invariant, and the honesty guard's `_Z0_TOL` deviation. Both loop copies are
affected (`:2251` pure-MSL lane, `:3234` ladder/mixed lane). 23 files exercise this
extractor, including `validation/crossval/{06b,07}` and the `msl_notch_e4` fixture
(recorded `re_z0_median_ohm = 31.38`, which should be re-read after a fix).

Note this is very likely also implicated in the Sheen benchmark's reference-impedance
puzzle (#507): the Z0 at which rfx matches openEMS there is 46.5 ohm against openEMS's own
reported 51.35 — a -9.5% gap in the same direction and of the same size as this defect.

### Relationship to #507

Two independent defects, both real:

1. **This one** — the V path. Dominant: ~12% on V, 11.6% on power, and it contaminated the
   impedance arbitration.
2. **The single-ratio rule** (#507) — `S[j,i] = b_j/a_i` is only valid when `a_j = 0` at
   passive ports, and the passive port reflects ~20% in amplitude. Independently confirmed
   there: the exact algebra `b1/a1 = S11 + S12*(a2/a1)` holds to 3.8e-16, and
   `|S12*(a2/a1)| = 0.1938` against the shipped `|S11| = 0.1985` — essentially the entire
   reported reflection is the far port's echo.

Fixing only one of them would move the numbers without making them right. They should be
fixed together, with the corrected V established first, because #507's reference-impedance
question cannot be settled while V is biased.

### Suggested falsifiers for the fix

- `Re(V*conj(I)) / flux_spectrum` must be 1.000 +/- 0.01 at several planes (it is the
  extractor-independent oracle, and it is what caught this).
- Two-plane `Zc` must come out essentially real (measured `angle 0.11-0.38 deg`) and
  within the documented staircase envelope of the analytic anchor, at more than one dx.
- `Zc` must stop depending on trace-cell thickness: 1-cell and 2-cell stamps should agree
  after the fix, where today they differ by 3.1 ohm.

bk-squared:
Closes #511. Closes #507.

Two independent defects in `compute_msl_s_matrix`, fixed together because fixing either
alone moves the numbers without making them right. Both were systematic, and both were
invisible to the existing tests because each was self-consistent with itself.

## #511 — the modal voltage integrated one Ez edge too many

`range(k_lo, k_hi + 1)` summed `n+1` edges over an `n`-cell substrate, at every mesh size.
`k_hi` is the **trace** cell: `_msl_yz_cells` builds an inclusive span up to
`position_to_index(h_sub) = round(h_sub/dx)`, and `k_top = k_hi` is where the PEC search
starts. Since `Ez[k]` occupies the z-edge `[k·dz, (k+1)·dz]`, ground→trace-underside is
`k_lo..k_hi-1`.

The extra edge lies inside the one-cell PEC trace, where `apply_pec_mask` **deliberately**
preserves the normal E component as surface charge (`rfx/boundaries/pec.py:90-93`). Two
correct-in-isolation decisions composing into a defect: the boundary condition is right,
summing it into a ground-to-trace potential difference is not. Measured contribution:
`|V_extra/V| = 0.120-0.121` at `∠ −177°…−179°` — near-real and negative, so V read ~12%
low.

**The oracle, which is extractor-independent** — `Re(V·conj(I)) / flux_spectrum` over
7 planes × 12 frequencies, with a normalisation witness of 1.0000 against the library's own
`flux_spectrum` (the DFT-plane-probe and flux-monitor kernels are identical, so the ratio is
convention-free):

| V path | ratio |
|---|---|
| shipped, 4 edges | 0.881 – 0.885 |
| substrate-only, 3 edges | **1.006 – 1.009** |

## #507 — the off-diagonal used the single-ratio rule

`S[j, d] = b_j / a_d` is the d-th column of S only when `a_j = 0` at every passive port.
Measured `|a_passive/a_driven| = 0.07-0.51` across three fixtures. The exact algebra
`b_1/a_1 = S11 + S12·(a_2/a_1)` holds to 3.8e-16, and `|S12·(a_2/a_1)| = 0.1938` against
the shipped `|S11| = 0.1985` — so essentially the whole reported reflection was the far
port's echo returning through the thru.

Every port is already driven (`:2140`), so the wave system is square and was simply never
solved. Replaced with `S = B·A⁻¹` over all drives — the same algebra as the coax lane's
`solve_two_port_from_wave_amplitudes` (#489), generalised to n ports, with a `cond(A)`
degeneracy warning and a warned fall-back if the solve goes non-finite.

## Verification

**Pre-declared falsifier, measured after** (committed
`scripts/diagnostics/thin_conductor_cell_thickness_probe.py`, dx = 84.67 µm):

| quantity | before | after | predicted before running |
|---|---|---|---|
| Re(Z0) | 38.77 Ω | **44.11 Ω** | 44.1 ± 1.5 |
| mean\|S21\| | 1.0195 | **0.9991** | ≤ 1 |
| mean\|S11\| | 0.2233 | **0.0501** | — |

44.11 against the flux-corrected two-plane invariant's 44.27–44.33 is **0.4%**, from a
different estimator measured in a different session — neither arranged to check the other.
Settling −110.0/−113.3 dB, identical to before, as it must be since both fixes are
post-processing of the same recorded fields. Preflight unchanged; the load-bearing
advisories remain the 3-substrate-cell staircase warning and the ±y lateral clearance.

The residual reflection is now consistent with the reference mismatch alone:
`Γ = (44.11 − 47.895)/(44.11 + 47.895) = −0.0411` against a reported mean `|S11|` of
0.0501. And 44.11 vs the anchor is **−7.9%**, inside the fixture's own documented staircase
advisory — which is the retraction posted in #507 confirmed end to end: *the analytic anchor
was approximately right all along*, and the 19% gap I had attributed to it was this defect.

**New tests** — `tests/test_msl_modal_voltage_and_wave_solve.py`, 18 tests, no FDTD.
Both defects have a mutation twin, and **both were verified to fail on revert**: reverting
the V range reds 7 tests, reverting the solve reds 7 tests, and the `γ=0` case correctly
stays green because the two assembly rules agree there — so the tests discriminate on the
mechanism rather than on a magnitude.

**Regenerated** `tests/fixtures/msl_replay_golden_f64.npy` from the **stored** accumulators
(raw field data untouched) after applying the same two fixes to the numpy reference in
`scripts/capture_msl_replay_fixture.py`. That fixture exists to prove jnp/numpy structural
equivalence, so both sides must implement the same algorithm — regenerating it against an
unchanged reference would have baked in a false mismatch. Column power on it:
1.0038–1.0121 → **1.0000–1.0007**. `test_replay_float64_equivalence` passes.

**The committed gate passes, and no bound was touched.**
`test_msl_thru_line_passive_gate` (dx = 80 µm, `slow`):

| gate | bound | measured | |
|---|---|---|---|
| mean \|S11\| | < 0.15 | 0.0746 | PASS |
| mean \|S21\| | > 0.90 / < 1.05 | 0.9970 | PASS |
| mean Re(Z0) | (40, 65) | 42.41 | PASS |

`git diff` on that file shows **no assert line changed** — only the docstring's recorded
calibration, which had gone stale (0.118 / 0.972 / ~57 Ω) and whose prose attributed the
~57-vs-48 Ω gap to "the documented Yee-staircase bias". Most of it was these two defects.

**The headline, as corrected by the review (F2) and re-measured.** An earlier revision of
this body claimed the fix collapsed the cross-mesh Z0 spread "48.5% → 3.9%", with a
coverage-fraction story explaining the residual. **That narrative is retired — it described
the F2-defect state**, in which the span was one edge short of the trace it was nominally
measuring, and its apparent cross-mesh agreement (44.11 vs 42.41) was cancellation.

With the span anchored on the rasterized trace node (commit `a5acfa9`), the honest picture:

| | dx = 84.67 µm (aligned) | dx = 80 µm (bisecting) |
|---|---|---|
| rasterized geometry | trace at z = 254 µm | trace at z = **320 µm** (254 µm dielectric + 66 µm air gap) |
| Re(Z0) | 44.11 Ω | 57.58 Ω |
| flux identity `Re(V·I*)/flux` | 1.006–1.009 (measured) | expected ~1 by construction (V and I share one conductor plane); not yet measured on this mesh |

The two meshes rasterize **genuinely different structures** — a strip 66 µm higher over an
air gap SHOULD read a higher Z0 — so the cross-mesh difference is geometry quantization,
which preflight's mixed-cell danger-zone warning already describes. The extractor now
reports each mesh's own rasterized geometry faithfully, with V and I anchored on the same
conductor plane by construction.

**Post-F2 measurement scorecard** (falsifiers declared before the jobs returned):

- **P1 HELD**: aligned-mesh probe bit-unchanged (1-cell 44.11 Ω, 2-cell 41.86 Ω) — trace
  anchoring is a no-op where the trace node equals the rounding proxy.
- **P2 HELD**: dx = 80 gate returns to 57.58 Ω / mean|S11| 0.1160, all four bounds PASS,
  none touched. (The pre-#511 inclusive span was accidentally correct at this mesh; the
  anchored span reproduces it, now for the stated reason.)
- **P5 gate PASS, prediction FALSIFIED and disclosed**: cross-length Z0 spread is **4.96%
  against the 5% bound — 0.04pp of margin**. I predicted a return to the sub-1% class;
  wrong. The L = 6 mm leg is the outlier under BOTH spans (44.0 vs 42.2/42.4 before; 60.2
  vs 57.3/57.6 after), so the spread is a property of the short-line fixture's fit
  conditioning, not of the span. This lock is now effectively margin-less on the bisecting
  mesh; follow-up filed rather than the gate touched.
- **P6**: NU gate PASS with output **bit-identical to clean main** (max|S11| 0.9814,
  in-band min 0.9044 vs the 0.70 floor — margin 0.20; Z0 median 57.64). Explained: the NU
  fixture is 1-port (the solve degenerates to b/a) and its trace also rasterizes one node
  above the proxy, so main's inclusive span was accidentally correct there and the
  anchored span reproduces it. **Correction of my own earlier comment**: the "0.0067
  razor-thin dip margin" I reported conflated the off-band match-point dip (0.7067) with
  the in-band 0.70 floor; the true in-band margin is 0.20, and the 0.7067 state was the
  F2-defect span anyway.

**Blast-radius sweep, green**: `pytest tests/ -k "msl or sparam or mixed or thin_conductor
or passivity or port"` → **737 passed, 2 skipped, 4 xfailed, 0 failed** (29:36). That
selection covers both lanes the change touches (`compute_msl_s_matrix` and
`compute_mixed_s_matrix`).

**Other gates run**: contract tests (24), `scripts/check_api_reference.py`, `ruff check rfx/
tests/`, `tests/test_thin_conductor_honesty.py` (10), `tests/test_msl_sparam_ad.py` (3,
including the replay structural-equivalence test).

**Not run**: the whole fast suite in one process. I tried and it was killed at the 50-minute
timeout with no output — because passing `-m` on the command line *replaces* the
`pyproject.toml` addopts, so my invocation silently included `slow_physics` (which the
default lane excludes) and overran. That attempt produced **zero** coverage and I am not
counting it. The sharded default lane is CI's job; the targeted sweep above is what I can
stand behind locally.

### What is NOT verified, stated plainly

**Gradient accuracy is unverified, and it turns out it could not have been verified
locally.** While checking whether the solve's matrix inverse disturbs the AD path I found
that `test_compute_msl_s_matrix_ad_smoke_has_finite_gradient` **passes on a gradient of
exactly `0.0`** — it asserts only `isfinite` and `not isnan`, so a severed tape passes.

**CORRECTED after review (F3).** I originally wrote here that the 0.0 gradient "is true on
`main` too, so it is not a regression". That was false: clean `main` prints
`grad = 3.912601e-01`; this branch prints `0.000000e+00`. My "main" comparison had run both
invocations inside the worktree (a `cd` at the start of one shell call persisted across the
second pytest), so I compared the branch against itself. The vacuity of the smoke test is
**introduced by this PR's assembly change**, not inherited — with the caveat, measured by
the review, that the forward objective's finite-difference slope is also ~0 under the new
solve (θ = 0.6/0.7/0.8 → −2.85e-08 / 0.0 / −2.85e-08), i.e. the synthetic fixture genuinely
maps to near-zero sensitivity rather than the tape being severed; unit-level
`jax.grad` through `msl_solve_s_from_waves` matches central FD to 1.165e-9. Details and the
fix plan: #515. Measured sensitivities at `θ = 0.7` on this branch:

| objective | dS/dθ |
|---|---|
| `Re S21` — the test's own objective | `0.000000e+00` |
| `abs S11` | `1.30e-08` (f32 rounding against `S ~ O(1)`) |

And the lane's only genuine gradient gate, `tests/test_msl_ad_fd_converged.py`, is marked
**both** `gpu` and `slow`, so it runs in neither the fast lane nor the CPU slow lane — only
in the VESSL GPU suite. Filed as #515.

So: this change is not shown to have broken AD, and is **not shown to have preserved
gradient accuracy**. PR #486's "0.011% AD-vs-FD converged" claim needs
`scripts/vessl_gpu_suite.yaml` re-run before it can be repeated post-change. I have not
launched that — it is a billable external job and wants an explicit go-ahead.

## Two refactors that are part of the fix, not incidental

- `msl_modal_voltage()` and `msl_solve_s_from_waves()` extracted as helpers. The V
  integration was **duplicated** in two places (`:2251` pure-MSL, `:3234` mixed lane) with
  the same bug in both; one definition is what stops them drifting again, and it is what
  makes the defect unit-testable without a 6-minute FDTD run.
- `scripts/diagnostics/thin_conductor_cell_thickness_probe.py` now puts the checkout on
  `sys.path`. Running a script **file** never puts the cwd there, so `import rfx` silently
  resolved to the installed package — this probe reported pre-fix numbers from a post-fix
  tree, and the check I ran to rule that out used `python -c`, which is the one invocation
  form that *does* put cwd on the path.

## A partial fix deliberately refused (#514)

`rfx/probes/msl_wave_decomp.py` carries a parallel plane-primitive path with the **same V
off-by-one and** the pre-#80 single-slab Hy current (which `msl_loop_current`'s own
docstring records as undercounting I by ~1.5×). Since `Z0 = V/I`, V 12% low and I ~33% low
partially cancel — so fixing only V there would move that path *further* from the truth.
Warning added in place, filed as #514; repairing it properly needs Hz plane probes that
module never registers. Nothing claims-bearing rides on it today.

## Correction to the issue text

`|S11|² + |S21|² = 1 + |S11|²` is **not** a general identity of the single-ratio rule, as
#507 states it. Expanding for a symmetric S gives
`(1+γ²)(|S11|²+|S21|²) + 4γ·Re(S11·conj(S21))`, so the relative phase decides the sign and
the old rule can push column power either side of 1. The identity holds in the
**near-matched** case (true S11 ≈ 0), which is what the thru fixtures are. Both cases are
now pinned by separate tests — the failing test is what taught me this.

## R3

`R3: memory=project_msl_extractor_two_defects + feedback_agreement_is_not_independence + feedback_stale_editable_install_shadow (and rfx-known-issues.md:844-845, which already records the static-Z0 split as "a regression AWAY from de-embedding" with a concrete fix — this executes a recorded item) | R2-attempts=1 per defect, both closed against pre-declared falsifiers | falsifier=Re(V·conj(I))/flux_spectrum must stay within 1.00±0.01, and both mutation twins must red on revert`

Authored by me; **needs an independent reviewer pass** — I am not approving my own
extractor change, and two of my conclusions in #507 were already retracted this session.

---

## Author-pass response to the review (REQUEST_CHANGES)

The independent review's findings were verified and acted on; commits after `98032e7`:

- **F2 (accepted, fixed — commit `a5acfa9`)**: the V-span top is now anchored on the
  **rasterized trace node** (`trace_k_per_port[p][0]` — the same PEC search the
  Ampère-loop current uses) in both lanes, not on the `round(h_sub/dx)` proxy, which is
  one edge short whenever `frac(h_sub/dx) ∈ (0, 0.5)`. Verified by two new wiring tests
  that drive the real assembly with z-profiled synthetic planes on BOTH mesh classes
  (aligned dx = 84.67 µm → edges 0..2; bisecting dx = 80 µm → edges 0..3, trace node 4).
  Helper docstring and the falsifier scoping rewritten. Full measurement scorecard in the
  headline section above — P1/P2 held, P5's magnitude prediction falsified and disclosed.
- **F3 (accepted, corrected above)** — and #515 updated.
- **F4 (accepted)**: the E2E historical golden is **re-baselined with the decision written
  into its docstring** — the "must NOT be regenerated" contract existed to prove a pure
  refactor precision-only, and two deliberate algorithm changes (#468 passivity default,
  #511/#507) made that premise unsatisfiable. The test is now an honest cross-version
  drift lock; #509 updated.
- **F5 (accepted)**: `reliable`-mask blast radius corrected in `probes-sparams.mdx`,
  `sparameter_support_matrix.md`, and the `_spec.py` docstring — under `S = B·A⁻¹` a
  collapsed wave pair contaminates all of `S[:, :, k]`, and the only safe per-bin filter is
  `np.all(reliable, axis=0)`.
- **F6 (accepted, annotated)**: the mixed lane's single-ratio residue is now annotated at
  `_assemble_mixed_power_wave_s` with the same class of warning as #514; follow-up filed as
  **#517** (measurement-first: form mixed S both ways on a recorded run against the 9%
  reciprocity residual).
- **F1**: all four slow gates now have verdicts (review comments); the sole red is the E2E
  golden, resolved by F4.
- Also fixed while in the file: `cond_a`'s `1e-300` floor underflowed to 0.0 in float32
  (NEP-50 weak promotion) — SVD now cast to f64 before the ratio; and the `enable_x64`
  import in the new test module is guarded like its sibling.


bk-squared:
# Review of PR #516 — `fix/511-msl-modal-voltage-off-by-one` (4 commits over main@75d84fa)

**Verdict: REQUEST_CHANGES.**

The two core fixes are real: the #511 V-span physics is PROVED for node-aligned meshes by independent derivation, and the #507 `S = B·A^-1` solve is algebraically and AD-sound at unit level. But the PR as shipped contains defects in the change itself and in its own text:

1. its body makes a **false claim** about AD coverage (the smoke gradient regressed 0.391 → exactly 0.0 under this PR; the body says "true on main too" — it is not);
2. it **silently deepens a weekly-CI red 5.6x** (E2E golden 0.0527 → 0.2927, both MEASURED-HERE) while definitionally killing that test's "precision-only" premise, with zero disclosure or reconciliation artifact;
3. the new helper's docstring-certified correctness (flux oracle 1.006–1.009, R3 falsifier "1.00±0.01") was measured only on an aligned mesh and **does not hold on the repo's own dx=80 gate and replay fixtures**, where the new span drops the last in-phase conductor-gap edge and anchors V and I on different conductor planes;
4. committed public docs (`reliable` mask column-containment) are now **actively wrong** under the new assembly and were not touched;
5. 3 of 4 decisive slow gates **never produced a verdict** (processes died without output; relaunched by this review, see below).

None of this requires abandoning the fix. The required changes are scoping, disclosure, docs, and gate verdicts — listed at the end.

---

## 1. What this review verified independently vs. what remains on the author's word

**Independently verified (derivation or MEASURED-HERE, not taken from the PR):**

- Ez[i,j,k] occupies the z-edge [k·dz,(k+1)·dz] — derived from the actual `update_h`/`update_e` curl staggering in `rfx/core/yee.py`, corroborated by the `pec.py` ghost-cell comment. `apply_pec_mask` measured directly: a 1-node PEC sheet leaves the sheet-node Ez alive; a 2-node stamp zeroes it. This confirms the −12%@~180° mechanism and the "2-cell unchanged" claim.
- The flux oracle **replicated by an independent pure-numpy re-derivation** from the author's dump (own implementations of both V variants, the Ampère loop, and the plane Poynting flux): old 0.8805–0.8850, new 1.0057–1.0090 over 7 planes × 12 freqs; flux-normalisation witness 0.9990–1.0003 — **on the aligned dx=84.67 µm mesh only**.
- Solve transpose algebra verified numerically on non-symmetric, non-reciprocal planted S: recovery to 4.4e-16 (2-port) / 1.2e-15 (3-port); the transposed alternative differs by ~2. Unit-level `jax.grad` through `msl_solve_s_from_waves` matches central FD to 1.165e-9 relative (f64; skeptic's independent fixture: 1.7e-10).
- Fixture hygiene at git level: `msl_replay_accumulators.npz` **bit-identical to main** (blob `0c3f894e…`); only `msl_replay_golden_f64.npy` changed (max|dS| = 0.2915 vs main's; column power 1.0038–1.0121 → 1.0000–1.0007, matching the PR verbatim); regenerating the golden from the stored accumulators reproduces the committed file **bit-identically**.
- `test_msl_port_integration.py` diff is docstring-only (33+/6−, zero assert lines). The dx=80 thru passive gate re-measured: mean|S11| = 0.0746, mean|S21| = 0.9970, mean Re(Z0) = 42.41 ohm — exactly the author's calibration (preflight context quoted in section 4).
- 18 new unit tests + 3 fast replay/AD tests pass (MEASURED-HERE, 1.4 s + 15.8 s). The author's "737 passed" sweep log verified and its collection reproduced (742/3196, 2454 deselected; slow tests excluded by default addopts exactly as admitted).
- R3 audit artifact present in commit 278977c.

**Still on the author's word (unverified):**

- Mutation-twin "red on revert" (7+7) — verifying requires mutating tracked files, forbidden here; committed guards cover the wiring except solve orientation (all committed fixtures are symmetric — a shared S-vs-S^T bug is undetectable).
- Thin-conductor probe numbers 44.11 / 41.86 (consistent with the gate numbers but not re-run).
- The docstring range "|V_extra/V| = 0.120–0.121" — **unreproduced**; re-derivation from both available dumps gives 0.1215–0.1279. Substance (~12% anti-phase) confirmed; the 3-decimal committed range is not.
- End-to-end gradient accuracy (#515) — the only real gate is gpu+slow, run in no scheduled lane.

---

## 2. Findings, ranked (skeptic verdict / file:line / action)

**F1 — BLOCKING (CONFIRMED): decisive slow gates had no verdict at cutoff; the one since harvested FAILED 5.6x worse than main's pre-existing red.**
`tests/test_msl_sparam_ad.py:387`. All four Lens-E gates were still mid-FDTD at the forced cutoff. Since then: the E2E golden harvested as **FAIL at 2.927e-01** (PR) vs **5.266e-02** (clean main, MEASURED-HERE — see section 4); the other three processes died without ever writing pytest output. **This review relaunched them detached** (`/tmp/lensF_thru_invariance.log`, `/tmp/lensF_nu_gate.log`, `/tmp/lensF_issue80_patch.log`; workers 566598/567283/567965 confirmed running). Action: harvest all three before merge; attribute any red against clean main.

**F2 — MAJOR (CONFIRMED): V span drops the last substrate edge whenever the trace rasterizes one node above k_hi — including the repo's own dx=80 gate and replay fixtures.**
`rfx/api/_sparams.py:129` (helper :91–131; call sites :2360/:3419; trace search :2162–2182). Box rasterization is half-open over NODE coordinates, so for frac(h_sub/dx) ∈ (0, 0.5) the trace lands at node ceil(h_sub/dx) = k_hi+1 while k_hi = round(h_sub/dx). MEASURED-HERE through the actual pipeline at dx=80 µm (h_sub/dx = 3.175): k_hi=3, PEC trace node **4**; production `trace_k_per_port` returns [(4,4),(4,4)] while `msl_modal_voltage` is called with k_hi=3 — **V and I reference different conductor planes**. Edge decomposition of the committed fixture: edges 0–3 carry 0.238/0.244/0.254/**0.264** all in-phase (edge 3 has full substrate eps 3.66); the anti-phase surface-charge edge is edge **4** (0.127 at ∠179.5°), which neither span includes at dx=80. |V_old/V_new| = 1.3578 = 57.58/42.41 exactly; the docstring's −12%/flux-1.006 justification is the aligned-mesh signature and is **arithmetically impossible** at dx=80. With fixed analytic Z0_hj = 47.89, Γ moves +0.092 → −0.061 — the dx=80 gate improvement (0.118 → 0.0746) is substantially **cancellation**, not demonstrated extractor accuracy at that mesh. Preflight warns only for frac ∈ [0.10, 0.40] and describes a different failure mode; frac ∈ (0,0.10)∪(0.40,0.5) gets nothing; none of the 18 new tests covers trace-at-k_hi+1. Action: anchor the top of the V span on `trace_k_per_port[p][0]` (already computed in the same function), or at minimum warn when the PEC search lands above k_hi; re-scope the docstring and the R3 falsifier to aligned meshes (on dx=80 the honest flux ratio is ~0.74, not 1.00 — a future reader re-running the stated falsifier on the gate mesh would misdiagnose a regression).

**F3 — MAJOR (CONFIRMED): PR-body AD claim is false as stated; the fast lane's only AD signal regressed 0.391 → exactly 0.0 under this PR.**
`tests/test_msl_sparam_ad.py:313`. MEASURED-HERE: main prints `[AD smoke] grad = 3.912601e-01`; PR prints `[AD smoke] grad = 0.000000e+00`. The test file is byte-identical between the two, so the change is entirely the extractor. The PR body's "passes on a gradient of exactly 0.0 … That is true on main too, so it is not a regression here" is false on the value claim. Not a severed tape: forward objective at θ=0.6/0.7/0.8 = −2.85e-08 / 0.0 / −2.85e-08, so FD is also ~0 — the multi-drive solve maps this synthetic fixture to Re(S21)≈0 with ≈0 sensitivity, and the test asserts only isfinite/not-isnan. Mitigating: the helper itself is AD-sound (unit AD vs FD 1.165e-9). Action: rebuild the smoke fixture for genuine θ-sensitivity with a nonzero gradient floor; correct the record on #515 and in the PR body — the vacuity was introduced here, not inherited.

**F4 — MAJOR (CONFIRMED twice, by two independent lenses): E2E historical golden deepened 0.053 → 0.293 with a now-false docstring and no reconciliation, undisclosed.**
`tests/test_msl_sparam_ad.py:350`. The red is **pre-existing on main — #509, red weekly since 2026-07-06** — but this PR deepens it 5.6x (numbers in section 4) and definitionally kills the docstring's premise: "any end-to-end gap is precision-only, not a logic divergence" cannot survive two deliberate algorithm changes (max rel deviation 4.79 is not f32 rounding). Worse, the PR regenerated `msl_replay_golden_f64.npy` to the new algorithm, so `test_replay_float64_equivalence` now certifies new==new while the E2E golden stays an old-algorithm capture — the structural-proof chain the docstring cites is broken by the PR's own design. The PR body's "Not run" section does not name this test; grep of the full body finds zero hits for `historical`, `msl_s_matrix_golden`, `509`. The author's own #509 comment forbids quiet regeneration. Action: commit a decision artifact with the merge — xfail with explicit #509/#511/#507 reason, or re-baseline with written justification superseding the "must NOT be regenerated" contract — rewrite the "precision-only" docstring either way, and update #509 with the measured 0.293.

**F5 — MAJOR (CONFIRMED): public `reliable`-mask semantics are now wrong under S = B·A^-1.**
`docs/public/guide/probes-sparams.mdx:202` (+ `docs/guides/sparameter_support_matrix.md:144`, `rfx/api/_spec.py:~1289`). The doc guarantees a false entry contaminates only `S[:, p, k]`; under the multi-drive solve a bad wave pair in any drive contaminates all of `S[:, :, k]`. Demonstrated MEASURED-HERE on synthetic waves: corrupting only drive 0 at one bin (near-null collapse) leaves everything finite (fallback silent), **cond(A) = 59.2** (below the 1e3 warn threshold — the new guard is silent), and moves the drive-1 column by |ΔS| = 0.30/0.68 at that bin. No doc, docstring, or support-matrix line was updated; the `np.all`-over-ports filter example happens to remain the only safe one. Action: update the three locations; state the per-bin blast radius; consider a per-bin fallback or lower cond threshold.

**F6 — MAJOR (CONFIRMED): the #507 single-ratio defect class survives, unannotated, in the mixed-port lane.**
`rfx/api/_sparams.py:754` (docstring; implementation `_assemble_mixed_power_wave_s` ~:832–870 — note the finding's original `:3966` cite is the coax lane, corrected by the skeptic). The mixed lane received the #511 V fix (:3419) but still forms `S[i,j] = b_i/a_j` per drive; the driven-MSL diagonal is the exact #507 quantity and leaks into the **default flux-channel magnitudes** via `P_inc = P_net/(1−|S_jj|^2)`. The PR set its own precedent by annotating the identical divergence in `msl_wave_decomp.py` (#514) and did not extend it here. Exposure bounded: lane is fenced experimental (#488 arc) with a running reciprocity witness. Action: add the same class of annotation; file a follow-up for extending the multi-drive solve (V/I per (run, port) is already available to the assembly) or documenting why the Kurokawa assembly cannot use it.

**F7 — MAJOR (CONFIRMED): msl_notch_e4 / Palace-referee fixture chain is green-but-stale with no committed producer.**
`tests/test_msl_notch_e4_comparison_gates.py:24`, `tests/test_msl_notch_palace_referee_gates.py:95`. These gates run **no FDTD** (13 passed in 0.09 s) over an rfx JSON produced by the superseded extractor (PR #263, provenance verified). Margins are thin: s21 mean-diff 0.105 vs gate 0.13 (0.025 margin) while the isolated extractor shift on identical field data is |S21| max 0.0119 / |S11| max 0.117 (MEASURED-HERE from old-vs-new replay goldens) — near the notch (strong reflection) the shift plausibly exceeds the margin. `validation/crossval/06b` cannot regenerate the leg (different geometry); the matched geometry survives only as prose in the fixture meta. Action (follow-up, not blocker): regenerate the rfx dx=50 leg on the new extractor, commit a producer this time, recommit fixture+summary together; until then the E4 "characterized" claim describes the superseded extractor.

**PLAUSIBLE-UNVERIFIED overflow (7 findings).** Seven further finder-stage major findings never reached skeptic verification before cutoff. They are carried as PLAUSIBLE-UNVERIFIED — neither confirmed nor dismissed — and their full texts exist only in the session lens logs. Per the lens summaries they cluster on: `sparameter_support_matrix.md` stale numeric claims (48.6-ohm notch Z0, "AD validated 0.000110"); the broad-E5 envelope WARNING whose fencing rests on now-retracted evidence (PR's own probe measured 0.0501 clean-alignment vs the "0.10 UNREACHABLE" claim); `scripts/diagnostics/replay_msl_3probe_dump.py` as a dead evidence lane (still single-ratio assembly + schema-v1 key); and `compute_mixed_s_matrix` end-to-end unmeasured with its promised slow-lane battery absent. Do not treat any of these as established without a verification pass.

**Minors/notes (unverified by design; most decision-relevant):** aligned-mesh-only R3 falsifier stated unscoped (~0.74 implied on dx=80 — F2 corollary); k_hi<=k_lo ValueError fires only post-run though indices are known pre-run (`_sparams.py:2102`); no committed test discriminates S vs S^T (all fixtures symmetric; shared bug cancels in the replay equivalence); unguarded `from jax.experimental import enable_x64` in the new test module (removed in JAX 0.8.0; sibling file has the fallback; recorded env-drift incident class); cv06b/cv07 committed rfx JSONs describe the superseded extractor with nothing marking them superseded, and **no scheduled CI lane runs any MSL crossval case** (weekly lane = 01/02/03/04/09/10 only); `raw_3probe` dump metadata still attributes production S to the N-probe fit with `schema_version` unbumped despite the semantic change; thru-gate Re(Z0) margin over the untouched (40, 65) floor shrank 17 → 2.41 ohm; three warning strings still describe the pre-#507 design; two issue-80 diagnostics scripts still carry the #511 inclusive span unannotated; the flux-oracle numbers are pinned by **no committed artifact** (docstring prose only — `research/CLAUDE.md` requires a named artifact for claims-bearing numbers); "0.4% independent Z0 cross-check" overstates independence (both estimators share the 3-edge V; same-morning artifacts).

---

## 3. Slow-gate table (the spine): branch vs. main attribution

All runs from inside the checkout with `-m "" -q -s -p no:randomly -p no:cacheprovider`; box has no CUDA jaxlib (CPU fallback).

| Gate (all `slow`) | PR branch (/root/wt_i511) | Clean main (75d84fa) | Attribution |
|---|---|---|---|
| `test_compute_msl_s_matrix_end_to_end_matches_historical_base` | **FAIL** — `max_abs_dev = 2.927e-01`, mismatched 30/40 (75%), max rel 4.786385, vs rtol 5e-3/atol 2e-3; 429.79s. MEASURED-HERE (`/tmp/lensE_e2e_golden.log`; independently reproduced by skeptic at 411.66s, identical numbers) | **FAIL** — `max_abs_dev = 5.266e-02`, mismatched 38/40 (95%), max rel 0.71370162; 411.26s. MEASURED-HERE (scratchpad `e2e_main.log`; matches #509's CI record 0.05266097) | **Pre-existing red on main (#509, since 2026-07-06). PR deepens it 5.6x**; neither the test nor the golden is touched by the PR, so the amplification attributes to the PR's `rfx/` changes — expected in direction (deliberate extractor change vs frozen golden), undisclosed in magnitude |
| `test_msl_thru_line_passive_gate` (dx=80) | **PASS** — mean\|S11\| = 0.0746, mean\|S21\| = 0.9970, mean Re(Z0) = 42.41 ohm; 400s. MEASURED-HERE, matches author's calibration; no assert changed | not run (docstring-only diff; main baseline in docstring: 0.118 / 57.58) | Improvement is real at the gate level but per F2 substantially cancellation on this bisecting mesh |
| `test_msl_thru_line_z0_length_invariance_and_positive_sign` | **NO VERDICT** — original process died without output; relaunched, pending in `/tmp/lensF_thru_invariance.log` | not run | UNMEASURED. Risk flagged: 42.41 sits 2.41 ohm above the untouched 40-ohm floor |
| `test_msl_nu_sparam_gate.py` | **NO VERDICT** — died without output; relaunched, pending in `/tmp/lensF_nu_gate.log` | not run | UNMEASURED (~900k-cell NU run) |
| `test_issue80_patch_s11_regression.py` | **NO VERDICT** — died without output; relaunched, pending in `/tmp/lensF_issue80_patch.log` | not run | UNMEASURED (num_periods=280, ~88k steps). 1-port, so #507 is a no-op; #511 active |

Preflight/honesty context for the E2E numbers, quoted verbatim per repo rule (both runs printed `[run] preflight found 7 advisory issue(s) - pass skip_preflight=True to suppress`; no settling-dB witness line is printed by this test — the warnings below are the run's own honesty output):

- Both: "MSL port 'msl_0': only 3 substrate cell(s) in z (h_sub=254µm, dx=80µm). Yee staircase at dielectric interface is O(dx) — Z0 staircase error >5% expected." and "h_sub/dx = 3.175 (fractional part 0.175) lands in the [0.10, 0.40] mixed-cell danger zone."
- Both: "standing-wave null at the port plane: 2 bins in [4.5000, 5.0000] GHz have |V|,|I| below 10% of band median — wave-split S-parameters are unreliable there".
- PR: "reported Z0 for MSL port 'msl_0' = 42.23 ohm deviates 11.8% from analytic Hammerstad-Jensen 47.89 ohm"; "10 of 10 frequency bins were non-passive as extracted, worst sigma_max = 1.001 at 3.500 GHz."
- Main: "reported Z0 for MSL port 'msl_0' = 57.58 ohm deviates 20.2% …"; "worst sigma_max = 1.102 at 5.000 GHz. 5 bins exceeded the 1.05 extraction envelope — at those bins the RAW value is a measurement artifact".

Fairness note, factual from the logs: the PR **reduces** the non-passivity envelope on this fixture (worst sigma 1.102 with 5 bins over the 1.05 envelope on main → 1.001 with none on the PR), consistent with the echo-removal mechanism being physical.

---

## 4. UNMEASURED at merge time (what merging now would be trusting)

- The three relaunched slow gates (thru Z0/length invariance, NU gate, issue80 patch) — pending in `/tmp/lensF_*.log`.
- `test_msl_ad_fd_converged` (gpu+slow) — the ONLY real gradient gate; no scheduled lane runs it. Any MSL gradient-accuracy claim (incl. PR #486's 0.011%) is unsupported under the new assembly until it runs.
- cv06b/cv07 rfx legs under the new extractor (author-admitted; committed JSONs unmarked).
- The notch-e4 / Palace-referee chain under the new extractor (F7).
- Mixed-port lane end-to-end under the new V (F6).
- Mutation-twin red-on-revert (author's word).
- Flux identity on bisecting meshes (F2 predicts ~0.74; unverified by run).
- The 7 PLAUSIBLE-UNVERIFIED overflow findings.

---

## 5. Required changes and follow-ups

**Required in this PR before merge:**
- [ ] Correct the PR body: the AD-smoke "true on main too" claim (F3) and disclosure of the E2E golden deepening (F4).
- [ ] E2E golden reconciliation artifact: xfail with #509/#511/#507 reason OR written re-baseline decision superseding the no-regeneration contract; rewrite the "precision-only" docstring; update #509 with 0.293 (F4).
- [ ] Re-scope `msl_modal_voltage` docstring + R3 falsifier to aligned meshes; anchor the V span on `trace_k_per_port` or warn when the rasterized trace lands above k_hi; add a bisecting-mesh test case (F2).
- [ ] Fix `reliable`-mask semantics in `probes-sparams.mdx`, `sparameter_support_matrix.md`, and the `_spec.py` docstring (F5).
- [ ] Harvest `/tmp/lensF_*.log`; attribute any red vs clean main before merge (F1).
- [ ] One-liner: guard the `enable_x64` import like the sibling file.

**Follow-up issues (existing):** #509 (golden decision — must be updated, not quietly bypassed), #514 (msl_wave_decomp, already filed), #515 (AD smoke — needs correction that the vacuity is PR-introduced).

**Follow-up issues (new, to file):**
- [ ] Mixed-lane #507 residue: annotation + multi-drive-solve extension or documented impossibility (F6).
- [ ] Notch-e4/cv07 rfx-leg regeneration on the new extractor + committed producer; mark the committed JSONs superseded until then (F7 + minors).
- [ ] Commit a named artifact for the flux-oracle claim (per `research/CLAUDE.md` claims-bearing rule) including a bisecting-mesh measurement.
- [ ] Skeptic-check the 7 overflow findings (support matrix, broad-E5 fencing, replay diagnostic lane, mixed-lane battery).
- [ ] `raw_3probe` metadata/schema_version + stale warning strings; S-vs-S^T discriminating test (one non-symmetric planted-S case).

*Reviewer of record; the author (a different model in this session) cannot self-approve. All numbers marked MEASURED-HERE were measured or independently reproduced by this review's lenses/skeptics; everything else is labeled unverified.*

bk-squared:
## F1 resolved: all three relaunched slow gates have verdicts

Completing the review's slow-gate table (section 3). All three runs harvested from the
detached processes the review left running (`/tmp/lensF_*.log`), PR branch, CPU,
`-m "" -p no:cacheprovider`:

| Gate | Verdict | Key numbers | Margin note |
|---|---|---|---|
| `test_msl_thru_line_z0_length_invariance_and_positive_sign` | **PASS** (18:08) | Z0 per length [44.00, 42.23, 42.41] Ω; mean\|S11\| 0.0164/0.0472/0.0746; ±x symmetry ≤0.3% | cross-length spread **4.14%** vs the 5% gate — **0.86pp margin**, and the pre-PR record for this lock was 0.49%. The widened spread is consistent with F2: L=6mm reads 44.0 while L=8/10 read 42.2–42.4 on the bisecting dx=80 mesh |
| `test_msl_nu_sparam_gate` | **PASS** (21:53) | max\|S11\| = 0.9843; dip @ 11.100 GHz, \|S11\| = 0.7067 | dip margin over the 0.70 edge-fed floor is **0.0067** — razor thin. Must be re-measured after any F2 re-anchoring |
| `test_issue80_patch_s11_regression` | **PASS** (31:40) | max\|S11\| = 0.9917; dip @ 11.300 GHz (off-resonance match point, outside the [9.0, 9.42] resonance band as required) | 1-port fixture, so #507 is a no-op here; #511 active |

Updated verdict on the slow lane as a whole:

- **4 of 5 slow MSL gates PASS on the PR branch** (thru passive, length invariance, NU, patch).
- **1 FAILS**: the E2E historical golden at 0.293 vs main's pre-existing 0.053 (F4). That
  remains the one red, and it needs the reconciliation artifact the review requires — it is
  a frozen old-algorithm capture, not a physics gate, but the decision to retire or
  re-baseline it must be written, not implied.

This closes F1. The remaining required changes are unchanged: F2 (V-span anchoring on
`trace_k_per_port` + bisecting-mesh test + falsifier re-scoping), F3 (PR-body AD claim
correction — already corrected on #515), F4 (golden reconciliation artifact + #509 update),
F5 (`reliable`-mask docs), and the two thin margins above are added context for F2: both the
4.14% spread and the 0.0067 dip margin sit on the mesh where the current span is one edge
short, so the F2 fix should be expected to move them and both gates re-run after it.

bk-squared:
# Re-review of PR #516 author-pass — delta 98032e7..a5acfa9 (+ d148ab4, the pushed PR head)

## 1. Verdict

**The author pass discharges the review. Recommendation: APPROVE_WITH_FOLLOWUPS** — anchored on **d148ab4** (current PR head), not a5acfa9 alone.

| Finding | Status | Evidence |
|---|---|---|
| F1 (slow-gate verdicts) | **Discharged pre-commit** (13:16 comment) — settled first-review item, not relitigated | PR comment timeline |
| F2 (V-span anchored on proxy, not trace node) | **Discharged** | Both call sites use `trace_k_per_port[p_idx][0]` (`/root/wt_i511/rfx/api/_sparams.py:2408` pure, `:3471` mixed); helper docstring rewritten; capture-script numpy reference mirrored (replay golden regenerates bit-identically); two new wiring tests (`test_v_span_on_aligned_mesh_stops_below_the_trace_node`, `test_v_span_on_bisecting_mesh_reaches_the_rasterized_trace`) kill both historical bad rules under mutation (k_hi±1 reds both). Full measurement scorecard P1/P2/P5/P6 now reproduced or arithmetically pinned (section 2). |
| F3 (cond_a underflow / x64 guard / verdict print) | **Discharged** | f64 cast before the 1e-300 floor with NEP-50 rationale at `_sparams.py:198-204`; guarded import; sign-aware print shipped (no artifact exercises the new print — cosmetic). |
| F4 (E2E golden red) | **Discharged** | Golden re-baselined with the decision rule written into the test docstring; deviation history 0.053 → 0.293 → 0.107 posted to #509 (1.0734e-01 recomputed independently from the two goldens); E2E re-run twice independently, `max_abs_dev = 0.000e+00` same-machine. |
| F5 (reliable-mask blast radius) | **Discharged** | `/root/wt_i511/rfx/api/_spec.py:1296-1300` (contamination = entire `S[:, :, k]` slice; safe filter `np.all(reliable, axis=0)`), mirrored at `docs/public/guide/probes-sparams.mdx:219` and `docs/guides/sparameter_support_matrix.md` (in delta stat). |
| F6 (mixed-lane single-ratio residue) | **Discharged as annotate + fence + file** | Annotation at `_sparams.py:753` and `:2510-2567`; **#517 OPEN** with the correct scope (verified via gh). |
| F7 (notch-e4/cv07 rfx-leg regeneration) | **Remains open and UNFILED** | Was follow-up-only in the first review; no issue exists. |

**The d148ab4 caveat**: at a5acfa9 the thru-gate docstring still asserted the retired F2-defect calibration (0.0746 / 42.41 Ω + the coverage-fraction narrative) — the exact stale-calibration rot class this PR criticizes. The author self-caught it and pushed d148ab4 (docstring-only: 1 file, 25+/22−, **zero assert/bound lines changed** — verified at git level), which records the final 0.1160 / 0.9930 / 57.58 and documents the intermediate state AS the F2 defect. A verdict citing a5acfa9 alone would approve a tree containing the stale block; the PR head does not contain it.

## 2. Independently verified here vs taken on author/lens artifacts

**Re-measured in this re-review (all from inside `/root/wt_i511`, HEAD d148ab4):**
- **P5 invariance gate, full re-run to completion** (cwd verified via `/proc/739727/cwd -> /root/wt_i511`): `1 passed, 1015.47s`; `[MSL z0-len] mean|Z0| per length [60.23, 57.34, 57.58] Ω; spread = 4.96%` — **exact match to the author's `/tmp/f2_invariance.log`**. The last author-measured-only scorecard entry is now independently reproduced. Log: `/tmp/claude-0/-root-workspace-byungkwan-workspace-research-rfx/be57fe0b-0122-4a1f-a3be-708950b225c8/scratchpad/p5_invariance_a5acfa9.log`. Preflight context, verbatim, per repo rule:
  - `[PREFLIGHT] MSL port 'msl_1': only 3 substrate cell(s) in z (h_sub=254µm, dx=80µm). Yee staircase at dielectric interface is O(dx) — Z0 staircase error >5% expected. Refine to dx ≤ 64µm (4+ substrate cells) for <5% Z0 bias.`
  - `[PREFLIGHT] MSL port 'msl_1': h_sub/dx = 3.175 (fractional part 0.175) lands in the [0.10, 0.40] mixed-cell danger zone. ...`
  - `[PREFLIGHT] 'pec' z-extent 80µm = 1.0 cells — below 1 cell resolution. ...`
  - `S-matrix projected onto the passive set (singular values clipped to 1): 30 of 30 frequency bins were non-passive as extracted, worst sigma_max = 1.001 at 3.603 GHz.`
- Wiring/unit test file: **20 passed, 6.62s**.
- **Mixed-lane fast selection at HEAD: 56 passed, 2 skipped, 119.73s** (`pytest tests/ -k "mixed"`) — post-F2 smoke over the mixed lane's fast surface (new evidence; the body's 737-test sweep predates F2).
- Committed replay golden column power measured directly: **1.000010–1.001111** (PR body says "1.0000–1.0007" — superseded number, minor confirmed).
- Both V-span call sites, delta stat, d148ab4 docstring-only, cond_a f64 cast, F5 text in all three files, #517 open, PR head SHA.
- Docstring staleness at HEAD: invariance test lines 409-410 still claim "~10x the measured spread"; thru-gate lines 134-135 carry the correct 4.96% margin-less note.

**Taken on author artifacts / lens verification (not re-measured):** P1 probe 44.11/41.86 (byte-compared to the pre-F2 log by lens 3), P2 gate (`/tmp/f2_gate.log`, 1 passed, 428s), P6 NU pair (`/tmp/f2_nu.log` vs `/tmp/f2_nu_wt.log`), E2E re-runs (lenses 1+2, `max_abs_dev = 0.000e+00`), replay-golden bit-identity regeneration from stored accumulators (lens 2), wiring-test mutation testing (lens 1), NU span-equivalence arithmetic (lens 1), falsifier pre-declaration timestamps.

## 3. Findings, ranked (none blocking)

1. **[minor] Invariance test docstring is now false at HEAD**: `tests/test_msl_port_integration.py:409-410` still records "cross-length spread ~0.49% ... the 5% spread gate is ~10x the measured spread", while the measured post-F2 spread is 4.96% — the gate is 1.008x, margin-less. Same rot class the PR criticizes; d148ab4 fixed the sibling docstring but not this one. **Action: one-line amendment (can ride any follow-up commit; no bound change).**
2. **[minor] The margin-less P5 lock has no standalone tracker**: the body's "follow-up filed" is a comment on #511, which this PR closes ("Closes #511"). No open issue covers the L=6mm short-line fit-conditioning investigation (checked #509-#517). **Action: file an open issue before or at merge.**
3. **[minor] PR body stale remnants presented as current**: the Verification section's gate table (0.0746/42.41), the "737 passed" blast-radius sweep (measured at 98032e7, pre-F2 code), the replay colP "1.0000–1.0007", and P6's "bit-identical to clean main" (author's own log pair differs at one bin: |S11| 0.97580 vs 0.97581 at 6.8 GHz — gate-identical, not bit-identical). The post-F2 scorecard corrects the first two 30 lines later, but nothing flags them as superseded where they appear. **Action: one body edit pass.**
4. **[minor] F4 provenance pointer dangles**: the golden docstring says the capture script is "recorded in PR #516", but it exists only as volatile `/tmp/f2_e2e_capture.py` (0 hits in body/comments); the docstring also cites the superseded 2.93e-1 deviation as this PR's (shipped algorithm: 1.07e-1). **Action: paste the script into a PR comment (or commit under scripts/) and fix the number.**
5. **[minor] First-review follow-ups still unfiled**: F7 regen chain, the committed flux-oracle artifact (including the bisecting-mesh flux identity the body itself marks "not yet measured on this mesh"), the 7 PLAUSIBLE-UNVERIFIED overflow findings, raw_3probe metadata / S-vs-S^T test. Only #517 exists. **Action: file or explicitly decline each.**
6. **[note] Mixed-lane V-span anchor has no dedicated wiring test** — both new tests drive `compute_msl_s_matrix` only. Mitigated: the change is character-identical to the pure lane (verified at `:3471`) and my 56-test mixed fast sweep is green at HEAD; but no test would catch a future divergence of the two lanes.
7. **[note] cond_a NEP-50 fix has no regression test**: the pre-existing `test_cond_a_reports_a_degenerate_drive_system` runs under `enable_x64()` with a near-singular (γ=0.999), not exactly singular, system — it cannot trip the f32 underflow the fix addresses.
8. **[note] Both wiring fixtures rasterize 1-node traces**, so `trace_k_per_port[p][0]` vs `[1]` (trace-bottom vs trace-top anchor) is indistinguishable to them.

## 4. Remaining UNMEASURED

- Bisecting-mesh (dx=80) flux identity `Re(V·I*)/flux` — disclosed as unmeasured in the body itself.
- Full 737-test blast-radius selection post-F2 (covered piecewise: 20 wiring + 3 AD + 10 honesty + 56 mixed + 5 slow gates; not as one sweep).
- P6 full NU gate at d148ab4 exactly (author ran at the a5acfa9 code state; d148ab4 is docstring-only, so a re-run is arithmetically a no-op — accepted, not measured).
- The f32 cond_a underflow path (no test; fix verified by inspection and by lens reproduction of the underflow).
- The new sign-aware verdict print (the P1 log predates the edit; physics numbers unaffected).

## 5. Follow-up checklist delta vs first review

- **Closed since first review**: F1 (13:16 comment), F2, F3, F4, F5, F6 (a5acfa9 + d148ab4); #517 filed.
- **Carried over, still open**: F7 regen chain; flux-oracle artifact; overflow skeptic-check; raw_3probe metadata item — all unfiled.
- **New from this re-review**: invariance-docstring 10x fix (item 1); standalone tracker for the margin-less invariance lock (item 2); PR body edit pass (item 3); F4 provenance (item 4); optional cond_a regression test and mixed-lane wiring test (items 6-7).

**Memory audit (R1/R3)**: consistent with `feedback_summary_compression_audit` (artifacts quoted verbatim before closing — applied above), `feedback_gate_can_bind_artifact` (the author pre-declared falsifiers and disclosed the falsified P5 prediction rather than adjusting the gate), and the no-silent-gate-loosening rule (verified zero assert/bound changes across 98032e7..d148ab4 in the gate files). R2: no non-closing attempts here — this pass was measurement/verification only. Falsifier for this verdict: re-running `pytest tests/test_msl_port_integration.py -m "" -q` at d148ab4 must keep all four thru-gate bounds green with the docstring numbers 0.1160/0.9930/57.58.

---

*Author-pass disposition of this re-review's findings (post-report, commits/issues above this comment):*
- *minor 1 (invariance docstring "~10x") and minor 4 (golden docstring 2.93e-1 + volatile capture script) — fixed in `7f889d8`; the capture procedure is now committed as `scripts/capture_msl_e2e_golden.py` with the re-baseline-needs-a-written-reason contract.*
- *minor 2 (margin-less P5 lock tracker) — filed as **#518**.*
- *minor 5 + notes 6/7 (F7 regen chain; flux-oracle artifact incl. the unmeasured bisecting-mesh identity; S-vs-Sᵀ test; mixed-lane wiring test; cond_a singular test; raw_3probe metadata; overflow findings) — filed as **#519** (regen chains) and **#520** (verification-debt umbrella).*
- *minor 3 (PR-body stale remnants) — supersession flags edited into the body at the four named spots.*

PR #521, opened 2026-07-31
docs(msl): reliable-mask scope — one-sided detector, all 8 surfaces

bk-squared:
Post-merge sweep of `f95240f` (PR #516) found my own F5 fix there was **both incomplete and, where it landed, overstated**. Docs/docstring only, no behaviour change.

## Incomplete — 5 of 8 surfaces still taught the unsafe rule

The corrected wording reached 3 surfaces. These five still said "exclude the driven-port column `S[:, p, k]`", which is unsafe under the multi-drive solve because S is solved jointly across drives:

- `docs/public/api/results-observables.mdx:98-99`
- `docs/public/api/sources-ports.mdx:78`
- `docs/public/validation/reference-lane.mdx:29`
- `docs/guides/simulation_methodology.md:141-142`
- `docs/guides/sparameter_support_matrix.json:139` — the .md's own machine-readable companion, contradicting the corrected .md in the same directory

## Overstated — the part that matters more

My F5 text asserted that `reliable[p, k]` *"still tells you WHICH port's plane collapsed"* and that `np.all(reliable, axis=0)` is *"the only **safe** per-bin filter"*. Neither is supportable:

```python
# rfx/api/_sparams.py — what the mask is fed
v_port = np.stack([raw_v[p, p, 0, :] for p in range(n_ports)])   # own-drive only
i_port = np.stack([raw_i1[p, p, :]   for p in range(n_ports)])

# what the solve consumes
wave_a[driven][j] = 0.5 * (v0_j + z0_j * i_j)   # every (driven, port) pair
```

A collapse at a **passive** port's plane during another port's drive never reaches the mask, yet under `S = B·A⁻¹` it corrupts the whole slice. So the `np.all` filter can pass a poisoned bin, and the index `p` does not reliably attribute the collapse. I replaced a wrong claim (column containment) with a differently wrong one (attribution + filter sufficiency).

## What all 8 surfaces now say

Two-sided, consistently:

- a **False** entry condemns the entire frequency slice `S[:, :, k]`, not one column;
- a **True** entry is neither an accuracy guarantee **nor full coverage** — the mask watches own-drive records only;
- `np.all(reliable, axis=0)` is documented as **necessary but not sufficient** — the strongest screen this mask supports, not a clean-bin guarantee.

The underlying coverage gap (mask inputs vs solve consumption) is a design decision — widen the mask's inputs, or keep it own-drive and scope the docs — and is filed separately. This PR stops the docs promising what the mask cannot deliver in the meantime.

## Verification

- `grep` for column-only claims across `docs/`: **none remain**
- support-matrix + AD-surface contract tests: **24 passed**
- `scripts/check_api_reference.py`: OK · `ruff check rfx/ tests/`: clean
- JSON parses

`R3: memory=feedback_gate_can_bind_artifact (a True that does not measure what is consumed) | R2-attempts=0 | falsifier=grep returns no column-only claim; contract gates green`

### DIFF
--- a/docs/guides/simulation_methodology.md
+++ b/docs/guides/simulation_methodology.md
@@ -139,8 +139,11 @@ or to CW drive.
 - Treat Harminv frequency and Q as estimates from the supplied time window;
   compare stability across windows and meshes.
 - For MSL S-parameters, `reliable[p, k] == False` means the voltage/current
-  split for driven port `p` is too weak at bin `k`; exclude column `S[:, p, k]`
-  from RF interpretation. `True` is not a general accuracy guarantee.
+  split for port `p` is too weak at bin `k`; exclude the whole slice
+  `S[:, :, k]` from RF interpretation, since S is solved jointly across drives
+  (#507). `True` is neither a general accuracy guarantee nor full coverage — the
+  mask watches own-drive records only, so a passive-port plane collapse at that
+  bin is invisible to it.
 - On nonuniform MSL models, use the supported normalization documented by the
   calculator. Do not assume a uniform-grid normalization option is accepted.
 - For RCS, subtract an incident-only reference field before forming scattered
--- a/docs/guides/sparameter_support_matrix.json
+++ b/docs/guides/sparameter_support_matrix.json
@@ -136,7 +136,7 @@
       "known_limits": [
         "nonuniform mode=laplace or mode=uniform has no external comparison",
         "Result.s_params is not used; call compute_msl_s_matrix(...)",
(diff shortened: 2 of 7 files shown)


bk-squared:
Superseded by #526, which includes this commit plus the two code fixes that make its wording true: #522 (the mask now covers every record the solve consumes, so attribution and the `np.all` screen are actually supportable) and #523 (the assembly marker). Closing to keep one reviewable flow rather than two PRs where the second rewrites the first.

Issue #525, opened 2026-07-31
flux oracle for the MSL modal V.I is not yet correctly constructed (factor-2 convention + reactive-residue fixture)


bk-squared:
The modal V·I pair does not measure the guided power on a **bisecting** mesh — the flux identity that certifies the corrected voltage span holds only where `h_sub/dx` is integral. Measured today on merged main (`f95240f`).

## What was measured

The extractor-independent oracle for the modal voltage is

```
Re(V · conj(I)) / flux_spectrum  ==  1
```

Poynting flux is accumulated by a different kernel, so it shares no code with the V·I path. PR #516 measured it on the **aligned** mesh (dx = 84.67 µm, `h_sub/dx` = 3.0000) and got 1.006–1.009 with the corrected span. That mesh is the easy case: the rounding proxy and the rasterized trace node coincide there, so PR #516's F2 anchoring change is a *no-op* on it (measured bit-identical). The mesh class where F2 actually bit had never had this identity measured — the PR body says so verbatim ("expected ~1 by construction; not yet measured on this mesh").

Now measured. Single driven MSL port, far end CPML-absorbed, 5 planes × 6 frequencies, dx = 80 µm (`h_sub/dx` = 3.1750, trace rasterized at node 4 vs rounding proxy 3), `num_periods=30`:

| span | ratio range |
|---|---|
| **shipped** (trace-anchored, edges 0..3) | **0.5382 – 0.6935** |
| proxy / pre-F2 (edges 0..2) — control, same recorded fields | 0.3968 – 0.5117 |

Both spans, all planes, all frequencies. Monotone with frequency (0.69 at 2.0 GHz → 0.54 at 4.5 GHz) and mildly plane-dependent.

Preflight on this fixture, verbatim, quoted per repo rule:

```
[PREFLIGHT] MSL port 'msl_0': only 3 substrate cell(s) in z (h_sub=254µm, dx=80µm).
  Yee staircase at dielectric interface is O(dx) — Z0 staircase error >5% expected.
  Refine to dx ≤ 64µm (4+ substrate cells) for <5% Z0 bias.
[PREFLIGHT] MSL port 'msl_0': h_sub/dx = 3.175 (fractional part 0.175) lands in the
  [0.10, 0.40] mixed-cell danger zone. The substrate-air interface bisects the same
  Yee cell that holds the trace ... To snap onto a safe alignment, set dx = 63.5µm
  (= h_sub/4) or 84.7µm (= h_sub/3).
[...]

Issue #520, opened 2026-07-30
verification debt from PR #516's two review passes — committed flux-oracle artifact, overflow findings, metadata, discriminating tests


bk-squared:
Umbrella tracker for the verification items both review passes of PR #516 left open. Each is small; none blocks the merge; all are the difference between 'measured once in a session' and 'checkable by the next person'.

- [ ] **Committed flux-oracle artifact** (research/CLAUDE.md requires a named artifact for claims-bearing numbers): a `scripts/diagnostics/` script measuring `Re(V·conj(I))/flux_spectrum` at several planes, run on BOTH mesh classes. The aligned-mesh numbers (0.881-0.885 old / 1.006-1.009 new) live only in docstrings + session logs; the **bisecting-mesh identity has never been measured at all** (PR body says 'expected ~1 by construction; not yet measured'). This is also the falsifier future extractor changes should re-run.
- [ ] **S-vs-Sᵀ discriminating test** for `msl_solve_s_from_waves`: every committed fixture is symmetric, so a transposed solve would pass everything today (first review, minors). One non-symmetric planted-S case closes it (the re-review verified the orientation numerically; nothing pins it).
- [ ] **Mixed-lane V-span wiring test**: both new wiring tests drive `compute_msl_s_matrix` only; the mixed lane's identical change (`_sparams.py:3471`) has no test that would catch a future divergence of the two lanes.
- [ ] **cond_a exactly-singular regression test**: the NEP-50 f64-cast fix (a5acfa9) is exercised only by a near-singular (γ=0.999) case under x64; an exactly-singular f32 case would have divided by zero pre-fix and nothing pins the fix.
- [ ] **raw_3probe dump metadata**: `metadata_json.deembedding` still attributes production S to the N-probe fit and `schema_version` is unbumped despite the semantic change to the assembly (first review, minors).
- [ ] **Warning strings** (~3) still describing the pre-#507 single-ratio design (first review, minors; one was fixed in a5acfa9's honesty-guard comment, the strings themselves were not audited).
- [ ] **Skeptic-check the 7 PLAUSIBLE-UNVERIFIED overflow findings** from the first review (support-matrix stale numeric claims incl. the '48.6-ohm notch Z0' and 'AD validated 0.000110'; broad-E5 envelope fencing resting on now-retracted evidence; `scripts/diagnostics/replay_msl_3probe_dump.py` as a dead evidence lane; mixed-lane slow battery absent). Each needs confirm-or-dismiss, not silent carry.

[...]

PR #529, opened 2026-08-01
fix(test): rebuild the MSL AD-vs-FD comparator in float64 and gate its resolving power (closes #527)

### LINKED ISSUE #527: MSL AD-vs-FD gate: the f32 single-h FD comparator cannot resolve the 0.10 bound (not an AD defect)
The MSL AD gradient no longer matches finite differences: **`rel_err = 0.8519`** against a 0.10 gate. Measured on merged main (`f95240f`) by the maintained GPU harness.

This is the re-measurement #515 called for after PR #516 replaced the S-matrix assembly. The pre-#516 record was `rel_err = 0.00011`.

## The run

VESSL **369367250602**, `scripts/vessl_gpu_suite.yaml`, `remilab-c0` / gpu-rtx4090, main@`f95240f`, 2026-07-31. Log backed up to `/tmp/gpu_369367250602.log`; full pytest log in the mounted artifact `.omx/gpu-suite/20260731T053033Z-f95240f/gpu_suite.log`.

```
1 failed, 193 passed, 70 skipped, 2912 deselected, 2 xfailed  in 4061.29s (1:07:41)
FAILED tests/test_msl_ad_fd_converged.py::test_msl_ad_fd_converged_tight
```

**One failure in the whole GPU suite**, and it is the gradient gate.

## The numbers

```
[MSL-FD-TIGHT] n_steps=26226, checkpoint_segments=141 (~sqrt=161.9)
[MSL-FD-TIGHT] forward |S| range: [0.0080, 1.0000]
[MSL-FD-TIGHT] loss = 1.600595e+01
[MSL-FD-TIGHT] g_ad = -4.236159e-03  (AD wall-time: 68.5s)
[MSL-FD-TIGHT] g_fd = -2.861023e-02  (FD wall-time: 34.9s, h=0.001)
[MSL-FD-TIGHT] rel_err = 0.8519  (threshold: 0.10)
[MSL-FD-TIGHT] sign agreement: g_ad=-4.2362e-03 g_fd=-2.8610e-02
[MSL-FD-TIGHT] num_periods=20, n_freqs=8
```

| | AD | FD | rel_err |
|---|---|---|---|
| pre-#516 (f64 referee, #486) | −2.11425e-01 | −2.11449e-01 | **0.00011** |
| post-#516 (this run) | **−4.2362e-03** | **−2.8610e-02** | **0.8519** |

Signs still agree, and the forward pass is healthy (`|S|` in [0.008, 1.000], well inside the guard's [0, 1.2]).

**The shape of the change is the informative part.** Both gradients dropped, but not together: FD fell ~7x (0.211 → 0.0286) while AD fell ~50x (0.211 → 0.00424). A changed extractor legitimately changes the loss and therefore both gradients — it does **not** explain them ceasing to agree with each other. The disagreement is the finding.

## What is NOT the explanation

- **Not a broken forward pass** — `|S|` range healthy, loss finite, both gradients nonzero with matching sign.
- **Not the smoke-test vacuity of #515** — that was a synthetic-fixture artifact; this is the real end-to-end gate on a physical fixture.
- **Not gate drift** — the threshold is untouched at 0.10, and the test's own docstring says: *"If rel_err stays >10% this test will fail (deliberately — do NOT loosen the gate to force a pass; report as a gradient accuracy finding instead)."* Following that.

## The lead I would start from

**This fixture is dx = 80 µm, `h_sub/dx` = 3.175 — the bisecting mesh class of #525.** Its preflight, verbatim from the run:

```
[PREFLIGHT] MSL port 'msl_0': h_sub/dx = 3.175 (fractional part 0.175) lands in the
  [0.10, 0.40] mixed-cell danger zone. The substrate-air interface bisects the same
  Yee cell that holds the trace ... To snap onto a safe alignment, set dx = 63.5µm
  (= h_sub/4) or 84.7µm (= h_sub/3).
[PREFLIGHT] MSL port 'msl_0': only 3 substrate cell(s) in z ... Z0 staircase error >5%
```

On that same mesh class, #525 measured the modal V·I pair recovering only **0.54–0.69** of the Poynting flux (vs 1.006–1.009 on an aligned mesh). Two independent failures on the same mesh class is the strongest single clue available, and #525's probe (span sweep × alignment class) is running now.

That said — **two symptoms on one mesh is a correlation, not a mechanism.** A V·I pair that mis-measures power by a fixed factor would move AD and FD *together*, not apart, so #525 cannot be the whole story here. Something in the AD path specifically is losing sensitivity.

## Candidate mechanisms, none tested

1. The multi-drive solve puts `jnp.linalg.solve` on the tape. Unit-level AD-vs-FD through `msl_solve_s_from_waves` was verified at 1.165e-9 (PR #516 review), so the primitive is sound — but its **conditioning** interacts with reverse-mode differently than the old per-column ratio did.
2. Something upstream of the solve became effectively non-differentiable or gradient-attenuating for this parameter (the fixture differentiates an `eps_r_sub` override; #483's lesson was that a launch fixture can silently freeze what the tape sees).
3. The checkpointed segmented scan (`checkpoint_segments=141`) interacting with the new assembly.
4. The FD reference is itself noisy — the test's own docstring records a CPU h-sweep showing 27% non-monotone FD scatter. At `g_fd` = 0.0286 (7x smaller than before) that noise floor is proportionally much larger, so part of the 0.85 could be reference noise. **An h-sweep on the current code is the cheapest way to bound this before chasing AD.**

## Next

Start with (4) — it is cheap and bounds how much of the 0.85 is real. Then (2), because a frozen/attenuated parameter path is the failure mode this repo has hit before (#483) and it predicts exactly this signature: AD far smaller than FD.

Blocks the AD-related claims in `docs/guides/sparameter_support_matrix.md` (the "converged f64 referee 0.000110" bullet, flagged in #520) until resolved. Related: #515, #525, #520.

bk-squared:
## The comparator, not the gradient

`test_msl_ad_fd_converged_tight` compared AD against a central difference whose
two loss evaluations ran in **float32**. Near `loss = 16.00599` one float32 ULP
is `1.9073e-06`, and at the gate's `h = 1e-3` the true difference `2h|g|` is
`8.47e-06` — **4.4 ULP**. The subtraction was returning a count of quantisation
steps, not a derivative.

Measured on the gate's own fixture and h:

| comparator | `g_fd` | `rel_err` | resolving power |
|---|---|---|---|
| float32 (shipped) | −2.861023e-02 | **0.8519** | 4.4 ULP |
| float64 (this PR) | −4.1005380e-03 | **0.0331** | 2.31e+09 ULP |

`g_ad = −4.236159e-03` (gpu-rtx4090) / `−4.242490e-03` (CPU). **The AD is
correct.** This was a comparator defect for the gate's whole lifetime.

The proof is arithmetic, not statistical: a 7-point h-sweep (VESSL
369367250742) returned FD values that were **exact integer multiples of
`ULP/(2h)` at every h** — 14, 1, −30, −7, −31, −52, −76 — and at `h = 5e-4` the
entire measured "gradient" was one single ULP. The sign flip that looked like
physics is which side of a representable value each evaluation landed on. No h
repairs it: 1% resolution needs `h > 0.0225`, where `h²` truncation takes over.

### Why it surfaced now

`sum|S|²` is pinned near the passivity value `2·n_freqs = 16`; the measured loss
is `16.00599`, so **0.037% of the objective is signal** riding on a structural
constant. PR #516 removed extractor error, so that residue — and its derivative
— shrank 50x, from `−2.1143e-01` to `−4.2425e-03`, while the loss magnitude and
therefore the ULP stayed put. The gradient fell off the resolvable end of
float32.

Corroborated by a number recorded **before** #516: the #477 work logged a "27%
NON-monotone FD scatter" on this fixture. The same sweep now reads 2017%, and
`27% × 50 ≈ 1350%` is the same order. Read the other way, the gradient collapse
is evidence that #516 worked — the old gate was substantially differentiating an
extractor artefact.

## Changes

1. **The FD reference runs in float64** under a scoped `enable_x64()`. Never at
   module level: process-global, flips at collection, reds every pytest-split
   shard. `probes.py` already keys accumulator dtype off `jax.config.x64_enabled`,
   so x64 reaches `S` (verified `complex128` / `float64`; scoping verified
   non-leaking across tests).
2. **A comparator-validity assert, placed BEFORE the accuracy gate.** The FD
   span must exceed `_MIN_FD_ULP_SPAN = 1e4` ULP. This is the check whose
   absence let this issue be filed against the AD path — the honest reading of
   `rel_err = 0.85` on a 4.4-ULP reference is *"the instrument is broken"*. Its
   failure message says exactly that and tells the reader not to touch the
   extractor or the tape.
3. **`scripts/msl_ad_fd_f64_referee.py`** — the referee, committed. The f64
   measurement behind the docstring's `rel_err = 0.00011` claim existed only as
   a sentence; no script in the repo reproduced it. Third instance of that class
   in this area (#520, #525).
4. The docstring sentence *"the 0.10 threshold now carries real margin"* is
   replaced by the measurement that falsified it.

**The 0.10 threshold is unchanged. Nothing here loosens a gate.**

## Cost, measured on the lane

A float64 reference sounds expensive; on this lane it is not. f32 AD 59.1s, f64
FD 41.7s and 37.7s per h (two forwards each) — ~19s per f64 forward against ~15s
for f32, whole gate ~140s. The FDTD is memory-bandwidth-bound, so the RTX4090's
1/64 float64 throughput barely shows. On CPU the same forward is ~500s, which is
why this was decided from a run on the lane the gate actually uses (VESSL
369367250775) rather than a local timing.

## Falsifier

`test_comparator_floor_rejects_the_f32_reference_that_caused_527` (fast, no
simulation) pins the new floor from **both** sides: it must reject the exact
float32 configuration that caused this issue (4.4 ULP) and accept the float64
one (2.4e9 ULP), with a margin assert so a future edit that halves the headroom
is visible. If someone later lowers the floor to quiet a red, that test tells
them what they are re-admitting.

## Preflight (quoted, per repo rule)

The fixture emits, unchanged by this PR and identical on both platforms: `'pec'
z-extent 80µm = 1.0 cells — below 1 cell resolution`; `pec_faces={z_lo} creates
an INFINITE PEC boundary AND the geometry contains finite PEC objects`; `all
dielectric(s) ['ro4350b'] are perfectly lossless in an open (CPML) domain`; and
per port `only 3 substrate cell(s) in z … Z0 staircase error >5% expected` and
`h_sub/dx = 3.175 … lands in the [0.10, 0.40] mixed-cell danger zone`. The last
one matters for interpreting the fixture but not for this change: the gate's
existing `s_max <= 1.2` forward guard covers it, and nothing here alters the
geometry.

Closes #527.

`R3: memory=feedback_root_cause_before_gate_change ("compute the discretization/Yee bound … FIRST" — the written root cause is the float32 ULP quantisation, measured before any gate edit) + feedback_gate_can_bind_artifact (an ADDED gate needs a falsifier) | R2-attempts=0 (comparator repair after a closed root cause, not a mechanism retry) | falsifier=test_comparator_floor_rejects_the_f32_reference_that_caused_527 pins the floor from both sides; note the usual domain-size invariance witness does not apply here — the added gate is arithmetic on the comparator's resolving power, not a physics claim`

### DIFF
--- a/scripts/msl_ad_fd_f64_referee.py
+++ b/scripts/msl_ad_fd_f64_referee.py
@@ -16,20 +16,30 @@ A float32 value near ``loss = 16.005951`` can only move in steps of one ULP,
 ``2h|g| = 8.4723e-06`` — **4.4 ULP wide**. The subtraction therefore returns an
 integer number of quantisation steps, not a derivative.
 
-Measured on the gate's own fixture (VESSL 369367250742, gpu-rtx4090), every
-value in an h-sweep came back an EXACT integer multiple of ``ULP/(2h)``:
+Measured on the gate's own fixture (VESSL 369367250742, gpu-rtx4090), the span
+stayed between 1 and 76 ULP across the whole sweep, while the FD value scattered
+2017% and changed sign:
 
-    h        g_fd            ULP/(2h)      g_fd / quantum
-    3.0e-04  +4.450480e-02   3.1789e-03    14.0000
-    5.0e-04  +1.907349e-03   1.9073e-03     1.0000   <- one single ULP
-    1.0e-03  -2.861023e-02   9.5367e-04   -30.0000   <- the gate's h
-    2.0e-03  -3.337860e-03   4.7684e-04    -7.0000
-    5.0e-03  -5.912781e-03   1.9073e-04   -31.0000
-    1.0e-02  -4.959106e-03   9.5367e-05   -52.0000
-    2.0e-02  -3.623962e-03   4.7684e-05   -76.0000
+    h        g_fd            span (ULP)
+    3.0e-04  +4.450480e-02       14
+    5.0e-04  +1.907349e-03        1     <- the entire measurement is ONE ULP
+    1.0e-03  -2.861023e-02       30     <- the gate's h
+    2.0e-03  -3.337860e-03        7
+    5.0e-03  -5.912781e-03       31
+    1.0e-02  -4.959106e-03       52
+    2.0e-02  -3.623962e-03       76
 
-Seven for seven. The sign flip that looked like a physics puzzle is just which
-side of a representable value each evaluation landed on.
+The sign flip that looked like a physics puzzle is just which side of a
+representable value each evaluation landed on.
+
+CAUTION for anyone reusing this reasoning: those spans are exact integers, and
+an earlier draft of this docstring offered that integrality as the proof. It is
+not evidence of anything — ANY two float32 values in the same binade differ by
+an integer number of ULPs (200k random pairs, including deliberately
+well-resolved ones 47186 ULP apart, all reproduce it). Integrality shows only
+that the values were float32. The load-bearing fact is the MAGNITUDE of the
+span. Do not carry the integer-multiple test forward as a diagnostic; it fires
+on healthy comparators too.
 
 No h repairs this: 1% quantisation resolution needs ``2h|g| > 100 ULP``, i.e.
 ``h > 0.0225`` — a 2.3% perturbation of ``alpha``, where ``h^2`` truncation
--- a/scripts/msl_ad_fd_f64_referee.py
+++ b/scripts/msl_ad_fd_f64_referee.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python3
+"""Float64 finite-difference referee for the MSL AD gate (issue #527).
+
+WHY THIS EXISTS
+---------------
+``test_msl_ad_fd_converged_tight`` compares reverse-mode AD against a central
+finite difference whose two loss evaluations are computed in **float32**::
+
+    alpha0  = jnp.float32(1.0)
+    f_plus  = float(objective(jnp.float32(float(alpha0) + _FD_H)))
+    f_minus = float(objective(jnp.float32(float(alpha0) - _FD_H)))
+    g_fd    = (f_plus - f_minus) / (2.0 * _FD_H)
+
+A float32 value near ``loss = 16.005951`` can only move in steps of one ULP,
+``1.907349e-06``. At the gate's ``h = 1e-3`` the true difference is
+``2h|g| = 8.4723e-06`` — **4.4 ULP wide**. The subtraction therefore returns an
+integer number of quantisation steps, not a derivative.
+
+Measured on the gate's own fixture (VESSL 369367250742, gpu-rtx4090), every
+value in an h-sweep came back an EXACT integer multiple of ``ULP/(2h)``:
+
+    h        g_fd            ULP/(2h)      g_fd / quantum
+    3.0e-04  +4.450480e-02   3.1789e-03    14.0000
+    5.0e-04  +1.907349e-03   1.9073e-03     1.0000   <- one single ULP
+    1.0e-03  -2.861023e-02   9.5367e-04   -30.0000   <- the gate's h
+    2.0e-03  -3.337860e-03   4.7684e-04    -7.0000
+    5.0e-03  -5.912781e-03   1.9073e-04   -31.0000
+    1.0e-02  -4.959106e-03   9.5367e-05   -52.0000
+    2.0e-02  -3.623962e-03   4.7684e-05   -76.0000
+
+Seven for seven. The sign flip that looked like a physics puzzle is just which
+side of a representable value each evaluation landed on.
+
+No h repairs this: 1% quantisation resolution needs ``2h|g| > 100 ULP``, i.e.
+``h > 0.0225`` — a 2.3% perturbation of ``alpha``, where ``h^2`` truncation
+dominates instead. The usable window has closed, so Richardson extrapolation
+over h would extrapolate over quantisation noise.
+
+This did not used to bite. Before PR #516 the gradient was ``2.1143e-01``; the
+extractor fix moved ``|S11|`` from 0.2233 to ~0.05, shrinking it 50x to
+``4.2362e-03`` while the loss magnitude — and therefore the ULP — stayed put.
+The signal fell off the resolvable end of float32:
+
+    2h|g| at the gate's h:   pre-#516 ~222 ULP    post-#516 4.4 ULP
+
+Independently corroborated by a number recorded before that change: the #477
+work logged a "27% NON-monotone FD scatter" on this fixture; the same sweep now
+reads 2017%, and 27% x 50 ~ 1350% is the same order.
+
+WHAT THIS SCRIPT DOES
+---------------------
+Evaluates the loss under ``jax.experimental.enable_x64()`` (a scoped context —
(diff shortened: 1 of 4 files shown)


bk-squared:
## The gate runs green on the lane it is marked for

This was the gap I flagged against my own PR: `gpu`+`slow` means CI never executes this test, so it shipped without end-to-end evidence. VESSL **369367250794** runs the pytest file itself — not the referee script, which exercises the same physics but none of the test's asserts or their ordering — on `remilab-c0`/`gpu-rtx4090` at `9bbacbd`.

```
[MSL-FD-TIGHT] loss = 1.600595e+01
[MSL-FD-TIGHT] g_ad = -4.236159e-03  (AD wall-time: 55.8s)
[MSL-FD-TIGHT] g_fd = -4.100538e-03  (FD wall-time: 41.8s, h=0.001)
[MSL-FD-TIGHT] FD reference resolving power: 2.31e+09 ULP of the loss (floor 1e+04)
[MSL-FD-TIGHT] rel_err = 0.0331 (threshold: 0.10, enforced below)
[MSL-FD-TIGHT] sign agreement: g_ad=-4.2362e-03 g_fd=-4.1005e-03
[MSL-FD-TIGHT] total wall-time: 117.5s
[MSL-FD-TIGHT] PASS
3 passed in 119.51s (0:01:59)
```

Three things this establishes that nothing before it did:

1. **x64 actually engaged on the GPU.** The dtype assert sits inside the measured path and did not fire. Before that assert existed, a green run proved nothing — the resolving-power number read ~2.4e9 whether the loss was float32 or float64.
2. **The comparator floor passes on real data**, 2.31e+09 against a 1e+04 floor, with the check running *before* the accuracy gate as designed.
3. **The print/enforce mismatch is gone** — `threshold: 0.10, enforced below`, where it used to print 0.05 while enforcing 0.10.

Cost: 117.5s total, against ~89s for the f32 version it replaces.

A CPU run of the same file is in flight as a second platform (`g_ad = -4.242490e-03`, matching GPU to 0.15%); it is supporting evidence, not the authority — the lane run is.

**Preflight, quoted verbatim** (unchanged by this PR, identical on both platforms): `'pec' z-extent 80µm = 1.0 cells — below 1 cell resolution`; `pec_faces={z_lo} creates an INFINITE PEC boundary AND the geometry contains finite PEC objects`; `all dielectric(s) ['ro4350b'] are perfectly lossless in an open (CPML) domain`; and per port `only 3 substrate cell(s) in z (h_sub=254µm, dx=80µm). Yee staircase at dielectric interface is O(dx) — Z0 staircase error >5% expected` and `h_sub/dx = 3.175 (fractional part 0.175) lands in the [0.10, 0.40] mixed-cell danger zone`.

Logs archived: `docs/research_notes/vessl_logs/20260801_369367250794_msl_ad_gate_529.log`, alongside the referee (369367250775), the h-sweep (369367250742), and the local CPU referee.

PR #549, opened 2026-08-04
PR #516 verification debt: the bisecting-mesh V·I/flux identity measured for the first time (HELD) + three discriminating tests (#520)

bk-squared:
## What / why

Four of #520's legs, each with a mutation twin proving discriminating power. Leg 5 (stale warning strings, raw_3probe metadata, the 7 PLAUSIBLE overflow findings) is honestly UNSTARTED — it needs its own research pass.

## Leg 1 — the flux oracle (the standing falsifier #520 asked for)

`scripts/diagnostics/msl_vi_flux_oracle.py` + committed JSON. Built to #525's own follow-up bar: the convention is stated AND runtime-asserted (no asymmetric ½ factor, via a real anti-regression guard — see below), an admissibility witness (reactive fraction + plane-to-plane spread from the flux monitor's own accumulators) gates each cell before the ratio is trusted, the far port is genuinely terminated (excite=False → resistive Z0, not bare absorption), and BOTH mesh classes run.

**First-COMMITTED bisecting-mesh measurement (#525 measured it twice in an uncommitted harness): the identity HELD on both classes, 30/30 admissible cells** — aligned h_sub/3: Re(V·I*)/flux = 1.0083–1.0090; bisecting 80 µm (the gate mesh): 1.0105–1.0118.

This does NOT reproduce #525's as-posted 0.54–0.69 bisecting read, and the reconciliation is now explicit in the script's append-only correction note: #525's own comment 2 supplies the convention-corrected figure (0.54–0.69, un-halved, → **1.20–1.34**), and review's independent measurement on the same unterminated fixture class, reconstructed (#525's own harness was never committed) measured **1.0898–1.3867**, matching #525's corrected figure to ~1%. The provenance of the reactive-witness evidence was also fixed: the 472%-spread/−122.9 dB figures are #525's own **ALIGNED**-mesh measurement, not bisecting (#525's own text says so); the witness leg that actually refuses that unterminated fixture is **reactive fraction** (0.963–0.995), not spread (0.011–0.067, which passes this script's 0.5 threshold). Both corrections are documented in the script's docstring.

New this round: the aligned-mesh number (1.0083–1.0090) resolves a question #525 comment 2 left open — whether PR #516's original 1.006–1.009 paired the convention correctly. It does: this script's independently convention-audited aligned measurement lands almost exactly on PR #516's own number, which stands validated, not coincidentally close.

Also fixed on review: the script's own "convention regression guard" was tautological (`p_vi/flux == (0.5·p_vi)/(0.5·flux)` — true for any fixture, the same defect class PR #531's review blocked in `scripts/msl_flux_ratio_dof.py`). Removed; replaced with `_oracle_ratio()` (the single choke-point implementation) plus a guard that plants TWO independent mutations, both routed through the real function (not a hand-inlined duplicate): a ½ on the numerator (`v`) side only — #525's own bug, ratio moves 1.0 → 0.5 — and a ½ on the denominator (`flux`) side only, the mirror-image bug an earlier version of the guard missed, ratio moves 1.0 → 2.0 (the shape of #525's own "convention-corrected 2.01–2.02" misdiagnosis: a script-side factor bug surfacing as a false physics over-report rather than a caught regression). The committed JSON now also carries a free mutation twin — the retired `round(h_sub/dx)` proxy span re-applied to the already-recorded fields (zero new FDTD): on the bisecting mesh it reads 0.7443–0.7452, hard outside [0.85, 1.15], proving the oracle can tell the retired span from the corrected one — and `|V|`/`|I|` per cell, since `scripts/msl_flux_ratio_dof.py` shows this ratio (and, structurally, both admissibility witnesses, which read only the flux monitor's own fields) is exactly invariant under the power-preserving rescale `(V,I) → (aV, I/a)` — a blind direction now stated explicitly rather than implied closed.

## Legs 2–4 — discriminating tests

- **S-vs-Sᵀ** (`test_msl_modal_voltage_and_wave_solve.py`): a deliberately NON-reciprocal planted-S fixture (any physical MSL S is symmetric, so only such a fixture can expose a transpose bug); the mutation twin shows the historical blind spot — the dropped-transpose bug is caught on the new fixture and invisible on every committed symmetric one.
- **Mixed-lane V-span wiring** (`test_mixed_port_sparam.py`): per-z-node Ez marker through the lane's own scan hook, raw V read back via return_diagnostics; reverting the call site to the retired `round(h_sub/dx)` proxy fails loudly (verified, then reverted clean).
- **cond_a exactly-singular**: a zero drive column reaches bit-exact 0.0 in float32 SVD (two near-identical columns do NOT — the first attempt was caught by its own mutation twin and replaced); the pre-fix uncast arithmetic reproduces `inf`, the shipped floor masks correctly.

## Verification

Tests pass across the two touched test files (own process), ruff clean, no gates/tolerances touched, no module-level x64. All commits carry R3 lines; the oracle script's original pre-declaration is append-only disciplined for every DECIDABLE element (band, mesh classes, fixture, thresholds) — this round's fixes are a dated append plus in-place corrections to non-falsifiable explanatory prose only, documented in the script itself.

R3: memory=#525's closure record (consistent — the oracle gates the measurable DOF with an admissibility witness; the historical non-reproduction is now explained AND reconciled to ~1% against #525's own convention-corrected number) + `scripts/msl_flux_ratio_dof.py` (DOF-blindness framing) | R2-attempts=1 oracle campaign, pre-declared, plus two review-response passes (the JSON-regenerating pass RE-EXECUTED the full 2-mesh campaign — not skipped — with `p_vi_real`/`flux_real`/`ratio`/`reactive_fraction` bit-identical to the original committed run: reviewer-verified, and independently re-confirmed field-by-field from the two committed JSON commits in this pass, so the new fields are pure re-analysis of the same physics) | falsifier=every leg's mutation twin, committed; the proxy-span twin's numbers are in the committed JSON; the guard's own two mutations are verified to move the ratio to 0.5 and 2.0 respectively

### DIFF
--- a/rfx/api/_sparams.py
+++ b/rfx/api/_sparams.py
@@ -137,8 +137,9 @@ def msl_modal_voltage(ez_plane, *, j_centre: int, k_lo: int, k_hi: int,
     unterminated single-port reading of this same identity on the
     bisecting mesh (issue #525) read low (0.54-0.69) for two compounding,
     non-extractor reasons — a convention slip (see #525's own correction)
-    and a reactive, non-travelling fixture (reactive fraction 0.963-0.995) —
-    neither of which is present in the committed measurement.
+    and a reactive, non-travelling fixture (reactive fraction 0.963-0.995,
+    PR #549 review reconstruction) — neither of which is present in the
+    committed measurement.
     """
     if k_hi <= k_lo:
         raise ValueError(
--- a/rfx/api/_sparams.py
+++ b/rfx/api/_sparams.py
@@ -128,6 +128,17 @@ def msl_modal_voltage(ez_plane, *, j_centre: int, k_lo: int, k_hi: int,
     frequencies — **on the aligned dx = 84.67 µm mesh**; that identity is
     the falsifier for THIS span (ground→trace underside) and holds only
     when the top anchor is the true trace node.
+
+    The BISECTING mesh (dx = 80 µm, the mesh class this span anchoring
+    actually changes behaviour on) has since been measured too, on a
+    properly terminated two-port fixture (issue #520 leg 1;
+    ``scripts/diagnostics/msl_vi_flux_oracle.py`` + its committed JSON):
+    HELD, ratio 1.0105-1.0118, 30/30 admissible cells. An earlier
+    unterminated single-port reading of this same identity on the
+    bisecting mesh (issue #525) read low (0.54-0.69) for two compounding,
+    non-extractor reasons — a convention slip (see #525's own correction)
+    and a reactive, non-travelling fixture (reactive fraction 0.963-0.995) —
+    neither of which is present in the committed measurement.
     """
     if k_hi <= k_lo:
         raise ValueError(
--- a/scripts/diagnostics/msl_vi_flux_oracle.py
+++ b/scripts/diagnostics/msl_vi_flux_oracle.py
@@ -14,8 +14,16 @@ V and I are built from the trace-anchored span (``msl_modal_voltage`` /
 Poynting flux is accumulated by a DIFFERENT kernel entirely
 (``rfx/probes/probes.py:flux_spectrum``, fed by ``rfx/simulation.py``'s own
 H-field spatial-averaging + half-step phase correction — see
-``rfx/simulation.py:1470-1523``). The two paths share no code, so agreement
-is a genuine, independent check, not a tautology.
+``rfx/simulation.py:1470-1523``). Agreement is not tautological: R is not
+computed from its own numerator or denominator restated (unlike the
+now-deleted convention check this PR replaced, see CONVENTION below). But
+code-path separation is not the relevant strength claim, and this script
+does not claim "genuine, independent" as if that closed the question --
+``scripts/msl_flux_ratio_dof.py`` (the #525/#531 closure artifact) shows R
+constrains exactly ONE real degree of freedom of (V, I) -- the product
+``Re(V·conj(I))`` -- while the wave split the extractor actually uses
+(``a, b = 0.5·(V ± Z0_ref·I)``) and the separately reported analytic Z0
+both depend on V and I SEPARATELY. See DOF BLINDNESS below.
 
 PR #516 measured this identity ONLY on the aligned mesh (dx = h_sub/3 =
 84.667 µm, h_sub/dx = 3.0000) and got 1.006-1.009 — the EASY case, where the
@@ -45,9 +53,10 @@ This script is the committed re-measurement #525's own follow-up comment
      test_msl_thru_line_passive_gate``, drives only the near port, and
      leaves the far port passively matched (``add_msl_port(..., excite=
      False)`` → resistive Z0 termination via the port's own σ distribution,
-     ``rfx/sources/msl_port.py`` ``MSLPort`` docstring: "False → passive
-     matched termination only") — NOT relying on CPML alone to absorb the
-     guided mode;
+     ``rfx/api/__init__.py`` ``add_msl_port`` docstring: "``True`` →
+     resistive termination + active source; ``False`` → passive matched
+     termination only") — NOT relying on CPML alone to absorb the guided
+     mode;
   4. BOTH mesh classes: aligned (dx = h_sub/3, frac(h_sub/dx) = 0) and
      bisecting (dx = 80 µm, frac = 0.175, the MESH-ALIGNMENT RULE danger
      zone — docs/agent-memory/rfx-known-issues.md, "Added 2026-07-30").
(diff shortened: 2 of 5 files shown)


bk-squared:
Addressed in 77f3c72 (PR body updated to match). Summary, with the exact numbers:

**MAJOR 1 (provenance)** — fixed. The 472%-spread/−122.9 dB figures are #525's own ALIGNED-mesh measurement (#525's text: "at dx = 84.67 µm the net flux is ~1e-30 W with 472% ... on a record settled to −122.9 dB"); #525's earlier "Probe 1" comment independently shows the bisecting column was "smooth, monotone ... consistent across frequency" while the aligned column was sign-flipping. Corrected via an append-only dated note (original pre-declaration/post-run text untouched) plus an in-place fix to the ADMISSIBILITY WITNESS section's physics explanation. Confirmed which witness leg actually discriminates: reactive_fraction (0.963–0.995), not spread (0.011–0.067, passes the 0.5 threshold) — matches your numbers.

**MAJOR 2 (comparison)** — added explicitly. #525 comment 2's own convention-corrected bisecting figure (0.54–0.69 → 1.20–1.34) is quoted verbatim, alongside your broader reconstruction (1.076–1.387) and your unterminated-fixture reproduction (1.0898–1.3867, ~1% match). Also added: this script's aligned number (1.0083–1.0090) resolves #525's open doubt about PR #516's convention pairing — in PR #516's favour.

**MAJOR 3 (tautological assert)** — removed. Replaced with `_oracle_ratio()` (single choke point) + `_assert_half_factor_regression_is_caught()`, which plants a ½ at the `p_vi` build site only and requires the ratio to move (1.0 → 0.5). Verified standalone.

**MODERATE 1 (proxy mutation twin)** — added, zero new FDTD (recomputed from the same recorded fields). Re-ran the campaign for the new JSON fields: bisecting proxy-span ratio range is **[0.7443, 0.7452]**, hard outside [0.85, 1.15] — matches your number exactly. Aligned mesh: proxy == corrected span by construction (k_top == trace_lo there), honest null result, [1.0083, 1.0090].

**MODERATE 2 (DOF framing)** — reworded in place, cross-references `scripts/msl_flux_ratio_dof.py` explicitly, new DOF BLINDNESS subsection states both admissibility-witness legs are also scale-invariant under `(V,I)->(aV,I/a)` (neither reads V or I). `v_abs`/`i_abs` now recorded per cell in the JSON.

**MODERATE 3/4 + minors** — settling switched to max(peak)-of-tail (was mean); `msl_modal_voltage`'s docstring in `rfx/api/_sparams.py` now points at the bisecting result + artifact path; "first-COMMITTED" not "first-ever"; fixed the `excite=False` citation (it's `add_msl_port`'s docstring in `rfx/api/__init__.py`, not `MSLPort`'s); fixed "untermined" typo; PR body "three touched test files" → two.

Re-ran the full 2-mesh campaign (same fixture/params/band as before — this is a re-analysis producing new derived JSON fields, not a new investigation). Overall verdict unchanged: **HELD**, 30/30 admissible cells both meshes.

Verification: `pytest tests/test_msl_modal_voltage_and_wave_solve.py tests/test_mixed_port_sparam.py` — 50 passed. `ruff check` clean on all three touched files. Both new self-tests (`_assert_flux_convention_unchanged`, `_assert_half_factor_regression_is_caught`) pass standalone.

SHA: 77f3c7291b7c53130ac219285fd445fea132f4c8

bk-squared:
Addressed in 4735b12 (PR body updated to match).

**1. Guard mechanism honesty + coverage** — fixed both halves:
(a) The mutation now calls `_oracle_ratio(0.5*v, i, flux)` — routed through the real function, not the hand-inlined `0.5 * np.real(v * np.conj(i))` the docstring wrongly claimed matched it. Assertion 1 (the unmutated call reading 1.0) is now explicitly documented as the actual guard.
(b) Added the flux-side mutation: `_oracle_ratio(v, i, 0.5*flux)`, required to move the ratio to 2.0. Verified both directions against a real injected bug in `_oracle_ratio` (numerator-half → assertion 1 itself fails at ratio=0.5; denominator-half → fails at ratio=2.0), checked and reverted cleanly.

**2. Wording nits** — all four:
(i) CORRECTION header now states precisely what's protected: every decidable pre-declaration element (band, mesh classes, fixture, thresholds), not the prose — since the prose WAS corrected in place.
(ii) "the SAME unterminated fixture" → "the same fixture class, reconstructed (this geometry, far port removed)" in all three locations.
(iii) R3 line corrected to the stronger true claim. I independently re-verified it before writing it down: diffed the two committed JSON files (d58659d vs 77f3c72) field-by-field — `p_vi_real`/`flux_real`/`ratio`/`reactive_fraction` are bit-identical across both mesh rows, confirming the regeneration pass really was a re-analysis of the same physics, not a new investigation.
(iv) `rfx/api/_sparams.py`'s reactive-fraction number now annotated "(PR #549 review reconstruction)".

**Verification**: self-tests pass (both guard mutations behave as specified), `pytest tests/test_msl_modal_voltage_and_wave_solve.py tests/test_mixed_port_sparam.py` — 50 passed, `ruff check` clean on both touched files. No FDTD regeneration needed this round — only docstring/guard-function wording changed, `run_one()`'s computation is untouched.

SHA: 4735b126f82e20f2de269b2054b78f2440a3521f

Issue #557, opened 2026-08-04
WR-90 obstacle-lane f0 residual attributed: sub-cell aperture-edge offset, O(dx); flat interior PEC faces are exact


bk-squared:
Follow-up research to case 19 / #499, on the professor's direction after merge. The case committed a reproducible +12.08 MHz f0 residual (rfx above both the mode-matching and FDFD oracles on identical as-realized geometry) with its origin recorded as unattributed. It is now attributed, by measurement.

## Result

**The residual is a sub-cell effective-position offset of the iris APERTURE EDGE in the staircase raster.** Flat interior PEC faces carry none of it.

| probe | measurement | verdict |
|---|---|---|
| 4-mesh ladder (a/120, 90, 60, 45), one sensitivity estimator | residual_cells = **A + B·(dx/dx₉₀)**, A = −0.093 (mesh-invariant), B = −0.022; max misfit 0.0025 cell | pure dispersion rejected (would be ∝dx, ~2× too big at a/45); pure constant rejected (a/45 = −0.1370 vs −0.12) |
| closed cavity: interior full-width septa (filter's own midpoint recipe, t_c = 9 and 5) vs boundary-wall reference, TE10-halfwave eigenfrequency | f_septa − f_boundary = −0.000 MHz → face offset **< 0.001 cell**, thickness-independent | flat rasterized PEC faces register EXACTLY on the node plane — the (L_c+1)·dx cavity convention is now pinned 3 orders tighter than the case's indirect 0.04–0.17-cell bound |
| two-iris resonators vs the same oracle (per-face offsets; every claims-bearing row settled to column power ≤ 1.02, up to num_periods = 4800) | per face: 10.16 mm class −0.1021 (a/90) / −0.1189 (a/45); 6.096 mm class −0.0571 (a/90) / −0.0623 (a/60) / −0.0220 (a/45) | aperture-edge carries the offset; aperture-dependence non-monotonic; small-aperture (12-cell) linearity in dx BREAKS at a/45 |
| same-mesh closure at a/90: probe faces + oracle-measured cavity weights (w = [−0.059, +0.559, +0.559, −0.059], single-cavity ±0.2-cell perturbations, Σ sens −103.94 vs −103.20 all-four) | A_pred = −0.1136 vs filter total −0.1171 → **0.0035 cell (97 %)** | the filter's residual is quantitatively reconstructed from single-iris face measurements |
[...]

Issue #371, opened 2026-07-14
NU-path thin-conductor gaps: csg mask_on_coords argmin/half-open split + auto-mesh blind to thin conductors


bk-squared:
Two related latent bugs in the non-uniform (dz_profile) handling of thin conductors, surfaced during the #369 investigation (both are SEPARATE from #369, which was the NU runner silently dropping thin conductors — fixed in PR #370).

## Bug 1 — `Box.mask_on_coords` picks different cells for thin vs matching-thickness box on a non-uniform grid
`rfx/geometry/csg.py:76-116`, `_axis_mask` has two rules:
- thin branch (line ~104): `argmin(|coords - mid|)` — cell whose center is nearest the midpoint.
- volume branch (line ~108): half-open `(coords >= lo) & (coords < hi)`.

On a **genuinely graded** grid these can select a **different z-layer** for the same nominal position. Verified: a coarse-fine-coarse `smooth_grading` profile, 24 candidate z-offsets in the fine region → **10/24 disagreed** (off by exactly one layer). On a uniform grid all rules agree, so it does NOT fire there (which is why it was a no-op for #369's uniform-dz case). Low-severity but real: a thin sheet and a matching-thickness box at the same nominal z can land on different cells.

Repro / tracking test (xfail strict — flips to a visible XPASS if `_axis_mask` changes without fixing consistency; remove the marker when fixed):

```python
"""Focused regression test for the argmin-vs-half-open Box.mask_on_coords
rasterization split (found during the #369 investigation, SEPARATE from the
#369 symptom itself)."""
from __future__ import annotations
import numpy as np
import pytest
from rfx.auto_config import smooth_grading
from rfx.geometry.csg import Box
from rfx.geometry.rasterize import coords_from_nonuniform_grid
from rfx.nonuniform import make_nonuniform_grid


@pytest.mark.xfail(
    reason="known latent bug: Box.mask_on_coords's thin (argmin-nearest-"
    "center) and volume (half-open [lo,hi)) branches can select different "
    "z-cells for the same nominal position on a non-uniform grid -- "
    "confirmed during the #369 investigation (2026-07-14), not yet fixed. "
    "See rfx/geometry/csg.py:76-116, _axis_mask lines 103-108.",
    strict=True,
)
def test_thin_vs_matching_thickness_box_same_cell_on_nonuniform_grid():
    dx_coarse, dx_fine = 1.5e-3, 0.5e-3
    n_pre, n_fine, n_post = 10, 12, 10
[...]

Issue #48, opened 2026-04-15
NTFF box overlapping CPML corrupts far-field peak direction


bk-squared:
### Observed (VESSL #369367233541 / scripts/issue31_patch_validation.py)

Patch antenna NTFF at harminv f_res = 2.462 GHz reports peak at θ=72°, φ=270° — nearly grazing. A broadside patch should peak near θ=0° (zenith).

rfx preflight emits:
> UserWarning: NTFF box extends into CPML region along y-axis. Far-field results will be corrupted. Shrink NTFF box to interior.

The warning is accurate: NTFF surface integrals require the box faces to be in vacuum — CPML cells have `eps_r ≠ 1`, non-physical absorbing behaviour, and the equivalent surface currents become meaningless.

### Scope

1. Fix the validation script (`scripts/issue31_patch_validation.py`): shrink NTFF box so it sits at least `cpml_layers × dx` inside the interior on all faces.
2. Optionally: harden rfx itself — when the user passes an NTFF box that overlaps CPML, promote the warning to an error (currently it computes silently garbage output). See also #37 preflight tone.

### Acceptance
- Rerun patch viz; new peak lies at θ ≲ 10° (broadside).
- Broadside gain within 1–2 dB of what 05_patch_antenna.py reports as a sanity check.

bk-squared:
### Deep-dive verdict (2026-04-15)

The NTFF CPML-overlap fix (this issue's original scope) was correct but did not solve the off-broadside peak. Root cause is upstream.

**Four isolation cases** (VESSL 369367233551):
| Case | Mesh | Geometry | peak θ | broadside_ratio |
|------|------|----------|--------|-----------------|
| 1 | uniform | bare ez dipole | 90° | 0.010 ✓ |
| 2 | uniform 1mm | patch + ground | 76° | 0.70 |
| 3 | NU | patch + ground | 79° | 0.10 |
| 4 | NU | bare dipole | 90° | 0.010 ✓ |

Case 4 shows NU NTFF is healthy on a dipole → not an NU NTFF bug.

**Fifth run**: uniform dx=0.5mm patch with **lossless FR4** — harminv found no modes at all (cavity dispersion + coarse z).

**Sixth run** (369367233553): uniform dx=0.5mm patch with **lossy FR4 (tan δ = 0.02)** →
```
[peak] θ=6.2°  φ=339.8°  broadside_ratio=0.995
```
[...]

PR #564, opened 2026-08-05
fix(nonuniform): restore the missing bounding node so the NU grid realizes the requested extent (#562)

### LINKED ISSUE #562: NonUniformGrid builds every axis one cell short of the requested domain: PEC-bounded guides/cavities are geometrically wrong (+170 MHz measured, WR-90 TE101)
Found while measuring a mitigation for the case-19 / #557 aperture-edge residual. It is not specific to that case: **on the non-uniform path, the realized grid is one cell smaller than the requested `domain` on every axis.** Drawn geometry is unaffected (boxes are placed by physical coordinates), so this only bites dimensions that the *boundary* defines — i.e. exactly waveguides and closed cavities.

## Reproducer (deterministic, no FDTD, ~1 s)

```python
import numpy as np
from rfx.api import Simulation
from rfx.geometry.rasterize_grid import (coords_from_nonuniform_grid,
                                        coords_from_uniform_grid)
A, DX, NB = 22.86e-3, 0.254e-3, 8

def realized(**kw):
    sim = Simulation(freq_max=13e9, domain=(A, A, NB*DX), boundary="pec", dx=DX, **kw)
    g, c = ((sim._build_nonuniform_grid(), None) if kw else (sim._build_grid(), None))[0], None
    c = coords_from_nonuniform_grid(g) if kw else coords_from_uniform_grid(g)
    return g.shape, [(len(np.asarray(c[i])), float(np.asarray(c[i])[0]), float(np.asarray(c[i])[-1])) for i in range(3)]

print(realized())                                        # uniform
print(realized(dx_profile=np.full(90, DX)))              # NU
```

| build | shape | x samples | x span | vs requested 22.8600 mm |
|---|---|---|---|---|
| uniform `Grid` | (91, 91, 9) | 91 | 0.0000 – 22.8600 | **exact** |
| `NonUniformGrid` | (90, 90, 8) | 90 | 0.1270 – 22.7330 | **22.6060 mm — one cell short** |

Every axis loses one cell, including axes with no profile of their own (they inherit the same construction), and including `dz_profile` — so this is not specific to `dx_profile`.

## Direct physical confirmation (port-free, extractor-free)

Closed PEC cavity, requested `(L, A) = (15.748, 22.860) mm` at dx = 0.254 mm, TE101 by `rfx.harminv` (single mode in the window, reported by amplitude, no nearest-to-expected selection):

| | frequency | vs analytic at REQUESTED extents | vs analytic at REALIZED extents |
|---|---|---|---|
| analytic at requested (15.748, 22.860) | 11.558399 GHz | — | — |
| analytic at realized (15.494, 22.606) | 11.728728 GHz | +170.3 MHz | — |
| **rfx uniform path** | 11.557967 GHz | **−0.433 MHz** | n/a (realized = requested) |
| **rfx NU path** | 11.728274 GHz | **+169.9 MHz** | **−0.454 MHz** |

The NU solver is as accurate as the uniform one (−0.454 vs −0.433 MHz of 2nd-order Yee dispersion) — **about the geometry it actually built.** That geometry is not the one requested.

## Root cause

`rfx/nonuniform.py::make_nonuniform_grid` allocates **one sample per cell** (`nx = dx_full.shape[0]`, and `nx_interior = round(domain/dx)` when no profile is given), whereas `Grid` allocates `n_cells + 1` **nodes**. `coords_from_nonuniform_grid` then places sample *i* at the cell **centre**, while `coords_from_uniform_grid` places it at the **node** (`(arange - pad) * dx`). PEC is enforced at the outermost samples, so the realized wall-to-wall extent is the span of `n` samples = `n - 1` gaps = `requested − d_last`.

rfx already documents the convention split, in `rfx/geometry/rasterize_grid.py::_axis_cell_centers`:

> "This is off by half a cell from the legacy uniform-Grid convention (cell[cpml] center at 0) — the two conventions are not unified anywhere in rfx today."

What is not documented anywhere, and what this issue is about, is the **geometric consequence**: a PEC-bounded guide or cavity requested as `A` is built as `A − d`, and via the cutoff that is a large frequency error (1.1 % of width → +170 MHz here; for the guide-width axis alone, +37 MHz on a WR-90 TE101 resonator).

## How it stayed hidden

Only boundary-defined dimensions are affected. A raster readback of drawn geometry passes perfectly — in the run that found this, the irises rasterized to exactly 17/17 metal planes, iris thickness 2.0320 mm and cavity 15.8750 mm as intended, aperture 25 open nodes — while the guide the irises sat in was 1.11 % narrow. **Readback has to cover boundary-defined extents too, not only the geometry you drew.**

Discovery path, for the record: three NU configurations of the same two-iris resonator (fine bands with 4-cell margins, with 24-cell margins, and uniformly fine x with no grading interfaces at all) all read +46.5 ± 0.15 MHz against an oracle fed the *requested* width, versus +12.94 MHz on the uniform path. Feeding the oracle the *realized* width collapses all three to +9.5 – 9.8 MHz. The grading structure was irrelevant; the width was everything.

## Scope flag (structural, magnitude unmeasured)

`scripts/diagnostics/run_waveguide_wr90_nonuniform_broad_e5_sweep.py` builds `dy = graded_dy(A, DX, ratio)` — a profile summing to `A` — and compares against an analytic reference that uses `A` and `fc = c/(2A)`. By construction its effective guide is also narrow by about one edge cell. Whether that biases its gated magnitudes is **not measured here** and should not be assumed; it is a magnitude gate on a slab reflection, not a frequency gate, so the sensitivity is different from the resonator above.

## Proposed fix, staged (happy to carry it, review-gated)

1. **This issue + a failing regression test** pinning `realized == requested` for both builders (the reproducer above, table-driven over profile combinations).
2. **Blast-radius table**: every committed NU fixture/lane, classified by whether its claims-bearing dimensions are boundary-defined (affected) or geometry-defined (unaffected). Cheap — grid construction only, no FDTD.
3. **The fix**: unify on the node convention (`n_cells + 1` samples, coordinates at cumulative edges) so the NU path realizes the requested extent like the uniform path does, then re-measure the strongest available oracles (the closed-cavity analytic above; the NU lane's analytic Airy). Expected direction: agreements *improve*, because the geometry becomes the requested one.
4. Because step 3 moves committed NU numbers, it needs the measured before/after in hand and an independent approver — no silent fixture regeneration.

Interim workaround for users on today's code: add one cell to each PEC-bounded axis of `domain` (or one cell to the profile), and verify the realized extent from `coords_from_nonuniform_grid` before trusting any frequency.

bk-squared:
Closes #562. Draft until CI reports; the physics is measured and in the commit bodies.

## The defect

N cells are bounded by N+1 nodes. `make_nonuniform_grid` allocated **one E-node per profile cell**, and the non-uniform Yee stencil zeroes the last cell's H term (`inv_d_h[N-1] = 0`), so an N-cell profile realized only N-1 usable gaps: the wall-to-wall extent came out `sum(d) - d[-1]`. Independently, `coords_from_nonuniform_grid` placed sample *i* at the cell **centre** while the E-nodes the stencil differences sit on cell **edges** (`inv_d_e[i] = 2/(d[i-1]+d[i])` is a node's dual spacing), and while `coords_from_uniform_grid` returns edges — so rasterized material sat half a cell from the fields acting on it.

Drawn geometry is placed by physical coordinates and so rasterizes consistently; only **boundary-defined** dimensions were wrong. That is exactly waveguides and closed cavities, which is why this survived so long.

## Isolation (integer cell counts, so no dimension-snapping confound)

45 × 39 cells at dx = 1 mm, PEC box, TM110 by harminv:

| builder | realized extents | vs closed form @ requested | vs closed form @ realized |
|---|---|---|---|
| uniform `Grid` | 45.000 / 39.000 | **−0.007 %** | −0.007 % |
| `NonUniformGrid` | 44.000 / 38.000 | **+2.469 %** | −0.008 % |

Both solvers were accurate — only one built the requested cavity. Builder delta **+2.477 % (+126 MHz) → 0.000 % (bit-identical) after this PR.**

A WR-90 TE101 cavity requested as (15.748, 22.860) mm realized (15.494, 22.606) and read 11.728274 GHz; it now realizes the requested extents and reads 11.557967 GHz — the uniform path's value to every printed digit, −0.004 % from the closed form.

## The fix (~10 lines of production code)

1. **`_append_bounding_node`** — append one duplicate boundary cell per padded profile. That yields N+1 nodes *and* gives the far boundary node the mirror of the near one's coefficient: `inv_d_e[N] = 2/(d[N-1]+d[N]) = 1/d[N-1]` when the appended cell duplicates the last, matching the existing `inv_d_e[0] = 1/d[0]`. The appended cell's own H term is the one the stencil already zeroes, so it adds a node without adding physical extent.
2. **`coords_from_nonuniform_grid`** returns node positions (cumulative edges, origin at the first interior node) — the uniform builder's convention.

`tests/test_nonuniform_grid_extent_contract.py` pins the contract: realized extent == requested on every axis for both builders, node count == cells + 1, a uniform mesh expressed as profiles reproduces the uniform grid's coordinates elementwise, and a graded profile realizes its own sum with every node on a cumulative edge. **6 of its 7 cases failed before this change; the one that passed was the uniform builder.**

## Effect on committed analytic-gated tests (geometry unchanged)

| test | before | after |
|---|---|---|
| `test_nonuniform_xy_cavity_accuracy` TM110, x&y graded 4:1 | 2.43 % | **0.028 %** |
| `test_nonuniform_cavity_accuracy` TM111, z graded | 2.66 % | **0.025 %** |

~100× each, landing at the discretization floor. Both still pass; their gates (3.5 %, 4 %) are now ~125×/160× looser than the residual. **Tightening them is deliberately NOT in this PR** — gates pin measured envelopes, and these want re-measuring across resolutions first.

### A prior attribution is withdrawn, with the isolating measurement

`test_nonuniform_xy_cavity_accuracy`'s docstring recorded that same +2.4 % on both a graded and an ungraded profile and concluded it was "the standard coarse-Yee PEC-cavity extent-convention bias (~1 cell)". There is no such Yee/PEC bias — the uniform builder reads −0.007 % at integer cell counts — and both of its legs had gone through the NU builder, because a *uniform profile* still routes to `make_nonuniform_grid`. The docstring now carries the withdrawal, the before/after, and the two-builder isolation. **Its comparative finding is untouched and still valid: in-plane grading is faithful — both legs shared the same baseline.** (I also had to withdraw a claim of my own in #562: I wrote that the consequence was undocumented, which was false — that docstring had recorded it.)

## Blast radius, all classified (fast NU set: 196 passed, 0 failed)

- **7 distributed-composition tests** hand-built grid-shaped override arrays (eps, design mask, PEC occupancy) sized to CELL counts. Routed through one `_field_shape()` helper so the cells-vs-nodes rule is stated once instead of at 8 call sites. **API note for users:** a hand-built grid-shaped array on the NU path is now cells+1 per axis, as it always was on the uniform path.
- **3 node-count assertions** (`nz == len(profile) + pads`) gain the bounding node.
- **`test_coords_from_nonuniform_grid_cell_centers`** pinned the old centre convention by name → rewritten as `..._node_positions`.
- Supporting evidence the node convention was the intended one: `test_preflight_absorber_frame` already computed positions as cumulative edges, commented "mirrors Grid's node pad_x_lo convention" — only the count was wrong.

## A landmine this PR introduced, then closed

Four core sites sliced interiors as `d[pad_lo : len(d) - pad_hi]`, which now pulls in the trailing duplicate (or a pad cell) and overstates the extent by one cell — the same class of error as #562 itself. Consolidated into **one** definition, `rfx.nonuniform.interior_cells(d_full, pad_lo, pad_hi)`, rather than the rule restated four times. Subgridding's `nx - pad_hi` sites are NODE index bounds, not cell slices, and became more correct with the extra node.

## Open items for the approver — I am the author and am not merging this

1. **Committed NU fixtures are stale by construction.** They were generated under the one-cell-short geometry, and the NU gate tests *replay* fixtures rather than re-running FDTD, so they pass while describing the old geometry. Regeneration is a compute + review decision; expected direction is improved agreement (e.g. the NU broad-E5 lane's guide finally being `A` wide instead of `A − dy_edge`).
2. **Gate tightening** on the two cavity accuracy tests (see above) — evidence-first, separate change.
3. **Out of scope, unchanged:** the waveguide port samples its modal profile at cell centres on *both* builders (`cumsum(w) − 0.5·w`). That is a pre-existing internal port convention, not something this PR exposes or alters.

Found while measuring a mitigation for the #557 aperture-edge residual; the mitigation line itself turned out to be measuring this defect.

---

## Review round 1 (adversarial, PI-requested) — both blockers closed in `b5fe569`

F1 (smoother half-cell) and F2 (#325 advisory false negative) were real and mine; see that commit and the review reply for the measurements. F5/F6/F4/F7/F9 landed with them. F3 is now under test rather than only noted: the extent contract's new `cpml_layers>0` case pins interior span **and** absorber-depth symmetry, which is the reviewer's unclaimed improvement — the hi face carried one physical cell fewer than the lo face before this PR, so the CPML ramps were not mirror images. The out-of-scope NU+subpixel observation is filed as #565.

**Blast radius, amended (F3):** open-boundary NU runs change beyond the extent fix — the hi-face absorber gains its missing cell, so σ ramps become mirror-symmetric. An approver weighing fixture regeneration should read that as a second, independent reason NU numbers move.

R3: memory=this PR's own reduction pattern, reapplied to the two consumers the review found — "on a uniform profile the NU path must reduce to the uniform path elementwise" is what closes F1, and "call the production rasterizer instead of modelling it" is what closes F2; both were derived from the reviewer's measurements after re-deriving them here first (max|Δ| 2.4000 on eps_ex reproduced from the shipped state; the 4.50–5.50 mm box's real count of 1 confirmed against the centre model's 3 and a node-only model's 2)

### DIFF
--- a/docs/guides/sparameter_support_matrix.md
+++ b/docs/guides/sparameter_support_matrix.md
@@ -282,6 +282,21 @@ differentiation is implemented only with `normalize="flux"`.
 there is no corresponding nonuniform `sigma_override` AD-vs-FD test. Neither
 implementation nor gradient regression is RF validation.
 
+**Those nonuniform fixtures are STALE and pending regeneration (issue #562).**
+`tests/fixtures/waveguide_nu_broad_e5/waveguide_wr90_nu_flux_broad_e5_envelope.json`
+and
+`tests/fixtures/waveguide_nu_broad_e4/waveguide_wr90_nu_flux_broad_e4_comparison.json`
+were generated while the nonuniform grid realized every axis one cell short of
+the requested domain, so their guide is `a - dy_edge` wide rather than `a`, and
+their gate tests replay the frozen numbers rather than re-running FDTD — the
+tests therefore pass while describing the earlier geometry. Nothing was
+re-blessed: the numbers above stand exactly as measured, and the lane stays
+experimental, so the matrix is conservative rather than wrong. Regenerating
+them is expected to *improve* agreement (the guide finally being the requested
+width); until it happens, treat the cited magnitude differences as an upper
+bound obtained on a slightly narrow guide, and do not compare them against
+newly generated nonuniform numbers.
+
 **Setup restrictions:**
 
 - Prefer `normalize="flux"`; it uses a matched reference run for Poynting-flux
--- a/rfx/api/_execute.py
+++ b/rfx/api/_execute.py
@@ -1655,6 +1655,29 @@ class _ExecuteMixin:
         self._check_stencil_order_supported()
         from rfx.runners.nonuniform import run_nonuniform_path
 
+        # A grid-shaped override is (cells + 1) per axis, because N cells are
+        # bounded by N+1 E-nodes (#562) — the same shape the uniform path has
+        # always wanted. Caller code that sized arrays from the PROFILE length
+        # is one sample short, and without this check the mismatch surfaces
+        # deep inside XLA as an opaque broadcast TypeError (#562 review F9).
+        _grid = self._build_nonuniform_grid()
+        for _name, _arr in (("eps_override", eps_override),
+                            ("sigma_override", sigma_override),
+                            ("pec_mask_override", pec_mask_override),
+                            ("pec_occupancy_override", pec_occupancy_override),
+                            ("design_mask", design_mask)):
+            if _arr is None or not hasattr(_arr, "shape"):
+                continue
+            if tuple(_arr.shape) != tuple(_grid.shape):
+                raise ValueError(
+                    f"{_name} has shape {tuple(_arr.shape)} but this "
+                    f"non-uniform grid is {tuple(_grid.shape)}. Grid-shaped "
+                    f"arrays are one sample larger per axis than the cell "
+                    f"counts / profile lengths you requested: N cells are "
+                    f"bounded by N+1 field nodes (#562). Build the array from "
+                    f"the grid shape rather than from the profile length."
+                )
+
         result = run_nonuniform_path(
             self,
             n_steps=n_steps,
--- a/rfx/api/_preflight.py
+++ b/rfx/api/_preflight.py
@@ -2028,7 +2028,24 @@ class _PreflightMixin:
         if dz.size == 0:
             return
         edges = np.concatenate(([0.0], np.cumsum(dz)))
-        centers = 0.5 * (edges[:-1] + edges[1:])
+        # Sample positions must be the ones the RUN uses, and the occupancy
+        # rule must be the run's rule. This validator previously modelled both
+        # by hand — cell centres plus a bare half-open span — and was wrong on
+        # each count: coordinates are E-NODES (cell edges) since #562, and
+        # ``Box.mask_on_coords`` has a THIN-SHEET branch that snaps a box no
+        # thicker than its local cell onto a single nearest node (#48/#75/#371).
+        # With the hand model this check diverged from the rasterizer on boxes
+        # straddling a grading transition — the #325 signature it exists to
+        # catch — and went SILENT on one of them (#562 review, F2). So call the
+        # production path instead of imitating it: node positions through the
+        # same two steps the grid builder composes, and the shape's own mask.
+        # x/y are passed as single-sample arrays because only the z profile is
+        # needed here; either mask branch admits the box's own midpoint, so the
+        # combined mask's z profile is the z-axis mask.
+        from rfx.geometry.rasterize_grid import _axis_node_positions
+        from rfx.nonuniform import _append_bounding_node
+        z_nodes = np.asarray(_axis_node_positions(_append_bounding_node(dz), 0),
+                             dtype=np.float64)
 
         for entry in self._geometry:
             if not isinstance(entry.shape, Box):
@@ -2039,8 +2056,10 @@ class _PreflightMixin:
             if thickness <= 0.0:
                 continue
 
-            rasterized = (centers >= z_lo) & (centers < z_hi)
-            actual = int(np.count_nonzero(rasterized))
+            x_mid = np.array([0.5 * (float(c1[0]) + float(c2[0]))])
+            y_mid = np.array([0.5 * (float(c1[1]) + float(c2[1]))])
+            mask = np.asarray(entry.shape.mask_on_coords(x_mid, y_mid, z_nodes))
+            actual = int(np.count_nonzero(mask))
             local = (edges[:-1] < z_hi) & (edges[1:] > z_lo)
             if not np.any(local):
                 continue
--- a/rfx/geometry/rasterize_grid.py
+++ b/rfx/geometry/rasterize_grid.py
@@ -39,20 +39,26 @@ def coords_from_uniform_grid(grid) -> GridCoords:
     return GridCoords(x=x, y=y, z=z, shape=(nx, ny, nz))
 
 
-def _axis_cell_centers(d_arr: np.ndarray, cpml: int) -> np.ndarray:
-    """Cell-center positions for a padded cell-size array.
+def _axis_node_positions(d_arr: np.ndarray, cpml: int) -> np.ndarray:
+    """E-node positions for a padded cell-size array.
 
-    Matches the existing ``coords_from_nonuniform_grid`` z convention:
-    the first interior cell's LEFT edge is at physical position 0, so
-    its CENTER is at ``d[cpml]/2``. This is off by half a cell from
-    the legacy uniform-Grid convention (cell[cpml] center at 0) — the
-    two conventions are not unified anywhere in rfx today.
+    The nodes this grid steps sit on cell EDGES, not centres: the
+    non-uniform E update divides by ``2/(d[i-1]+d[i])``, the dual spacing
+    of a node straddling cells ``i-1`` and ``i``. Node ``cpml`` (the first
+    interior one) is the origin, so the interior spans
+    ``[0, sum(interior d)]`` and the last interior node lands exactly on
+    the requested domain face — the same convention
+    ``coords_from_uniform_grid`` uses (``(arange - pad) * dx``).
+
+    Until #562 this returned cell CENTRES, half a cell off the nodes the
+    stencil differences and half a cell off the uniform builder, which put
+    every rasterized material half a cell away from the fields acting on
+    it and (with the missing bounding node) made a PEC-bounded guide one
+    cell narrower than requested.
     """
     d = np.asarray(d_arr, dtype=np.float64)
     edges = np.insert(np.cumsum(d), 0, 0.0)           # len = n+1
-    offset = edges[cpml]                              # first-interior left edge
-    centers = (edges[:-1] + edges[1:]) / 2.0 - offset
-    return centers
+    return edges[:-1] - edges[cpml]                   # n nodes, origin at cpml
 
 
 def coords_from_nonuniform_grid(grid) -> GridCoords:
@@ -83,11 +89,10 @@ def coords_from_nonuniform_grid(grid) -> GridCoords:
             d_j = jnp.asarray(d_arr)
             cum = jnp.concatenate([jnp.zeros((1,), dtype=d_j.dtype),
                                    jnp.cumsum(d_j)])
-            offset = cum[pad_lo]
-            centers = (cum[:-1] + cum[1:]) / 2.0 - offset
-            return centers.astype(jnp.float32)
+            nodes = cum[:-1] - cum[pad_lo]
+            return nodes.astype(jnp.float32)
         d_np = np.asarray(d_arr)
-        return jnp.asarray(_axis_cell_centers(d_np, pad_lo), dtype=jnp.float32)
+        return jnp.asarray(_axis_node_positions(d_np, pad_lo), dtype=jnp.float32)
 
     x = _axis_centers(grid.dx_arr, pad_x_lo)
     y = _axis_centers(grid.dy_arr, pad_y_lo)
--- a/rfx/geometry/rasterize_grid.py
+++ b/rfx/geometry/rasterize_grid.py
@@ -21,7 +21,23 @@ from rfx.geometry._pole_keying import (
 
 
 class GridCoords(NamedTuple):
-    """Physical cell-center coordinates for rasterization."""
+    """Physical sample coordinates for rasterization.
+
+    **Two conventions live behind this type, deliberately.**
+    ``coords_from_uniform_grid`` and ``coords_from_nonuniform_grid`` return
+    E-NODE positions (cell edges) — the samples the Yee stencil differences and
+    that PEC/geometry act on. ``coords_from_fine_grid`` genuinely returns cell
+    CENTRES for the subgrid fine region (subgrid-fenced, unchanged).
+
+    Reading the wrong one is not cosmetic. This type's docstring said
+    "cell-center" for every producer until #562, and a consumer
+    (``compute_smoothed_eps_nonuniform``) named its variables ``centers_*``
+    accordingly and derived the node as ``centre - d/2``; when the NU producer
+    became node-based, that derivation silently inverted and displaced every
+    smoothed voxel by half a cell. Consumers that need both should derive the
+    centre FROM the node (``centre = node + d/2``), and any new producer must
+    say which convention it returns.
+    """
     x: jnp.ndarray  # (nx,)
     y: jnp.ndarray  # (ny,)
     z: jnp.ndarray  # (nz,)
@@ -29,7 +45,11 @@ class GridCoords(NamedTuple):
 
 
 def coords_from_uniform_grid(grid) -> GridCoords:
-    """Extract cell-center coordinates from a uniform Grid."""
+    """Extract E-NODE coordinates from a uniform Grid.
+
+    ``(arange - pad) * dx`` — node i at i*dx from the first interior node,
+    despite the historical "cell-center" wording this docstring carried.
+    """
     nx, ny, nz = grid.shape
     dx = grid.dx
     pad_x, pad_y, pad_z = grid.axis_pads
@@ -62,7 +82,7 @@ def _axis_node_positions(d_arr: np.ndarray, cpml: int) -> np.ndarray:
 
 
 def coords_from_nonuniform_grid(grid) -> GridCoords:
-    """Extract cell-center coordinates from a NonUniformGrid.
+    """Extract E-NODE coordinates from a NonUniformGrid (#562).
 
     All three axes use the per-cell spacing arrays (``dx_arr``,
     ``dy_arr``, ``dz``). The first interior cell on each axis is
@@ -80,7 +100,7 @@ def coords_from_nonuniform_grid(grid) -> GridCoords:
     pad_z_lo = int(getattr(grid, "pad_z_lo", grid.cpml_layers))
     nx, ny, nz = grid.nx, grid.ny, grid.nz
 
-    def _axis_centers(d_arr, pad_lo):
+    def _axis_nodes(d_arr, pad_lo):
         # Mesh-as-design-variable path: any axis cell-size profile may be
         # a JAX tracer. Route the cumsum / offset arithmetic through jnp
         # in-trace; fall back to the numpy path on concrete inputs to keep
@@ -94,9 +114,9 @@ def coords_from_nonuniform_grid(grid) -> GridCoords:
         d_np = np.asarray(d_arr)
         return jnp.asarray(_axis_node_positions(d_np, pad_lo), dtype=jnp.float32)
 
-    x = _axis_centers(grid.dx_arr, pad_x_lo)
-    y = _axis_centers(grid.dy_arr, pad_y_lo)
-    z = _axis_centers(grid.dz, pad_z_lo)
+    x = _axis_nodes(grid.dx_arr, pad_x_lo)
+    y = _axis_nodes(grid.dy_arr, pad_y_lo)
+    z = _axis_nodes(grid.dz, pad_z_lo)
 
     return GridCoords(x=x, y=y, z=z, shape=(nx, ny, nz))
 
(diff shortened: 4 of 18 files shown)


bk-squared:
## CI green on the full matrix — 13/13, out of draft

All four `fast-suite` shards and the 3-OS `clean-wheel-smoke` pass, so the whole suite sees this change, not only the NU subset I ran locally (196 passed / 0 failed there).

Recap of what is measured, for a reviewer coming to this cold:

- **The contract test is the anchor**: `tests/test_nonuniform_grid_extent_contract.py`, 6 of 7 cases red before the change, and the one that was green is the uniform builder — i.e. the test documents which builder was already right.
- **The two builders are now bit-identical** on an identical mesh (delta 0.000 %, was +2.477 % / +126 MHz on a 45 × 39-cell PEC box), and both read −0.007 % against the closed form.
- **Two committed analytic-gated tests improved ~100×** on unchanged geometry (2.43 % → 0.028 %, 2.66 % → 0.025 %), landing at the discretization floor. Their gates are deliberately left alone.
- **One prior attribution is withdrawn with the isolating measurement** (the "standard coarse-Yee PEC-cavity extent-convention bias" that turned out to be this defect on both legs), and the comparative finding that docstring existed to make — in-plane grading is faithful — is untouched.
- **A landmine this PR created is closed inside it**: four interior-slice sites that assumed `len(d) == pad_lo + N + pad_hi`, consolidated into one `interior_cells()` definition rather than the rule restated four times.

The three open items in the description are for the approver, not blockers on this diff: stale-by-construction NU fixtures (the gate tests replay fixtures, so they cannot see this change), gate tightening as a separate evidence-first change, and the port's cell-centre modal sampling which is common to both builders and out of scope.

I authored this, so I am not merging it.

bk-squared:
**Independent adversarial review (requested by the PI): the core grid fix is verified correct — every claim reproduced — but REQUEST-CHANGES on two consumers of the coordinate-convention change that were missed, both failing silently.**

## Verified independently (not from the PR tables)

- #562's reproducer at merge-base vs head: NU (90,90,8)/22.606 mm → (91,91,9)/22.860 mm; uniform byte-identical across both commits (every production hunk read; `coords_from_uniform_grid` untouched; the msl_port hunk is inside the NU branch after the uniform early-return).
- TM110 isolation re-run from scratch: builder delta +2.4768 % → 0.000000 % — NU and uniform return the same three modes to all six printed digits at head.
- Stencil math derived from the arrays: `inv_d_h[N-1]=0` lands on the appended duplicate (extent unchanged, `sum(d)−d[-1]` == profile sum); `inv_d_e` mirror symmetry exact in stored float32; PEC applied unconditionally so the duplicate's junk H never reaches a live E. A mode-split witness (asymmetric graded profile vs its reverse) shows max Δf = 0.78 Hz out of 6 GHz.
- The extent-contract test on main: 6 failed / 1 passed (the uniform case), verbatim as claimed.
- The withdrawn attribution survives scrutiny: `make_nonuniform_grid` forces `dx_profile[0]==dx_profile[-1]==dx`, so both legs lost the *same* edge cell — the comparative "grading is faithful" finding genuinely cancels the defect, and re-running the committed gate at head (0.028 % on now-correct geometry) confirms it independently.
- Two unclaimed **improvements** found: (a) the hi-face CPML was one node short of the lo face pre-PR (7 vs 8 physical cells, σ ramps not mirror-symmetric) — the extra node re-aligns it exactly; (b) the NU thin-sheet argmin now runs over nodes, which is where tangential Ex/Ey actually sit, *strengthening* the #371 not-a-bug record.

## BLOCKING

**F1 (CRITICAL) — `rfx/geometry/smoothing.py::compute_smoothed_eps_nonuniform` (:310-321, :380-385) was written against the centre convention and not updated.** It does `corner = coords.x − d/2`; pre-PR that recovered the node, post-PR it lands half a cell *outside*. Measured on an identical uniform mesh (8³, ε=4 slab): NU-vs-uniform smoothed-ε elementwise identity held at merge-base (max|Δ| = 0.0 on eps_ex/ey/ez) and is now broken (2.4 / 2.4 / 1.5). This re-introduces, on the public `subpixel_smoothing=True` path (`run()`/`forward()` kwarg → `runners/nonuniform.py:527-543`), the exact half-cell defect this PR's own commit message describes — and the committed gate (`test_subpixel_nonuniform.py::test_compute_smoothed_eps_nonuniform_interface_values`, a `1<ε<4` bound) passes under both conventions, so CI green does not cover it. Fix: derive `node_* = coords.*`, `centers_* = node_* + d/2`, and add the assertion that would have caught it — on a uniform profile, `compute_smoothed_eps_nonuniform == compute_smoothed_eps` elementwise (the sibling of this PR's own reduction test).

**F2 (HIGH) — the #325 graded-mesh rasterization preflight (`rfx/api/_preflight.py:2013-2073`) models the rasterizer with cell centres**, while the real rasterizer (`Box.mask_on_coords` on `coords_from_nonuniform_grid`) now uses nodes. Measured on `dz=[1mm]×5+[0.25mm]×8+[1mm]×5`: the validator matched the rasterizer on every probe case at merge-base; at head it diverges on 3 of 6 — all boxes straddling a grading transition, the #325 signature — and on the 4.50–5.50 mm case it goes **silent where it should warn** (validator counts 3, real count 1, threshold 2). The advisory named as a standing rule in the repo's CLAUDE.md is now a false negative in its target scenario. Fix: predict from `edges[:-1]` (nodes), or better call `Box.mask_on_coords` on the same coords the runner uses; add the 4.50–5.50 case as a regression.

**Same push, strongly recommended — F5**: the surviving "cell-center" names/docstrings in `rfx/geometry/rasterize_grid.py` (:24/:32/:65/:83/:97-99, plus the `_axis_centers` closure) are the mechanism by which F1 hid — `compute_smoothed_eps_nonuniform` names its variables `centers_x` *because of them*. Note `coords_from_fine_grid` genuinely still returns centres (subgrid-fenced), so the type now spans two conventions — say so at the type.

## Record before merge (not holding the PR alone)

- **F3**: the CPML hi-face realignment is a behavior change on every open-boundary NU run and is absent from the blast-radius list — an approver deciding on fixture regeneration needs it (it is an independent reason NU numbers move).
- **F4**: `rfx/runners/nonuniform.py:99-102` (and `tests/test_thin_conductor.py:258`) still say the thin-sheet conductor realizes at the cell CENTRE — the durable record of the #371 closure now states a convention the code doesn't use; update to "node" and note the argument is strengthened.
- **F6**: the new extent contract never exercises `cpml_layers>0` — the one part with a real open-domain delta (F3) is the untested part; one `boundary="cpml"` case (interior span == requested + pad symmetry).
- **F7**: the two frozen NU fixtures (broad-E4/E5) pass-but-describe the one-cell-short guide post-merge. Nothing was quietly re-blessed (verified), and both lanes are shadow-fenced, so the repo stays conservative-not-wrong — but no durable record of the staleness exists (`nu_known_limits.md` has no #562 entry). One ledger entry naming both fixtures as pending regeneration.
- **F8**: no `R3:` audit line on either commit or the PR body (the substance is all present; the artifact line is missing) — add to the body.
- **F9 (nit)**: the cells+1 override-array break surfaces as a raw XLA broadcast TypeError; a preflight shape check naming #562 would be kind.

## Flagged, out of scope here

NU + `subpixel_smoothing=True` end-to-end diverges wildly on a trivial closed-box fixture at **both** commits (uniform max|E|=0.0078 vs NU ~1.5e6, pre-existing) — low confidence the ad-hoc fixture is well-posed, but if it reproduces, the NU subpixel path has no end-to-end physics anchor at all; suggest a separate issue. Distributed NU with `pad_x>0` is benign by inspection but unexercised on this CPU box — suggest the GPU lane before the next release tag.

## Bottom line

The grid fix itself is right, minimal, and fully reproduced — with the F1/F2 consumer fixes (+F5) and the F3/F4/F6/F7/F8 records, this merges. The style of regression that closes it is the one this PR already invented for the grid: *on a uniform profile, the NU path must reduce to the uniform path elementwise* — applied to the smoother and the preflight model.

bk-squared:
## Both blockers closed in `b5fe569`, and F2 was worse than reported. Point by point.

Thank you — this is the review I needed rather than the one I wanted. Both blockers are mine and they share one cause: I flipped a producer's convention and audited the rasterizer's readers but not the smoother's or the preflight model's. Before acting I re-derived each of your measurements rather than trusting the report; both reproduced.

### F1 (CRITICAL) — confirmed, fixed, and your number reproduced exactly

`compute_smoothed_eps_nonuniform` needs both the node and the centre per axis and derived the node as `centre − d/2`. Post-#562 `coords` **are** nodes, so that subtraction landed half a cell outside. Reversed to `centre = node + d/2` — which is what the uniform sibling `_yee_coords` has always documented ("return the integer-grid positions; the caller **adds** +0.5·dx"). The uniform path was right all along; only the NU one inverted.

Your closing assertion is now committed as `test_compute_smoothed_eps_nonuniform_reduces_to_uniform`. Re-running the exact state that shipped at PR head: **max|Δ| = 2.4000 on eps_ex** — your 2.4, to four decimals. Green after the fix, and I verified it fires by mutation rather than assuming it would.

### F2 (HIGH) — confirmed, and the hand model was wrong on TWO counts

Fixing only the convention would not have closed this. `Box.mask_on_coords` has a **thin-sheet branch**: a box no thicker than its local cell snaps onto a single nearest node (#48/#75/#371). On your 4.50–5.50 mm box — extent exactly one local cell, straddling the transition — the three models give: centres **3**, nodes **2**, and the rasterizer realizes **1**. So a node-corrected hand model would still have been a false negative on the very case you found.

The model is therefore gone. The validator now calls the shape's own mask on node positions built by the same two steps the grid builder composes, so agreement is by construction, and a table-driven test asserts validator count == rasterizer count across six boxes.

That surfaced one more thing worth flagging: **`test_shifted_box_warns_with_actual_and_implied_counts` expected "rasterizes to 1 z cells", and the run actually realizes 2** (nodes at z = 1.0 and 2.0 mm, measured on the production path). The committed 1 was itself the centre-model artifact — the advisory had been reporting a number the run never produced. Updated to the measured 2 with the evidence in the test; the advisory still fires (2 < ceil(0.5 × 6.0)), which is what that test is about.

### F5, F6, F4, F7, F9 — landed in the same push

- **F5**: `GridCoords` now says two conventions live behind it — nodes from the uniform/NU builders, genuine centres from `coords_from_fine_grid` (subgrid-fenced) — and says explicitly that the blanket "cell-center" wording is *how F1 hid*, since the consumer named its variables `centers_*` from that docstring. `_axis_centers` → `_axis_nodes`. I also fixed the same misleading first line on `_yee_coords`.
- **F6/F3**: the extent contract gains a `cpml_layers>0` case pinning interior span **and** absorber-depth symmetry — so your unclaimed improvement (a) is under test, not just noted: lo and hi depths now measure equal at 8·dx. The blast-radius list in the PR body is amended to name the hi-face realignment as an independent reason NU numbers move.
- **F4**: corrected to E-node, with the note that the #371 argument is *strengthened* — the minimized quantity is now the distance to the plane the conductor actually lands on.
- **F7**: `docs/guides/sparameter_support_matrix.md` records both frozen NU fixtures as stale-by-construction and pending regeneration, including why their gate tests cannot see it (they replay rather than re-run) and that nothing was re-blessed.
- **F9**: mis-sized grid-shaped overrides now raise a `ValueError` naming #562 and the cells-vs-nodes rule.

### F8 and the out-of-scope items

`R3:` lines added to both the commit and the PR body. The NU + `subpixel_smoothing=True` divergence is **#565**, framed as "determine whether this reproduces", not as a claim — with your load-bearing second point carried over: if it does reproduce, that path has no end-to-end physics anchor at all, which is precisely how F1 shipped past CI. Distributed NU with `pad_x>0` is recorded for the GPU lane before the next release tag.

### Verification

255 passed / 0 failed across the nonuniform, subpixel, graded-rasterization, thin-conductor and smoothing selection; ruff clean. Full CI re-running on `b5fe569`.

Two things I want to name because they generalize past this PR. First, your framing — *"the style of regression that closes it is the one this PR already invented for the grid"* — is what actually produced both fixes: F1 closed by reduction-to-uniform, F2 by calling the production path instead of imitating it. Second, F2's deeper cause is that a **model** of a production behaviour existed at all; the fix that holds is deleting the model, not correcting it. I still owe you the same audit on anything else that models the rasterizer, and I have not found a third one — which is a statement about my search, not a guarantee.

bk-squared:
**Verify pass on b5fe569: CONFIRMED — merging.** Both blockers re-verified at the mechanism level:

- **F1**: the elementwise NU-vs-uniform smoothed-ε identity is restored (max|Δ| = 2.384e-07 = exactly one float32 ULP — the two paths now reach the same centre by different arithmetic, so the committed test's 1e-5 tolerance is the right call, noted here in case anyone tightens it later). The new reduction test was mutation-checked twice, including a wholesale revert of smoothing.py to the pre-fix state: it reds at worst = 2.4000 while the OLD gate stays green — locking the specific defect and re-confirming the old gate's blindness in one run.
- **F2**: the 3/2/1 escalation reproduces (thin-sheet snap makes even a node-corrected hand model false-negative on the 4.50–5.50 case) — deleting the model was necessary, not convenient. All 9 probe cases now agree; the validator's node array is elementwise identical to the runner's coords; the collateral "1 z cells → 2" correction is the measured production value with the advisory still firing.
- Core physics re-confirmed at this head: TM110 NU-vs-uniform builder delta 0.000000% (same three modes to six digits), CPML σ mirror-symmetry holds, uniform path byte-untouched hunk-by-hunk, #555 mutation ledger intact (16 red, same set), NU AD forward-grad 11 passed.

Three non-blocking test-coverage residuals are filed as **#568** (agreement test 2/6 effective — silent cases skip the assertion; the σ-mirror fix lacks a direct witness; test_thin_conductor.py's centre-convention wording) — per the reviewer's disposition, follow-up rather than another round: none can produce a wrong number, and holding a correct physics fix for docstring wording would be the wrong trade.

The F2 exchange is worth generalizing: the fix that held was deleting the model of production behaviour, not correcting it — same lesson as the smoother (derive centres FROM nodes, never the reverse). Both now carry regression locks in the style this PR invented: reduce-to-uniform, and call-the-production-path.

PR #569, opened 2026-08-05
test(nonuniform): close #568's residuals, and withdraw a #564 improvement claim that does not reproduce

### LINKED ISSUE #568: PR #564 follow-up: three test-coverage residuals from the verify pass (agreement-test silence, sigma-mirror assertion, thin-conductor wording)
Non-blocking residuals from PR #564's independent verify pass — all test/documentation debt; none can produce a wrong number. Recorded here so they don't rot with the PR thread.

1. **The new graded-rasterization agreement test is 2/6 effective**: four of its six cases don't fire the advisory, so `predicted is None` and the `predicted == real` assertion is skipped — and the skipped direction (validator silent where it should warn) is precisely the F2 failure mode the test exists to pin. Fix: in the else-branch, recompute the gate from `real` and assert the silence is justified.

2. **The CPML hi-face σ-mirror fix has no direct witness**: `test_extent_and_pad_symmetry_hold_with_cpml_pads` reds at the merge-base on its first two assertions, but its symmetry assertion passed under the OLD 7-cell hi-face too (node-position pad depth was already symmetric; the asymmetry lived in where `apply_cpml_e`'s `[-n:]` slice lands on the σ ramp). Fix: add `assert np.allclose(params.x_lo.sigma, params.x_hi.sigma[::-1])`.

3. **`tests/test_thin_conductor.py` still carries the retired centre convention at lines ~182/258/301/306/311**, including the load-bearing "#371: thin sheet must land on the NEAREST-centre layer" assertion message where the compared coords are now nodes. That file is the durable record of the #371 closure — the site a future reader will trust. (`rfx/runners/nonuniform.py`'s twin comment was already fixed in b5fe569.) Also: `docs/agent-memory/nu_known_limits.md` has no #562 entry yet (the support-matrix carries the stale-fixture record; the NU-limits ledger should cross-reference it).

Two observations, not defects: the new `_execute.py` shape guard builds the NU grid twice per forward() call (negligible vs the solve, AD-safe as measured — `test_nonuniform_forward_grad` 11 passed); and `_preflight.py` now imports two private names (`_axis_node_positions`, `_append_bounding_node`) across module boundaries — the right trade vs a copy that drifts, but a small public helper would make the coupling explicit.

Full verify evidence in the PR #564 thread.

bk-squared:
Closes #568.

Non-blocking debt from PR #564's verify pass. Two of the three items are places where **my** instruments were weaker than I claimed, and working item 2 turned up a claim in the #564 thread that is wrong — including in the part of that PR's body an approver would have read while deciding on fixture regeneration.

## Item 1 — the agreement test was 2/6 effective

When the advisory stayed silent, `predicted is None` and the `predicted == real` assertion was skipped. "Validator silent where it should warn" **is** the F2 failure mode the test exists to pin, so the test was blind in its own subject direction.

Now two-sided: derive whether the advisory *should* fire from the realized count and the #325 `implied` recipe, then assert either (fires **and** agrees with the rasterizer) or (justified silence). The `implied` formula duplicated into the test is cross-checked against the validator's own reported figure on every case that fires, so the copy cannot drift unnoticed — the failure mode that produced F2 in the first place.

## Item 2 — measured first, and the premise does not reproduce

#568 asked for `allclose(x_lo.sigma, x_hi.sigma[::-1])` as the direct witness for the hi-face CPML realignment. Measured across the merge before adding it, same request (60 cells, 8 CPML layers):

| | pre-#562 (`bb61494^`) | post-#562 |
|---|---|---|
| `nx` for 60 cells | 76 | 77 |
| **interior span** | **59.000 cells** | **60.000 cells** |
| lo / hi pad depth | 8.000 / 8.000 | 8.000 / 8.000 |
| σ mirror (`allclose(lo, hi[::-1])`) | **True** | True |

Two conclusions:

1. **The proposed assertion is not a witness** — it passes at the merge base. Both faces build σ from the same `n_layers` and the same boundary cell size, so the arrays are mirror images by construction regardless of how many usable cells sit behind them.
2. **The premise did not hold either.** The "hi-face CPML was one node short of the lo face (7 vs 8 physical cells, σ ramps not mirror-symmetric)" improvement recorded in #564's review — and repeated by me in that PR's body and in the merged blast-radius note — **did not happen**. Pad depths were symmetric on both sides; the only thing that changed is the interior span.

**That claim is withdrawn.** Concretely for an approver: NU numbers move for the *extent* reason alone, which removes the second, independent reason for fixture regeneration I gave. The regeneration case still stands on the extent change by itself.

Both CPML assertions stay, relabelled as invariants with the measurement in their docstrings rather than as evidence of a change — they are cheap, true, and would catch a future change that made the two faces derive σ independently (the NU z-faces already use independent cell sizes). The non-vacuity check matters for exactly that: a constant ramp would satisfy any mirror test.

## Item 3 — the #371 durable record

`tests/test_thin_conductor.py` still said cell CENTRE at five sites including the load-bearing `#371: thin sheet must land on the NEAREST-centre layer` assertion message, while the coordinates its argmin runs over are now nodes. Corrected, with the note that the #371 closure is **strengthened**: nodes are where tangential Ex/Ey sit, so the minimized quantity is now the distance to the plane the conductor is really realized on. `docs/agent-memory/nu_known_limits.md` gains a #562 entry (untracked in this public repo by design) carrying the stale-fixture record, the measurement above, and the withdrawal.

## Both review observations taken

- The two private cross-module imports in `_preflight.py` are replaced by one public helper, `rfx.nonuniform.node_positions_from_profile`. Explicit coupling is the right trade precisely because an implicit copy is what let F2 drift.
- The shape guard builds the NU grid only when an override array is actually present, so an override-free `forward()` pays nothing.

## Verification

421 passed / 0 failed across the nonuniform, subpixel, graded-rasterization, thin-conductor, smoothing and preflight selection; ruff clean. CI will run the full matrix.

R3: memory=verify the reviewer's claim before implementing their fix — F1/F2's numbers reproduced exactly (max|Δ| 2.4000 on eps_ex; the 4.50–5.50 mm box's real count 1 against a centre model's 3 and a node-only model's 2), and this round's improvement claim did not reproduce at all, which is why item 2 ships as a withdrawal plus an invariant rather than as the requested witness

### DIFF
--- a/rfx/api/_execute.py
+++ b/rfx/api/_execute.py
@@ -1660,15 +1660,21 @@ class _ExecuteMixin:
         # always wanted. Caller code that sized arrays from the PROFILE length
         # is one sample short, and without this check the mismatch surfaces
         # deep inside XLA as an opaque broadcast TypeError (#562 review F9).
-        _grid = self._build_nonuniform_grid()
-        for _name, _arr in (("eps_override", eps_override),
-                            ("sigma_override", sigma_override),
-                            ("pec_mask_override", pec_mask_override),
-                            ("pec_occupancy_override", pec_occupancy_override),
-                            ("design_mask", design_mask)):
-            if _arr is None or not hasattr(_arr, "shape"):
-                continue
-            if tuple(_arr.shape) != tuple(_grid.shape):
+        _overrides = (("eps_override", eps_override),
+                      ("sigma_override", sigma_override),
+                      ("pec_mask_override", pec_mask_override),
+                      ("pec_occupancy_override", pec_occupancy_override),
+                      ("design_mask", design_mask))
+        # Build the grid ONLY when there is an array to check, so the common
+        # override-free forward() does not pay for a second grid build
+        # (#568 observation 1 — negligible against the solve, but free to skip).
+        if any(_a is not None and hasattr(_a, "shape") for _n, _a in _overrides):
+            _grid = self._build_nonuniform_grid()
+            for _name, _arr in _overrides:
+                if _arr is None or not hasattr(_arr, "shape"):
+                    continue
+                if tuple(_arr.shape) == tuple(_grid.shape):
+                    continue
                 raise ValueError(
                     f"{_name} has shape {tuple(_arr.shape)} but this "
                     f"non-uniform grid is {tuple(_grid.shape)}. Grid-shaped "
--- a/rfx/api/_preflight.py
+++ b/rfx/api/_preflight.py
@@ -2107,10 +2107,8 @@ class _PreflightMixin:
         # x/y are passed as single-sample arrays because only the z profile is
         # needed here; either mask branch admits the box's own midpoint, so the
         # combined mask's z profile is the z-axis mask.
-        from rfx.geometry.rasterize_grid import _axis_node_positions
-        from rfx.nonuniform import _append_bounding_node
-        z_nodes = np.asarray(_axis_node_positions(_append_bounding_node(dz), 0),
-                             dtype=np.float64)
+        from rfx.nonuniform import node_positions_from_profile
+        z_nodes = np.asarray(node_positions_from_profile(dz), dtype=np.float64)
 
         for entry in self._geometry:
             if not isinstance(entry.shape, Box):
--- a/rfx/nonuniform.py
+++ b/rfx/nonuniform.py
@@ -180,12 +180,8 @@ def node_positions_from_profile(profile):
     worse, a hand-rolled copy, is how that check drifted in the first place —
     #562 review F2, #568).
     """
-    return _axis_node_positions_for(_append_bounding_node(profile))
-
-
-def _axis_node_positions_for(padded):
     from rfx.geometry.rasterize_grid import _axis_node_positions
-    return _axis_node_positions(padded, 0)
+    return _axis_node_positions(_append_bounding_node(profile), 0)
 
 
 def _profile_to_inv_arrays(profile_full: np.ndarray) -> tuple[jnp.ndarray, jnp.ndarray]:
--- a/rfx/nonuniform.py
+++ b/rfx/nonuniform.py
@@ -168,6 +168,26 @@ def interior_cells(d_full, pad_lo: int, pad_hi: int):
     return d_full[pad_lo : len(d_full) - pad_hi - 1]
 
 
+def node_positions_from_profile(profile):
+    """Interior E-node positions for a RAW (unpadded) cell-size profile.
+
+    ``N`` cells give ``N+1`` nodes, the first at 0 and the last exactly on the
+    profile's total extent — the same positions
+    ``coords_from_nonuniform_grid`` produces for the interior of a built grid.
+    Public because callers outside this module legitimately need the
+    convention (the #325 graded-rasterization preflight models what the
+    rasterizer will do, and modelling it with a private-import composition or,
+    worse, a hand-rolled copy, is how that check drifted in the first place —
+    #562 review F2, #568).
+    """
+    return _axis_node_positions_for(_append_bounding_node(profile))
+
+
+def _axis_node_positions_for(padded):
+    from rfx.geometry.rasterize_grid import _axis_node_positions
+    return _axis_node_positions(padded, 0)
+
+
 def _profile_to_inv_arrays(profile_full: np.ndarray) -> tuple[jnp.ndarray, jnp.ndarray]:
     """Return ``(inv_d_e, inv_d_h)`` — the E-update and H-update inverse
     cell-spacing arrays for a padded 1-D cell-size profile ``d``.
--- a/tests/test_nonuniform_grid_extent_contract.py
+++ b/tests/test_nonuniform_grid_extent_contract.py
@@ -182,42 +182,44 @@ def test_extent_and_pad_symmetry_hold_with_cpml_pads():
     assert lo_depth == pytest.approx(8 * DX, abs=_ABS_TOL)
 
 
-def test_cpml_sigma_ramps_are_mirror_images_across_the_two_faces():
-    """The two faces' sigma ramps are mirror images — an INVARIANT, not a
-    witness (#568 item 2, resolved by measurement).
+def test_absorber_depth_is_symmetric_across_the_two_faces():
+    """Both absorber faces must be the same physical depth (#568 item 2, redone).
 
-    #568 asked for this assertion as the direct witness for a hi-face
-    realignment. It is not one: run against the pre-#562 commit it PASSES
-    (measured, `bb61494^`), because both faces build sigma from the same
-    n_layers and the same boundary cell size, so the arrays are mirror images
-    by construction regardless of how many usable cells sit behind them. The
-    same measurement shows the premise did not hold either — pad depths were
-    8.000/8.000 cells before and after — so there is no realignment to
-    witness, and the "unclaimed improvement" recorded in the #564 review, and
-    repeated by me in that PR's body, is withdrawn.
+    History worth keeping, because two wrong answers preceded this one. #564's
+    review recorded that the hi-face absorber had been one cell shallower than
+    the lo face. #568 proposed `allclose(sigma_lo, sigma_hi[::-1])` as the
+    witness; that assert is a TAUTOLOGY (cpml.py builds the hi profile by
+    flipping the same base, so it holds at every commit) and is gone. I then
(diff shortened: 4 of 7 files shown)


bk-squared:
**Independent adversarial review (PI-requested): REQUEST-CHANGES — item 1(a) and 1(c) are ready to merge as-is (the agreement-test fix was mutation-verified twice and is the real thing), but the #564-claim WITHDRAWAL is itself refuted by measurement, and two of the new assertions cannot fail.**

**Finding 1 (HIGH, blocking) — the withdrawal is wrong; the original claim reproduces.** Measured convention-free from raw `dx_arr` + pad counts (so the centres→nodes change cannot bias it), 9/9 configurations, uniform and graded:

| | pre-#564 | post-#564 |
|---|---|---|
| interior span | **exactly N cells** (never 59) | exactly N cells |
| lo / hi absorber depth | **8.000 / 7.000** (hi one cell shallower, at every CPML depth) | 8.000 / 8.000 |

Root cause of the mis-measurement: the withdrawal's pre-#562 leg used `interior_last = nx - pad_hi - 1` — the `-1` exists only to skip the bounding node `_append_bounding_node` adds, which the pre-#562 array does not have. Applied there it lands one node short, silently charging the missing node to the interior and giving the hi pad a phantom 8th cell. Corroboration from slice geometry: pre-#564 the hi `[-n:]` ramp starts ON the interior boundary node while the lo ramp starts one cell outside it — exactly the asymmetry #568 item 2 described.

**What IS correct in your withdrawal (independently confirmed): the σ arrays mirror by construction** (`cpml.py:338` builds hi as a flip of the same profile), so #568's suggested `allclose(σ_lo, σ_hi[::-1])` assert is a tautology and you were right to reject it as a witness. Correct scope: withdraw the "σ ramps were not mirror images" WORDING; keep the physical claim (hi absorber one cell shallower → open-boundary NU numbers move for the absorber reason, and on a CPML-padded axis that is the ONLY reason — the extent reason never applied there since the interior span was already exact). As written, this PR would delete the only correct rationale for regenerating exactly the open-boundary fixtures most likely to need it, and write the wrong record into the gitignored durable-memory file where no reviewer can see it. (Ripple: #574's attribution of the measured regeneration improvement to "extent alone" is at most the PEC-axis half of the mechanism for a CPML-bounded propagation axis.)

Replacement witness that actually discriminates (reds at bb61494^ 8-vs-7, greens on main):
```python
cum = np.insert(np.cumsum(np.asarray(grid.dx_arr, float)), 0, 0.0)
lo_depth = cum[grid.pad_x_lo] - cum[0]
hi_depth = cum[grid.nx - 1] - cum[grid.pad_x_lo + N]
assert lo_depth == pytest.approx(hi_depth, abs=_ABS_TOL)
```

**Finding 2 (MEDIUM)** — the new mirror test cannot fail: `dx_x_lo`/`dx_x_hi` come from the same Python float and σ_hi is a literal flip of σ_lo; no NU-grid change can red it. Replace with the standoff witness above (or move to a CPML test module as a change-detector). **Finding 3 (MEDIUM)** — the non-vacuity guard is vacuous: `lo.min() == 0.0` exactly (innermost layer), so `max > 10*max(min,1e-30)` is unconditionally true; compare against `lo[lo>0].min()` or assert ramp monotonicity. **Finding 4 (MEDIUM)** — the silence branch hard-codes the validator's `actual <= 4` cutoff as contract (the 0-5mm case asserts 5-realized-vs-20-implied gets NO warning); gate the silence assertion on the resolution condition only, so widening the advisory later doesn't red this test. Findings 5-8 (LOW): `_implied_cells` hand-copy unverified on silent cases; the `_execute.py` lazy-build is a benign behavior change (verified: the runner still builds the grid, only error provenance moves) but buys nothing; stale `# 4. ... cell centers` comment in test_nonuniform_xy.py:110; `_axis_node_positions_for` one-line wrapper.

Verified good: item-1(a) mutation runs (validator-always-silent reds 2 cases where the old test caught 1; the restored F2 defect reds with the named message); item-1(c) thin-conductor wording factually correct, no stale sites; all three production hunks classified (preflight = byte-identical refactor removing private cross-module imports — the right trade; nonuniform.py = additive public helper; _execute.py = benign); ruff/api-reference/354-test selection green.

bk-squared:
## Finding 1 confirmed against me — the retraction was the mis-measurement. Fixed in `1ae4944`.

I reproduced your method before changing anything. Convention-free from the raw cell array, 60 interior cells with 8 CPML layers:

| | nx | interior (cell-based) | lo depth | hi depth |
|---|---|---|---|---|
| pre-#562 (`bb61494^`) | 76 | **60.000 cells** | 8.000 | **7.000** |
| post-#562 | 77 | 60.000 cells | 8.000 | 8.000 |

Your root-cause diagnosis is exact: my retraction derived the interior's last node as `nx - pad_hi - 1`, and that `-1` exists **only** to skip the bounding node the fix *adds*. Applied to the pre-fix array it lands one node short, charges the missing node to the interior — producing the phantom 59.000 span I reported — and hands the hi pad a phantom 8th cell. The cell-based interior span was 60.000 at both commits, so the number I built the retraction on never existed.

**The unified mechanism is better than either of my framings, and yours plus mine reconcile without either being wrong:** the missing bounding node is charged to whatever sits at the outer end of the axis.

- **PEC-bounded axis** (pad = 0) → the **interior** loses a cell. That is #562's headline and why the cavity tests improved ~100×.
- **CPML-padded axis** → the **hi absorber** loses one. That is #564's claim.

Same defect under two boundary conditions. I read a contradiction into two measurements that were answering different questions.

**Withdrawn correctly, and only this:** the "σ ramps were not mirror images" wording. You confirm independently what I had also found — `cpml.py:338` flips the same base, so `allclose(σ_lo, σ_hi[::-1])` holds at every commit and #568's proposed assert is a tautology. That part of my reasoning survives; the physical claim it was bundled with does not.

### What changed in `1ae4944`

- **Finding 1**: the physical claim is restored, with the measurement table and the unified mechanism, in the test docstring and in `nu_known_limits.md`.
- **Findings 2 + 3**: `test_cpml_sigma_ramps_are_mirror_images_across_the_two_faces` is gone — a test that cannot fail, whose non-vacuity guard was itself vacuous (`σ.min()` is exactly 0.0, so `max > 10*max(min, 1e-30)` is unconditionally true; that one stings, since guarding against vacuity was the whole point of adding it). Replaced by your witness as `test_absorber_depth_is_symmetric_across_the_two_faces`. **Verified discriminating**: reds at `bb61494^` with `absorber depth lo 8.000 vs hi 7.000 cells — the hi face is short by the bounding node`, greens on main.
- **Finding 4**: the silence branch no longer treats the validator's `actual <= 4` cutoff as this test's contract. It asserts the two directions the resolution condition settles and stays silent about the band where the cutoff alone decides, so widening the advisory later cannot red it.
- **Findings 7, 8**: stale "cell centers" comment; the one-line wrapper is gone.

### The process point I want on the record

The wrong retraction went into `docs/agent-memory/nu_known_limits.md`, which is gitignored — so an unreviewed error sat exactly where no reviewer could catch it, **and** it deleted the only correct rationale for regenerating the open-boundary fixtures most likely to need it. Your phrasing ("write the wrong record into the gitignored durable-memory file where no reviewer can see it") is the part I'll carry forward: a measurement that contradicts a reviewer belongs in a PR first.

**Ripple handled**: #574's "extent alone" attribution is corrected — it is at most the PEC-axis half of the mechanism for a CPML-bounded propagation axis. Findings 5 and 6 (the `_implied_cells` hand-copy on silent cases; the `_execute.py` lazy build buying nothing) I'm leaving as recorded LOW items rather than churning the diff further — say the word if you want either closed in this PR.

50 passed on the affected selection, ruff clean.

bk-squared:
**Verify round 2 (1ae4944): CONFIRMED — merging.** The standoff witness discriminates at the merge base for the physical reason (lo 8.000/hi 7.000, first two assertions passing so the failure localizes to the hi face), the withdrawal is now correctly scoped (σ-wording withdrawn as tautology — right; physical claim restored), and the unified framing — *the missing bounding node is charged to whatever sits at the outer end of the axis* (PEC → interior loses a cell = #562's headline; CPML-padded → the hi absorber loses one = #564's claim) — dissolves the apparent conflict better than either prior position. Finding 4's rework was proof-tested both directions (widened-advisory false-red removed, both true-red mutations retained). The durable-memory retraction-of-a-retraction carries the right direction AND the process lesson unprompted. Merging; the #575-body ripple is handled on #575.

Issue #545, opened 2026-08-03
Weekly slow lane: runner-shutdown cancellations are RECURRING — shards 1+3 killed two consecutive weeks (07-27, 08-03)


bk-squared:
The 2026-08-03 weekly run (30803500409) repeated the 2026-07-27 pattern exactly: **shards 1 and 3 killed by "The runner has received a shutdown signal" (exit 143)**, shards 2 and 4 completed, crossval-external green. Two consecutive weeks, the same two shards — this supersedes the "one-off GitHub infra event" reading recorded during the #509 triage (PR #532 era).

**What the 08-03 evidence adds:**
- Zero test failures before either kill — with PR #532's `pytest -v` change, every completed test's identity is in the streamed log and ALL are PASSED (including the three #532-restored coax locks: `test_under_resolved_annulus_is_flagged`, both `test_coax_end_to_end_ad` tests). So the lane's test-level health is green; the red is pure infra.
- Kill times: shard 1 at ~24 min elapsed, shard 3 at ~51 min — same shape as 07-27 (~20 and ~51 min).
- Same-shard recurrence across weeks argues against random preemption; pytest-split assigns shards deterministically, so shards 1/3 may share some property (longest wall-clock? memory footprint? a specific heavy test?) that interacts with runner health. Note #533 (durations file covers ~35% of the suite) means shard balance is nominal-only — rebalancing may also change which shards die.

**Suggested triage order:**
1. Regenerate `.test_durations` (#533) and rebalance — if the kills follow the heavy shards, balance is the lever.
2. Check memory pressure: ubuntu-latest = 7 GB; the single-process full run was OOM-killed historically (exit 143 = SIGTERM, consistent with OOM-adjacent runner termination). Add a lightweight RSS print between tests or a `--memray`-style sample to the weekly lane to discriminate OOM-kill from infra shutdown.
3. If infra: split the weekly lane's 4 shards into 2 workflows or add a retry-on-cancel wrapper.

Related: #533 (durations coverage), #509 (closed — carried the one-off reading this issue supersedes).

bk-squared:
[...]

PR #570, opened 2026-08-05
test(nonuniform): end-to-end anchor — the NU solve reduces to the uniform solve (#565 premise falsified)

### LINKED ISSUE #565: NU + subpixel_smoothing=True has no end-to-end physics anchor, and a trivial closed box diverges (pre-existing)
Split out of the #564 adversarial review's "flagged, out of scope" item, so it does not get lost with that PR.

## What was seen

On a trivial closed-box fixture, `subpixel_smoothing=True` on the non-uniform path diverges wildly against the uniform path — uniform `max|E|` ≈ 0.0078 versus NU ≈ 1.5e6 — and it does so at **both** the pre- and post-#562 commits, so it is not caused by the coordinate-convention change. The reviewer explicitly recorded low confidence that their ad-hoc fixture is well-posed, so **this issue is a request to determine whether the observation is real, not a claim that it is.**

## Why it is worth a ticket either way

The reviewer's second point is the load-bearing one: if it does reproduce, **the NU subpixel path has no end-to-end physics anchor at all.** What exists today:

- `tests/test_subpixel_nonuniform.py` unit-tests the smoother's interface values (a `1 < ε < 4` bound) and, as of #564, its elementwise identity with the uniform smoother on a uniform mesh. Both are *smoother* tests — neither runs a solve.
- No committed test compares a full NU + subpixel run against a closed form, an external solver, or the uniform path.

That gap is how #564's F1 shipped: a half-cell displacement of every smoothed voxel on the public `subpixel_smoothing=True` path passed CI, because nothing downstream of the smoother was gated. The identity test added in #564 closes that specific hole, but only at the smoother's boundary.

## Suggested first step (cheap, decisive)

A uniform mesh expressed as profiles now reduces to the uniform grid exactly (#562), so the same reduction works end-to-end: run the *same* closed box with `subpixel_smoothing=True` on both builders and compare fields. If they agree, the ad-hoc fixture was ill-posed and this closes with a note plus that comparison committed as the missing anchor. If they diverge, the divergence is isolated to the NU solve path with the geometry and smoothing held identical — which is the strongest possible starting point.

Related: #562, #564.

bk-squared:
Closes #565. Two results: the reported divergence is **not** one, and the anchor #565 actually asked for now exists and is verified to catch the defect it exists for.

## The divergence is a source-normalization factor, not an instability

#565 recorded that NU + `subpixel_smoothing=True` diverges on a trivial closed box (uniform `max|E|` 0.0078 vs NU ~1.5e6) at both the pre- and post-#562 commits, and asked whether it reproduces. It reproduces — and it is a constant:

| measurement | result |
|---|---|
| correlation between the two traces | **1.00000000** |
| peak time step | **same in both** |
| tail amplitude vs head amplitude | equal in both — no growth anywhere |
| amplitude ratio vs `dt/(ε₀·dV)` at dx = 1.0 / 0.5 / 2.0 mm | **0.99999963** every time |
| same, with `subpixel_smoothing=True` | **1.00000008** |
| residual after removing the factor | **2.7e-5** of full scale (9.5e-6 with smoothing) |

It also has nothing to do with subpixel smoothing: an air-filled closed PEC box shows the same factor, as does every profile-axis combination (`x`, `z`, `xy`, `xz`, `xyz`) and both source bandwidths tried.

**Cause — a real, user-facing inconsistency.** The same `add_source(position, component)` call means a **field increment** on the uniform path and a **current in amperes** on the non-uniform path (`rfx.nonuniform.make_current_source`, deliberately: resolution-independent injected power, Meep's convention). The traces therefore differ by exactly `dt/(ε₀·dV)` — about 2.2e8 at dx = 1 mm, which is easily large enough to read as an instability when a script is ported between paths.

Frequencies, S-parameters (normalized by a reference run) and field ratios are unaffected, which is why no committed test ever noticed. This PR documents it at `add_source` with the factor and the measurement. **Whether to unify the two normalizations is a design call I have not made** — the NU convention is arguably the better one — and it is filed separately as its own issue.

## The anchor

#565's load-bearing second point was that if the observation were real, the NU subpixel path would have no end-to-end physics anchor at all — which is exactly how #562's F1 (a half-cell displacement of every smoothed voxel) passed CI. Every committed NU test gates a frequency, an S-parameter, or the smoother's own output; none ran a solve on both builders and compared fields.

#562 made that comparison possible, so `tests/test_nonuniform_uniform_end_to_end_reduction.py` adds it. It pins the **relationship** rather than either side — same waveform, same timing, neither leg growing, ratio == `dt/(ε₀·dV)`, residual < 2e-4 — so changing either path's normalization fails here and the message says which. A second test asserts the factor tracks cell size as the closed form requires (two octaves), so a constant offset from any other cause cannot satisfy it.

**Verified to catch its target defect:** reintroducing F1's inverted `centre = node − d/2` derivation reds `test_nu_solve_reduces_to_uniform_solve_up_to_source_normalization`; restoring it greens.

**Deliberately not marked `slow`** — an anchor the fast suite deselects would have missed F1 the same way the old suite did. 6.3 s for all three cases.

R3: memory=measure before believing a framing, including my own — I filed #565 as "NU + subpixel has no anchor and a box diverges", and the divergence half turned out to be a normalization artifact. The anchor half was right, and closing it required explaining the artifact first rather than building a test around it

### DIFF
--- a/rfx/api/__init__.py
+++ b/rfx/api/__init__.py
@@ -796,6 +796,26 @@ class Simulation(
         content, ~1e-4). This prevents static charge accumulation on PEC
         surfaces.
 
+        .. warning::
+           **The waveform amplitude means different physical quantities on the
+           uniform and non-uniform paths**, so absolute field amplitudes are not
+           comparable between them. On the uniform mesh the amplitude is added
+           to E directly (a field increment); on a mesh built with any
+           ``d{x,y,z}_profile`` it is treated as a CURRENT IN AMPERES and
+           converted as ``E += (dt / eps) * I / dV`` for resolution-independent
+           injected power (``rfx.nonuniform.make_current_source``, Meep's
+           convention). The resulting traces differ by exactly
+           ``dt / (eps0 * dV)`` — about 2.2e8 at dx = 1 mm — which is large
+           enough to look like an instability when a script is ported from one
+           path to the other (that reading is what issue #565 recorded).
+
+           Frequencies, S-parameters (normalized by a reference run) and any
+           ratio of fields are unaffected. Measured cross-path agreement after
+           removing the factor: 2.7e-5 of full scale, 9.5e-6 with
+           ``subpixel_smoothing=True`` — see
+           ``tests/test_nonuniform_uniform_end_to_end_reduction.py``, which pins
+           the relationship.
+
         Parameters
         ----------
         position : (x, y, z) in metres
--- a/rfx/api/__init__.py
+++ b/rfx/api/__init__.py
@@ -834,11 +834,28 @@ class Simulation(
            snapshots, flux and NTFF outputs are reported as-is and are affected
            either way.
 
-           Measured cross-path agreement after removing the applicable factor:
-           1.5e-5 of full scale (PEC), 1.1e-4 (CPML), 9.5e-6 with
-           ``subpixel_smoothing=True`` — see
-           ``tests/test_nonuniform_uniform_end_to_end_reduction.py``, which pins
-           both factors. See issue #571 for whether to unify the two
+           Measured cross-path agreement after removing the applicable factor,
+           all four combinations — the fourth is the one to read:
+
+           ==================  =====================  ====================
+           uniform boundary    ``subpixel_smoothing``  agreement (of scale)
+           ==================  =====================  ====================
+           ``pec``             off                    1.5e-5
+           ``pec``             on                     9.5e-6
+           ``cpml``            off                    1.1e-4
+           ``cpml``            **on**                 **1.98e-2**
+           ==================  =====================  ====================
+
+           **The open-boundary + subpixel case does NOT reduce**: a systematic
+           0.18 % amplitude error and 2 % of the waveform, stable across a 4x
+           record length. That is an open defect (issue #582), not a tolerance —
+           so do not read the three good rows as clearance for subpixel
+           smoothing on an open domain, which is the combination issue #565 was
+           about.
+
+           ``tests/test_nonuniform_uniform_end_to_end_reduction.py`` pins all
+           four (the last as ``xfail(strict=True)``, so it announces itself when
+           fixed). See issue #571 for whether to unify the two source
            conventions.
 
         Parameters
--- a/rfx/api/__init__.py
+++ b/rfx/api/__init__.py
@@ -799,22 +799,47 @@ class Simulation(
         .. warning::
            **The waveform amplitude means different physical quantities on the
            uniform and non-uniform paths**, so absolute field amplitudes are not
-           comparable between them. On the uniform mesh the amplitude is added
-           to E directly (a field increment); on a mesh built with any
-           ``d{x,y,z}_profile`` it is treated as a CURRENT IN AMPERES and
-           converted as ``E += (dt / eps) * I / dV`` for resolution-independent
-           injected power (``rfx.nonuniform.make_current_source``, Meep's
-           convention). The resulting traces differ by exactly
-           ``dt / (eps0 * dV)`` — about 2.2e8 at dx = 1 mm — which is large
-           enough to look like an instability when a script is ported from one
-           path to the other (that reading is what issue #565 recorded).
+           comparable between them.
 
-           Frequencies, S-parameters (normalized by a reference run) and any
-           ratio of fields are unaffected. Measured cross-path agreement after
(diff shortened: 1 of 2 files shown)


bk-squared:
**Independent adversarial review (PI-requested): REQUEST-CHANGES — the diagnosis and the anchor are sound (reproduced to the digit; the anchor mutation-verified against THREE defect classes: the F1 smoother inversion, the #562 bounding-node drop, and a 1% normalization shift — all red), but the documented factor is wrong by 4.6x on the more common boundary, in the public API reference, with no test covering it.**

**Finding 9 (HIGH, blocking):** the new `add_source` docstring states unconditionally "the traces differ by exactly `dt/(eps0*dV)`". The uniform path's source normalization is BOUNDARY-DEPENDENT (`rfx/runners/uniform.py:302-311`): PEC → raw field add (factor `dt/(ε₀·dV)` ✓ measured ratio/documented = 1.000000), but **CPML/UPML → already Cb-normalized, so the real cross-path factor is `1/dV`** (measured: ratio/documented = 4.644 at dx=1mm) — and open domains are rfx's common case. The anchor test uses `boundary="pec"` in every leg, so nothing catches it; the same unconditional table is in issue #571. Fix: state the invariant (the NU path divides by cell volume; the uniform path never does) + both factors with their boundary condition; add a `boundary="cpml"` leg to the factor test so the docstring has a gate; correct #571's table in the same pass.

**Finding 10 (MEDIUM):** the anchor's source sits at z = NZ·dx/2 — EXACTLY on the slab's upper face; the factor prediction rides on Box's exclusive-upper-bound convention (if it ever became inclusive, source-cell ε_r=4 → false red unrelated to the reduction). Move the source to a clearly-vacuum node (e.g. 3·dx). **Findings 11-12 (LOW):** the anchor covers the uniform-profile reduction only (graded-mesh `dV` uses the forward cell, not the node's dual — pre-existing, one disclosure sentence); the anchor deliberately pins the #571 inconsistency as an invariant (well-signposted — noting for the #571 owner).

Verified good: the divergence reproduces exactly and IS a constant (correlation 1.0000000000, residual after factor removal 9.5e-6); the mechanism at `rfx/nonuniform.py:598` is deliberate, documented pre-PR, Meep-convention — filing #571 rather than making the design call in a test PR was right; runtime 6.59s not-slow-marked is the correct lane choice and the stated reason (a fast-lane-deselected anchor would have missed F1) is exactly right; merge order 569→570 is conflict-free (trial-merged, 30 passed).

bk-squared:
**Verify round 2 (7faa024): CONFIRMED — one clause before merge (public-carrier claim-accuracy, session precedent).** Everything verified: the boundary-dependent factor table is right (my 4.644028 reproduced), the mutation guard on the docstring works (reverted factor → the new cpml leg reds), the moved source no longer rides Box's inclusion convention, and **the CPML leg you added found a real bug on first contact** — #582 reproduced from scratch to every printed digit (cpml+subpixel factor ratio 1.001804, residual 1.978e-2, record-length-independent at 600/1200/2400 steps — a systematic 0.18% amplitude error), xfail(strict=True) is the right pin. The Kerr/chi3 caveat is a correctness catch nobody asked for.

**R1 (the one clause):** `rfx/api/__init__.py:841-843` quotes agreement figures 1.5e-5 (PEC) / 1.1e-4 (CPML) / 9.5e-6 (subpixel) — all true, but the missing fourth combination `cpml + subpixel_smoothing=True` is **1.98e-2, three orders worse and an open defect**. A reader of this public warning concludes subpixel is fine on open domains — the exact combination #565 was about. Add the fourth number + the #582 pointer.

Fold in if touching the file anyway: the stale `:53` timing comment ("6.5 s for all three" → measured 29.8 s for five — and #545 tracks fast-lane time pressure), and a one-line "superseded, see comment below" atop #571's body (the correction lives only in a comment; the body still shows the unconditional table). Push and I merge — no further round.

bk-squared:
**Final verify (2a1550a): CONFIRMED — merging.** R1 went past the ask (the four-row table + the interpretive guard 'do not read the three good rows as clearance' forecloses the exact misreading), the timing/before-after and #571's superseded-banner both verified. One LOW carried to the #582 closure: the pec+smoothing row reads 9.5e-6 (the pre-Finding-10 fixture's value); measured on this head it is 8.2e-6 — conservative, same order, fold into whatever rewrites the fourth row when #582 closes.

PR #572, opened 2026-08-05
coax two-port: wire eps_scale AD channel (#489 leg 3)

### LINKED ISSUE #489: Coaxial 2-port / transition S-parameters (roadmap #2)

bk-squared:
## Summary

Wires the `eps_scale` differentiable design channel through
`compute_coaxial_two_port`, mirroring the 1-port `compute_coaxial_line_reflection`'s
own contract exactly. This is the AD leg (leg 3) of issue #489's active dev
track — it does **not** close #489 (the transition/TE11-referee leg remains
open).

- `Simulation.compute_coaxial_two_port(..., eps_scale=...)`: same name/semantics
  as the 1-port sibling (`eps_r <- eps_r * eps_scale`), applied once after both
  feed stamps so it reaches both drives' FDTD runs.
- Both drives route voltage extraction through the already-differentiable
  `coaxial_line_plane_voltage_jnp` (the same twin the 1-port path already uses)
  instead of the concrete `coaxial_line_plane_voltage`.
- `_assemble_coaxial_two_port_from_voltages` (`rfx/api/_sparams.py`) gained a
  jnp assembly branch that reuses the extractor's own existing jnp core
  (`coaxial_line_reflection_from_plane_voltages(..., _prefer_jnp=True)`) — not
  re-derived.
- `solve_two_port_from_wave_amplitudes` (`rfx/sources/coaxial_port.py`) gained
  a jnp core (`_solve_two_port_from_wave_amplitudes_jnp`, batched
  `S = B @ inv(A)` via `jax.numpy`'s native batched `linalg.inv`/`linalg.cond`).
- `eps_scale=None` stays byte-identical to the pre-AD numpy path — the concrete
  branches were moved into `else:`, not rewritten. Regression-locked by the
  existing tightly-pinned `slow_physics` fixture in
  `tests/test_coax_two_port_fdtd.py` (unchanged, still green).

### The #468 / #559-B1 finding (found on the first real measurement, not
hypothetical)

Every dispatch point (`_assemble_coaxial_two_port_from_voltages`,
`solve_two_port_from_wave_amplitudes`) uses an explicit `_prefer_jnp` flag,
not tracer-only auto-detection. A concrete FD probe on the `eps_scale` path
(e.g. one side of a finite-difference cross-check, called eagerly rather than
under `jax.grad`) is **not** a `jax.core.Tracer` — tracer-only dispatch was
measured during this change to silently fall through to the strict NumPy
`lstsq` in `coaxial_line_reflection_from_plane_voltages` and raise
`numpy.linalg.LinAlgError: SVD did not converge` on a marginal fit the jnp
path handles fine. Fixed by threading an explicit flag from
`compute_coaxial_two_port`'s own `eps_scale is not None` check down through
every dispatch point, so the AD call and an eager FD call are guaranteed to
hit the identical branch — the exact class the task briefing (and PR #468 /
#559-B1) warned about, caught empirically rather than by inspection alone.

There is no passivity-projection step on this path to disable on both sides:
`compute_coaxial_two_port` never calls `_project_passive` (unlike
`compute_msl_s_matrix`/`compute_mixed_s_matrix`) — its only post-assembly step
is `_finalize_sparam_result` → `_warn_if_nonpassive_smatrix`, a WARN-only
diagnostic that returns `result` unchanged either way and is already
tracer-safe.

### Gate

Flips the two committed blocker pins in `tests/test_coax_two_port_ad.py` into
positive gates:
- the name-predicate test now asserts the `eps_scale` channel **exists**
  (keeping its positive control),
- the `TracerArrayConversionError` pin is replaced by an AD-vs-FD gradient
  gate on a small real-FDTD fixture (domain `(0.008, 0.008, 0.012)`,
  `n_steps=600` (settled: both drives < -60 dB), `probe_count=3`).

Also flips `compute_coaxial_two_port`'s classification in
`tests/test_ad_surface_contract.py` from `NOT_TRACEABLE` to `GRAD_SAFE`.

**Severed-tape falsifier** (performed by hand, R3 discipline): wrapping the
traced `eps_scale` argument in `jax.lax.stop_gradient` before it reaches
`compute_coaxial_two_port` (reproducing the #483/#515 class of defect)
produced `grad == 0.0` exactly — confirming the new gate's floor
(`3.0e-4`, ~180x below the measured `|g_ad|`) is a real falsifier, not vacuous.

**Measured** (current, on the SETTLED fixture — `N_STEPS=600`,
`settling_db=[-61.36, -63.88]` dB; the float64 FD referee
`scripts/coax_two_port_ad_fd_f64_referee.py` regenerates these):

```
g_ad (f32)              = -1.205334e-02
g_fd (h=2e-3, float64)  = -1.199238e-02   rel_err = 0.00508
FD resolving power      = 2.16e11 ULP of the loss (committed floor 1e4)
```

Gate threshold **2%** = `gate_from_envelope(0.01053, quantum=100)` from the
worst rel_err in the converged window `{1e-3, 2e-3, 5e-3}` of the committed
7-point f64 sweep; the span x threshold >= 100 anchor is a collection-time
assert. (An earlier revision of this PR measured an f32-only 2-point sweep on
an under-settled `N_STEPS=400` fixture — gate 1%, `g_ad=-5.475e-02`; those
numbers are superseded by the settled-fixture f64 measurement above.)

`eps_scale=1.0` (jnp path) vs `eps_scale=None` (numpy path) on the same
fixture: max abs diff `3.6e-6`, max rel diff `1.9e-4` — confirms the AD
wiring doesn't change the physics.

**CORRECTION (2026-08-05, adversarial review):** an earlier revision of
this PR claimed *"no CPU float64 comparator is possible for this method"*
because of the `precision='float32'` pin. **That claim was false and is
withdrawn.** The pin governs the FDTD *field* arrays only; the comparator's
resolving power follows the *loss* dtype, which follows the DFT plane
accumulator — `rfx/probes/probes.py` keys it off `jax.config.x64_enabled`
independently of the precision pin. Under a scoped `enable_x64()` the loss is
float64 (fields still f32 — the MSL f64 referee's own documented baseline),
and the committed referee `scripts/coax_two_port_ad_fd_f64_referee.py` now
measures a 2.16e11-ULP resolving power (vs 9.2e3 for the f32-loss form, and
only ~404 ULP on the settled fixture — below the repo's 100x anchor — so the
f64 instrument was *necessary*, not optional). Attempt 1's two specific
errors: an unconditional `freqs` float32 downcast masked the effect, and the
check used "did the numbers move?" instead of `loss.dtype`. The
`FutureWarning` at `rfx/boundaries/cpml.py:855` concerns the H-field carry —
the thing the MSL referee also leaves at f32. `docs/guides/
sparameter_support_matrix.{md,json}` carry the corrected statement in
matching phrasing.

## Test plan

- [x] `tests/test_coax_two_port_ad.py` (all 5, incl. `slow_physics`)
- [x] `tests/test_coax_two_port_fdtd.py` (39, incl. the tightly-pinned
      `slow_physics` real-fixture regression lock — unchanged numeric gates
      still pass, confirming the concrete/eager path is byte-identical)
- [x] `tests/test_coax_two_port_solve.py` (17)
- [x] `tests/test_coax_end_to_end_ad.py` (2, `slow_physics` — 1-port sibling's
      own AD gate still green)
- [x] `tests/test_coaxial_line_extraction.py`, `tests/test_coaxial_line_calibration.py`
- [x] `tests/test_ad_surface_contract.py` (classification flip verified)
- [x] `tests/test_sparameter_support_contract.py` (doc/JSON contract)
- [x] Combined run, all markers included: **144 passed**, 3 warnings (2 pre-existing
      unrelated + the documented `FutureWarning` from the scoped-x64 FD check), 0 failures, 320.78s
- [x] `ruff check rfx/ tests/ --select E,F,W --ignore E501,F401,E741,E731,E701,E702,E402`
      → all checks passed
- [x] `test_compute_coaxial_two_port_drive_index_matches_physical_port`
      (drive-index/column semantics pin) — still green, unaffected


bk-squared:
Addressed all 10 findings from the adversarial review in `04c767a`.

**F1+F3 (blocking, the "not achievable" claim)**: confirmed wrong, empirically. `precision='float32'` governs only the FDTD field/CPML carry arrays. The loss dtype follows the DFT plane accumulator, which `rfx/probes/probes.py`'s `init_dft_plane_probe` keys off `jax.config.x64_enabled` independently of the precision pin. A scoped `enable_x64()` around the forward call is sufficient — `eps_scale`/`freqs` don't need to be float64 themselves (forcing them to float64, as the original attempt did, is counterproductive: it mixes `materials.eps_r` dtype with the float32 field carry and produces the misleading `FutureWarning` that caused the original misdiagnosis).

Rebuilt the gate around a real float64 FD reference. New script `scripts/coax_two_port_ad_fd_f64_referee.py` (mirrors `scripts/msl_ad_fd_f64_referee.py`) runs a 7-point h sweep (2e-4 to 2e-2), verifies `loss.dtype == float64` every evaluation, reports resolving power in ULPs, and identifies the converged window (tightest-agreeing adjacent h pair). The gate's threshold is derived via `gate_from_envelope` from the worst rel_err in that converged window (h ∈ {1e-3, 2e-3, 5e-3}), not the full diagnostic sweep — the two ends of the full sweep are explained, not hidden: small h (≤5e-4) is floor-noise-dominated (float32 field carry rounding doesn't vanish just because the loss computes at float64), large h (≥1e-2) is ordinary O(h²) truncation from real curvature at multi-percent perturbations. Full measured table and reasoning are in the test module's docstring.

Gate now: `h=2e-3`, threshold **2%** (was an unfounded 1% derived from an f32-only comparator that measured `span × threshold = 92.06` — under this repo's own ≥100 floor). Measured: `g_ad = -1.2053e-2` (f32, as shipped), `g_fd = -1.1992e-2` (f64, h=2e-3), **rel_err = 0.51%**, resolving power **2.16e11 ULP** (floor 1e4). The `loss.dtype == float64` check and the resolving-power floor assert both run and pass, before the accuracy assert, per the #527 discipline.

**F6 (settling, fixed first)**: the AD fixture's concrete twin read `settling_db = [-37.13, -37.70]` dB at `N_STEPS=400` — short of the -40 dB rule. `N_STEPS=600` clears it: `[-61.36, -63.88]` dB. All numbers above are measured on the settled fixture.

**F2**: `python scripts/check_api_reference.py --write` — `docs/guides/api_symbol_inventory.json` now carries `eps_scale`; check is green. `docs/api` is gitignored (`docs/.gitignore`'s `/api/`), not part of the tracked diff.

**F4**: the unity test's ">1000x margin" claim was wrong. Measured (settled fixture): margin is **5.48x vs rtol alone**, **70.75x combined bound**. Docstring corrected to state the real numbers.

**F5**: confirmed — `cond_warn` is silently inert whenever `eps_scale` is passed (the `_prefer_jnp` early return in `solve_two_port_from_wave_amplitudes` precedes the concrete path's degeneracy warning). By design, mirroring the `settling_db`/`status` losses (same "can't Python-branch on a tracer" reason) — now disclosed in `compute_coaxial_two_port`'s own docstring.

**F7**: `CoaxialTwoPortResult`'s docstring now documents the `"differentiable"` status value and the `settling_db`/`cond_warn` losses under trace.

**F8**: corrected — `tests/test_coax_two_port_solve.py` collects 17 tests, not 16.

**F9**: added a phase-bound assert to the unity test (measured `2.675e-4` rad, bound `1e-3` rad).

**F10**: `_assemble_coaxial_two_port_from_voltages` now raises a helpful `ValueError` for list inputs on the traced/`_prefer_jnp` path (was an opaque `AttributeError: 'list' object has no attribute 'shape'` — verified by hand).

`docs/guides/sparameter_support_matrix.{md,json}`'s `ad_evidence` field for the coax row corrected to match — no more "not achievable" claim, same phrasing across both carriers.

**Verification**: combined suite (`test_coax_two_port_ad.py` + `test_coax_two_port_fdtd.py` + `test_coax_two_port_solve.py` + `test_coax_end_to_end_ad.py` + `test_coaxial_line_extraction.py` + `test_coaxial_line_calibration.py` + `test_ad_surface_contract.py` + `test_sparameter_support_contract.py`, all markers): **144 passed**, 3 warnings (all expected/documented — the FutureWarning is the harmless CPML-coefficient mixed-precision note discussed above), 0 failures, 321.68s. `ruff check rfx/ tests/ scripts/coax_two_port_ad_fd_f64_referee.py` → all checks passed. `check_api_reference.py` → OK.

Head SHA: `04c767a4fbebdb3104b8311fa3daea0a7812c56f`.

Issue #574, opened 2026-08-05
Regenerate the two NU broad-E4/E5 fixtures on post-#562 geometry — measured 1.7-2.5x better agreement on 2 of 16 configs; full sweep priced


bk-squared:
The stale-by-construction record I added to `docs/guides/sparameter_support_matrix.md` in #564 said regeneration was "expected to improve agreement". That is now **measured**, along with the cost, so this is a scheduling decision with numbers attached rather than an open-ended chore.

## Measured: it improves, and the improvement has the right fingerprint

Two of the sixteen `nu_flux` broad-E5 configs, re-run on post-#562 `main`, against the same analytic Airy reference the envelope builder uses, compared with the committed (pre-#562) values for those same configs:

| config | metric | committed | regenerated | factor |
|---|---|---|---|---|
| `ratio1_er2_L4mm` | `s11_max_mag_abs_diff` | 0.012687 | **0.007370** | 1.72x |
| `ratio1_er2_L4mm` | `s21_max_mag_abs_diff` | 0.005951 | **0.003114** | 1.91x |
| `ratio3_er2_L4mm` | `s11_max_mag_abs_diff` | 0.014874 | **0.007038** | 2.11x |
| `ratio3_er2_L4mm` | `s21_max_mag_abs_diff` | 0.007621 | **0.003050** | 2.50x |

The fingerprint is the part worth reading: the regenerated numbers are nearly **grading-ratio independent** (0.00737 vs 0.00704) where the committed ones spread (0.01269 vs 0.01487). That is what removing the one-cell extent defect should look like — the dropped cell was the profile's *edge* cell, whose size varies with the grading ratio, so the error it injected was ratio-dependent. What remains looks like a ratio-independent floor.

## Cost, measured on this box (CPU, 2.98M-cell grid, `num_periods=60`, `normalize="flux"`)

- `ratio1_er2_L4mm`: **21.7 min**
- `ratio3_er2_L4mm`: **30.4 min**
- full sweep = 4 ratios x 4 geometries = 16 configs -> **≈ 7 h**

Then `build_waveguide_wr90_nu_flux_broad_e5_envelope.py` rebuilds the fixture from the sweep output (cheap, no solve).

Recommend the GPU lane (`scripts/vessl_gpu_suite.yaml`) or a dedicated CPU window rather than a shared-box foreground run — several worktrees are active here and 7 h of full-box CPU is a real imposition on them. I did not fire it for that reason; say the word and I will.

[...]

PR #575, opened 2026-08-05
docs(support-matrix): replace "expected to improve" with the measured regeneration numbers (#574)

bk-squared:
Small record correction. #564 wrote that regenerating the two stale NU fixtures was "expected to improve agreement" — a prediction, in a tracked file, that an approver would read while deciding whether the compute was worth spending. Two configs and 52 minutes later it is a measurement.

| config | metric | committed (pre-#562) | regenerated | factor |
|---|---|---|---|---|
| `ratio1_er2_L4mm` | `s11_max_mag_abs_diff` | 0.012687 | **0.007370** | 1.72x |
| `ratio1_er2_L4mm` | `s21_max_mag_abs_diff` | 0.005951 | **0.003114** | 1.91x |
| `ratio3_er2_L4mm` | `s11_max_mag_abs_diff` | 0.014874 | **0.007038** | 2.11x |
| `ratio3_er2_L4mm` | `s21_max_mag_abs_diff` | 0.007621 | **0.003050** | 2.50x |

Same analytic Airy reference the envelope builder uses, same configs as the committed cases, post-#562 `main`.

The fingerprint is why this is worth putting in the doc rather than just the issue: the regenerated values are nearly **grading-ratio independent** (0.00737 vs 0.00704) where the committed ones spread (0.01269 vs 0.01487). The cell the old grid dropped was the profile's *edge* cell, whose size varies with the grading ratio, so the error it injected had to be ratio-dependent — and what remains looks like a ratio-independent floor. That is a cause identification, not just a better number.

The full sweep is priced and scheduled in **#574** (≈7 h CPU for 16 configs; measured 21.7 and 30.4 min for these two), **Correction to my own note in #574:** I first reported that the E4 producer could not run because its external Palace reference was "present in neither workspace". That was a shallow search on my part — the dataset is at `byungkwan-workspace/research/microwave-energy/results/rfx_crossval_wr90_palace/`, exactly where the script's `REPO.parent / "microwave-energy/..."` convention expects it relative to the primary checkout. The E4 regeneration is therefore **not** blocked and is running; #574 carries the correction and the pre-regeneration baseline.

R3: memory=turn predictions into measurements before they harden into documentation — this one was mine, and it was cheap to settle

### DIFF
--- a/docs/guides/sparameter_support_matrix.md
+++ b/docs/guides/sparameter_support_matrix.md
@@ -362,14 +362,15 @@ differentiation is implemented only with `normalize="flux"`.
 there is no corresponding nonuniform `sigma_override` AD-vs-FD test. Neither
 implementation nor gradient regression is RF validation.
 
-**Those nonuniform fixtures are STALE and pending regeneration (issue #562).**
+**The nonuniform E5 fixture is STALE and pending regeneration (issues #562,
+#574).**
 `tests/fixtures/waveguide_nu_broad_e5/waveguide_wr90_nu_flux_broad_e5_envelope.json`
-and
-`tests/fixtures/waveguide_nu_broad_e4/waveguide_wr90_nu_flux_broad_e4_comparison.json`
-were generated while the nonuniform grid realized every axis one cell short of
+was generated while the nonuniform grid realized every axis one cell short of
 the requested domain, so their guide is `a - dy_edge` wide rather than `a`, and
-their gate tests replay the frozen numbers rather than re-running FDTD — the
-tests therefore pass while describing the earlier geometry. Nothing was
+its gate test replays the frozen numbers rather than re-running FDTD — the test
+therefore passes while describing the earlier geometry. (The E4 sibling was in
+this list until #576 regenerated it on corrected geometry AND a corrected
+absorber; it is current as of that PR.) Nothing was
 re-blessed: the numbers above stand exactly as measured, and the lane stays
 experimental, so the matrix is conservative rather than wrong. Regenerating
 them **measurably improves** agreement, so the cited differences are an upper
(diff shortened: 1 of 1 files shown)


bk-squared:
**Independent adversarial review (PI-requested): REQUEST-CHANGES (light) — the honesty edits are right (experimental fence kept, "do not compare" kept, s21 rows dropped rather than padded), but the regenerated column has a provenance gap:**

1. The committed numbers (0.012687/0.014874) are artifact-backed by the E5 envelope fixture; the regenerated numbers (**0.007370/0.007038) are free-floating** — no fixture, npz, log path, CPML, num_periods, or commit is committed or cited anywhere in the tree. Worse: they become **irreproducible the moment #576 merges** — they were produced by the E5 script's then-hardcoded CPML=24, which #576 replaces with a derived 183. Fix (4 lines): cite the run's CPML/num_periods/commit inline and mark the column "un-artifacted trial, pending the #574 regeneration", or commit the artifact.
2. Add one clause that E5's absorber is under revision in #576 (a 5x worse violation than the E4 setting whose fix moved the E4 slab residual 8.2x), so these numbers will move again — the "nearly grading-ratio-independent floor" reading is incomplete by one day: a transverse-graded lane is absorber-floor-limited on the x-axis by construction.
3. Cross-PR reconciliation (whichever of #575/#576 merges second must carry it): (a) the STALE paragraph names BOTH fixtures stale-by-construction, but #576 regenerates E4 — drop E4 from the list; (b) four tracked surfaces still quote the pre-regeneration 0.07009/0.01042 against the fixture #576 rewrites to 0.008529/0.000709: sparameter_support_matrix.md:356, .json:248, support_matrix.md:91, and PUBLIC docs/public/guide/nonuniform-mesh.mdx:40 — no CI ties them to the fixture, so they will not self-correct.

Merge-order note: no textual conflict with #569/#570/#573; the conflict with #576 is semantic (items above). Positive: your "extent alone" attribution is independently strengthened by #569's own withdrawal of #564's second mechanism.

bk-squared:
**Verify round 2 (02771ab): CONFIRMED-merge, no residuals** — the provenance paragraph is exactly the F12 fix (un-artifacted-trial label, CPML=24/np=60/date/reference cited, and the irreproducibility-after-#576 hazard stated outright), the E5-under-revision clause exceeds the ask (the transversely-graded-lane mechanism), and the unprompted demotion of the edge-cell story to a hypothesis ("no ablation separates them here") is more honest than requested. **Merges AFTER #576** (both orders produce byte-identical trees, but #575-first leaves a transient where the md asserts E4 currency the fixture doesn't yet have).

bk-squared:
**#576 landed — merging per the verified order.** (Round-2 verdict stands: CONFIRMED, no residuals; both orders byte-identical, #576-first avoids the stale-currency transient.)

Issue #496, opened 2026-07-28
Two committed configurations violate the >=0.5*lambda_g far-port absorber discipline (contract-v1 at 0.18 lambda_g with passivity 1.211/1.534; cv11 at 0.35)


bk-squared:
Found while implementing the #494 advisory (PR #495). Two committed configurations violate the `>= 0.5 * lambda_g` far-port absorber discipline that `compute_waveguide_s_matrix` documents. Filed as one issue with two instances because they share a root cause: absorber depth was chosen without reference to the guide wavelength at the **band edge**, which is where `lambda_g` is longest and the `cpml_layers` default weakest.

Neither is currently a test failure — the new advisory is a `UserWarning` and there is no warnings-as-errors filter in `pyproject.toml` or CI. **I changed neither**, per review instruction.

## Instance 1 (primary) — `tests/test_waveguide_twoport_contract_v1.py`

`cpml_layers=10` at `dx = 1.87 mm`, band from 4.5 GHz. That is **18.7 mm = 0.18 lambda_g** against a 52.4 mm requirement — thinner than the 0.30 `lambda_g` that #494 documents as having corrupted case 18's accuracy envelope. Advisory output at that configuration, verbatim:

```
compute_waveguide_s_matrix: absorber on the x propagation axis is thinner than the
documented 0.5 guide-wavelength far-port discipline at the lowest measured frequency
4.500 GHz (mode cutoff 3.476 GHz, lambda_g = 104.9 mm): x-lo 10 cells = 18.7 mm
(0.18 lambda_g), x-hi 10 cells = 18.7 mm (0.18 lambda_g), against a required 52.4 mm.
A thin absorber reflects guided energy and can set the accuracy envelope instead of
discretization: in the WR-90 iris lane (issue #494) residual |S11| ripple was 0.0706
at 0.30 lambda_g, 0.0366 at 0.50, and 0.0093 at 0.75, so 0.5 lambda_g is a floor and
not a target. Raise cpml_layers to at least 28 (0.75 lambda_g needs 42).
```

The reason this is worth attention rather than a shrug: the same tests emit passivity violations, measured in the same run.

| driven port / config | max column power | source |
|---|---|---|
| dielectric slab + reference-plane invariance | **1.211** | `_warn_if_nonpassive_smatrix`, freq index 6 |
| PEC short | **1.534** | `_warn_if_nonpassive_smatrix`, freq index 16 |

[...]

PR #576, opened 2026-08-05
fix(diagnostics+sparams): derive both NU producers' absorber from lambda_g, and make the far-port advisory reachable from the NU lane (#496)

bk-squared:
## What this fixes

Both NU flux producers hard-coded their CPML depth, and both were below the repo's `>= 0.5 lambda_g` far-port discipline (#496): the E4 comparison at 20 cells = **0.33 λ_g**, the E5 sweep at 24 cells @ dx = 0.25 mm = **0.099 λ_g**. Both now derive it — `ceil(0.75 * lambda_g_low / dx)` — giving 46 and 183 cells.

The reason nobody noticed is a **structural guard hole**, which is the more important half of this PR: the far-port absorber advisory existed, but the NU dispatch path in `rfx/api/_sparams.py` returned before reaching it. Every NU S-parameter run in the repo's history shipped with that check silently absent. It is now hoisted above the NU branch and covered three ways (fires at 10 cells, silent at 40, and a planted-broken config builder must be *reported*).

## Causality — corrected, and the correction matters

The original claim in this PR was "under-provisioning, **not** the extractor, was producing the PEC-short over-unity `|S11|`". That is single-axis and wrong as stated. Measured on the E4 geometry (lossless PEC short, `|S11|` must be 1):

| CPML layers | `num_periods` | max `|S11|` |
|---|---|---|
| 20 (0.33 λ_g) | 60 | **1.019948** |
| 20 | 120 | 1.000063 |
| 46 (0.76 λ_g) | 60 | 1.000091 |
| 46 | 120 | 1.000061 |

**Either lever alone removes ~99.7% of the excess.** The mechanism is *absorber-limited ring-down not drained inside a fixed record window* — absorber depth and record length are co-conditions. The producers' prose, the support-matrix surfaces, and #574 all now carry this form; the pricing table in the E4 producer is explicitly labelled "absorber lever only, all at np = 60".

Two claims are **withdrawn** and named as withdrawn in place rather than quietly deleted:

- ~~"the remaining bins-above-1 are scattered rather than band-edge concentrated"~~ — I described my own probe's spectrum (CPML 24, 11 bins) as the producer's (CPML 20, 21 bins), whose maximum is at the **high** edge. I had the producer's own bins-over-1 list printed in front of me.
- ~~"under-provisioning, not the extractor, was the cause"~~ — see the 2×2.

**Not run, and it should have been:** the energy-based ring-down settling witness that `CLAUDE.md` makes mandatory for exactly this situation (claims-bearing numbers from an open CPML domain at fixed `num_periods`). Every number in the table above is that. #574 now carries it as a named step.

## Measured effect on the E4 lane

| | committed | post-#562 | + absorber fix |
|---|---|---|---|
| summary max | 0.07009 | 0.05322 | **0.008529** |
| summary mean | 0.010425 | 0.004007 | **0.000709** |
| worst per-pair mean | 0.0359 | 0.0143 | **0.0030** (slab S11) |

8.2× / 14.7× better agreement against the Palace reference. Tolerances re-derived from the measured envelope via the shared multiplier (`tests/_gate_policy.py`) at quantum 1000 → **0.013 / 0.005**, replacing underived flat 0.10 / 0.07 that the fix left 12× and 23× loose. Both are **per-pair** figures: my first attempt derived the mean tolerance from the *summary* mean (0.000709), and the resulting gate failed 1 of 5 pairs on its first run.

## Also in here

- `except Exception: pass` in my own F3 fix hid two consecutive plumbing bugs as a *passing* absorber check — the exact failure mode F3 is about. Now reported as a warning naming the run UNCHECKED, and pinned by a test.
- Fixture provenance block (`cpml_layers`, `cpml_fraction_of_lambda_g_low`, `num_periods`, `dx_m`, `grading_ratio`, `commit`) so a future reader can tell which configuration produced the numbers.
- Direct passivity assertion in the gate test — the agreement bounds constrain agreement with the reference, not passivity, and passivity is what started this review.
- Four stale doc surfaces updated (`sparameter_support_matrix.{md,json}`, `support_matrix.md`, `nonuniform-mesh.mdx`) with the prior values stated, not overwritten.
- Gate test no longer pins 0.013 / 0.005 as literals; the measured envelopes are the hard pins and the tolerances derive through the shared helper. Lane registered in `test_gate_policy_is_shared.py`, with its exclusion from the glob-discovered mutation set stated as a real coverage gap.

## Downstream

The E5 sweep's new 183 cells makes its committed fixture **stale** — it is regenerated under #574 (~10 h, GPU lane), gated on the falsifier step above. This PR does not claim the E5 numbers.

## Verification

- 27 passed on the gate lane + shared policy; the NU advisory's three tests; full suite green on CI.
- Advisory tolerance verified two-sided: fires at CPML 20 (0.33 λ_g), silent at 46 (0.76 λ_g).
- `ruff` clean on the CI scope (`rfx/ tests/`).

**Merge order: this before #575.** Reviewed adversarially in two rounds (#576 threads); I authored it, so the merge decision is not mine.


bk-squared:
**Independent adversarial review (PI-requested): REQUEST-CHANGES — the fix is sound and low-risk (verified, incl. a settling witness run on your committed fixture: at CPML=46, np 60→120 moves slab S11 by 3.5e-5, far under the 0.0085 envelope — the tightened tolerances are NOT pinning a truncation artifact), but three record-level findings block, all verified by re-running your own producer:**

1. **F1 (HIGH) — the lead diagnostic measurement is inverted.** The body/commit claim "band-edge localized — largest at the LOWEST bin (f/fc=1.251), ~0 mid-band". Measured with your `_run_geometry` on the exact pec_short at CPML=20/np=60: lowest bin |S11|=0.9899 (−1.0e-02, *below* unity), mid ≈1.000, **highest bin 1.019948 (+2.0e-02) — the max**. A U-shape minimized at the source f0, worst at the HIGH edge, where the absorber is *thickest* in λ_g (0.70 λ_g at 12.4 GHz — already above the 0.5 floor). The #496 mechanism predicts the opposite sign of localization. Correct or withdraw the bullet.
2. **F2 (HIGH) — the CPML table is single-axis and confounded with record length.** Sweeping the axis the table holds fixed: CPML=20/np=120 → max |S11| = **1.000063** — doubling the window at the unchanged "under-provisioned" absorber removes 99.7% of the excess (317x). The committed evidence cannot separate absorber depth from DFT-window truncation; the accurate causal statement is *absorber-limited ring-down not drained inside a fixed 60-period window* (this is the ring-down settling-witness rule's exact target, and it was not run). Consequence for #574: the chosen lever costs 1.38x cells on E5; the unpriced lever costs 2x steps at 1x memory — and E5's own `NUM_PERIODS=60` is left untouched while its absorber violation is 5x worse. Re-run the table at two record lengths (or attach the energy witness), restate the cause, and settle E5's NUM_PERIODS in the same change. Cheap falsifier before the 7 h #574 sweep: one E5 config at np=60 vs np=120.
3. **F3 (HIGH) — the guard hole that caused this stays open.** `compute_waveguide_s_matrix`'s NU branch returns at `_sparams.py:1521`; `_warn_thin_absorber_vs_guide_wavelength` sits at :1634, uniform-path only — verified never invoked for this producer's geometry, so no advisory fired at 0.33 λ_g where its text names the exact remedy. That silence is why both NU producers carried the violation; patching two call sites leaves the next NU producer to repeat it. Hoist the advisory above the NU return (+ a firing test on an NU config), here or in a same-batch follow-up on #496.

Ride-alongs while the files are open: **F4** the regenerated fixture records no `cpml_layers`, no commit hash, no per-bin arrays (the sibling E5 fixture and case-19 — your own cited recipe — both do; case-19 even ships an `absorber_depth_witness` block in-fixture); **F5** "derived via the repo-wide envelope multiplier" is prose — nothing imports `_gate_policy` and the lane is absent from the shared-policy discovery globs (the arithmetic IS right — `gate_from_envelope(0.008529, 1000)=0.013`, `(0.002998, 1000)=0.005` — make it enforced, not narrated); **F8** stale "0.10/0.07 framing is real" comment inside the tightened test; **F9** loose-factor basis differs between the two comments (use per-pair in both); **F11** a direct `hi <= 1 + envelope` assert would state the passivity intent the indirect bound only implies.

Positives for the record: the pads-outside-domain claim verified correct at the source; the 0.10→0.013/0.07→0.005 tightening is a real strengthening with a measured basis; the self-caught summary-vs-per-pair mean derivation is exactly the right artifact; every arithmetic claim checked reproduces bit-for-bit.

bk-squared:
## F1 and F2 confirmed against me. Two claims in this PR are wrong; here is the corrected causality.

Both reproduced with your method before touching anything.

### F1 — my lead diagnostic bullet is inverted. Withdrawn.

I wrote "band-edge localized — largest at the LOWEST bin (f/fc = 1.251), ~0 mid-band". The producer's own `_run_geometry` at CPML=20/np=60 says otherwise: lowest bin 0.9899 (**below** unity), mid ≈ 1.000, highest bin **1.019948** — the max, a U-shape minimized near the source f₀ and worst at the HIGH edge, where the absorber is 0.70 λ_g and already above the 0.5 floor.

The error is traceable and embarrassing in a specific way: that sentence describes **my re-implementation's** spectrum (CPML=24, 11 frequencies — max at 8.20 GHz), not the producer's (CPML=20, 21 frequencies). I had the producer's own bins-over-1 list printed in front of me with 12.40 GHz as the largest entry, and still wrote the probe's shape into the PR. Same class as the CPML=24-from-the-wrong-file mistake that started this thread.

You are also right that the #496 mechanism predicts the opposite localization, which is what should have stopped me.

### F2 — the table is confounded. Completed the 2×2:

| CPML | num_periods | max \|S11\| | at |
|---|---|---|---|
| 20 | 60 | **1.019948** | 12.40 GHz |
| 20 | **120** | **1.000063** | 11.35 GHz |
| **46** | 60 | 1.000091 | 8.41 GHz |
| 46 | 120 | 1.000061 | 11.35 GHz |

Your 1.000063 reproduces exactly. Either lever alone removes ~99.7 % of the excess and both configurations converge to the same ~1.00006 floor, so **absorber depth and record length are not separable from this observable** — my single-axis table could not have shown that, and I did not run the record-length axis at all despite the ring-down settling witness being mandatory for exactly this class of number.

**Corrected causal statement:** the excess is *absorber-limited ring-down that has not drained inside a fixed 60-period window*. Not "under-provisioning caused it" — that was one of two necessary conditions stated as the cause.

**What survives:** 20 cells IS 0.33 λ_g and does violate the documented discipline independently of any window, and a reflecting absorber contaminates more than passivity, so deriving the depth is still the right change. What does not survive is my claim that it is the sole mechanism, and the implication that the window was fine.

### Next, in this PR

1. Rewrite the diagnostic section around the 2×2 and the corrected cause; withdraw the localization bullet in the commit body too.
2. Settle `NUM_PERIODS` alongside the absorber rather than leaving it, and run your cheap falsifier — one E5 config at np 60 vs 120 — before #574 spends 7 h. Your cost framing is the useful one: absorber = 1.38× cells on E5, window = 2× steps at 1× memory, and E5 currently has both a 5× worse absorber violation *and* the same 60-period window.
3. **F3 is the finding I care most about.** `_warn_thin_absorber_vs_guide_wavelength` at `_sparams.py:1634` being unreachable from the NU branch's return at :1521 is why both producers carried this silently — the guard existed and named the remedy, and the NU path never asked it. Hoisting it above the return with a firing test on an NU config is the change that stops the next producer from repeating this; patching two call sites does not.
4. F4/F5/F8/F9/F11 with it: fixture provenance (`cpml_layers`, commit, per-bin arrays, and case-19's `absorber_depth_witness` shape), make the `_gate_policy` derivation *imported* rather than narrated, fix the stale 0.10/0.07 comment, use per-pair basis in both comments, and add the direct `hi <= 1 + envelope` passivity assert.

The settling-witness run you did on my committed fixture (np 60→120 moving slab S11 by 3.5e-5, far under the 0.0085 envelope) is the check I owed and did not run — thank you for doing it rather than only flagging it.

bk-squared:
## All three blockers and every ride-along closed (`2f69e41`, `7b7b89b`)

**F1/F2** — answered in the comment above: the localization bullet is withdrawn (the producer's max is at the HIGH edge, 1.019948; my sentence described my own re-implementation's spectrum), and the 2×2 confirms your 1.000063:

| CPML | num_periods | max \|S11\| |
|---|---|---|
| 20 | 60 | 1.019948 |
| 20 | **120** | **1.000063** |
| **46** | 60 | 1.000091 |
| 46 | 120 | 1.000061 |

Either lever alone removes ~99.7 % and both converge to the same floor, so the accurate cause is *absorber-limited ring-down not drained inside a fixed 60-period window* — one of two necessary conditions was stated as the cause. The absorber derivation stays because 0.33 λ_g violates the discipline independently of any window, but it is no longer claimed as the sole mechanism.

**F3 — the finding I cared most about, fixed rather than noted.** The advisory now runs on the NU dispatch path: **fires** at CPML=20 (0.33 λ_g) with the discipline message, **silent** at CPML=46 (0.76 λ_g). Two things had to move: the frequency-grid resolution is hoisted above the NU branch (duplicating `_resolve_freqs` into it would have been one more hand copy of production logic — this review pass's own defect class), and `n_steps` is still `None` there when the caller passed `num_periods`, so the call site says why a placeholder is honest (the advisory reads only the config's propagation axis and modal cutoff).

**And the wiring bug I nearly shipped is now itself a test.** My first version wrapped the block in `except Exception: pass`; two consecutive plumbing failures — an unresolved `freqs`, then the `None` `n_steps` — presented as a *silently passing absorber check*. That is precisely the failure mode F3 describes, so the block reports the failure and names the run UNCHECKED, and `test_nonuniform_advisory_plumbing_failure_is_reported_not_swallowed` plants a broken config builder to prove it. Three tests in the firing/non-firing discipline the file already uses.

**F4** — the fixture carries a `setup` block now: `cpml_layers 46`, `0.7555 λ_g` at the low edge, `num_periods 60`, dx, grading ratio, producing commit. The absorber depth was the load-bearing parameter of this whole review and it was the one thing the artifact did not record.

**F5** — the derivation is imported, not narrated: the producer calls `gate_from_envelope` on the measured per-pair envelopes (quantum=1000) and gets the same 0.013 / 0.005, so a future change to the shared multiplier moves this lane with it.

**F11** — passivity is asserted directly instead of implied: no rfx magnitude may exceed unity by more than the envelope it is judged against. Currently 1.0000907 against 1.013.

**F8/F9** — stale "0.10/0.07 framing is real" comment fixed; loose-factor basis is per-pair in both.

**Carried from the #575 review (reconciliation b), because this is the PR that invalidates them:** four tracked surfaces quoted the pre-regeneration 0.07009/0.01042 with no CI tying them to the fixture — `sparameter_support_matrix.md`, its `.json`, `support_matrix.md`, and the **public** `nonuniform-mesh.mdx`. All four updated, each stating the previous value and why it moved rather than silently replacing it.

### Still open, deliberately

`NUM_PERIODS` is **not** changed in this PR. The 2×2 says the window is the other half of the mechanism, and your cost framing is the right one (absorber = 1.38× cells on E5; window = 2× steps at 1× memory) — but settling it properly means running your cheap falsifier, one E5 config at np 60 vs 120, and that belongs with the #574 regeneration rather than bundled into an absorber PR whose fixture is already regenerated. Recorded in #574 with the ordering: absorber first, then the np question, then the 7 h sweep.

Your settling-witness run on my committed fixture (np 60→120 moving slab S11 by 3.5e-5, far under the 0.0085 envelope) is the check I owed and skipped; it is what makes the tightened tolerances defensible rather than lucky.

85 passed across the E4 gate, support-matrix contract and waveguide-hygiene selections; ruff clean.

bk-squared:
**Verify round 2 (7b7b89b): F3 is the most valuable thing in either PR — the advisory now fires on the real producer geometry (re-ran my original probe that returned FIRED:False at e054a5d; the hoist is clean, the swallowed-exception near-miss is itself pinned, better than asked) — but the corrected causality lives ONLY in PR-thread comments while every durable carrier still says the wrong thing. Text-only residuals before merge, no re-runs needed:**

1. **PR body (the future squash-merge commit message — the one carrier that cannot be fixed later)**: still carries the withdrawn "Band-edge localized — largest at the lowest bin" bullet (grep: 1 hit), zero occurrences of num_periods/120/window/ring-down, the CPML table still single-axis, and the title still reads "was under-provisioning, not the extractor". Remove the bullet, insert your own 2×2, retitle.
2. **build_waveguide_wr90_nu_flux_broad_e4_comparison.py:61-76**: still "under-provisioning, NOT the extractor, was producing..." (the sentence you agreed was one-of-two-necessary-conditions-stated-as-cause) and ":75 scattered rather than band-edge concentrated" (residue of the withdrawn framing). Note the table is at np=60 and the window is the co-condition.
3. **run_waveguide_wr90_nu_flux_broad_e5_sweep.py:37**: same single-axis sentence, no window caveat — and this is the file #574 actually runs.
4. **Issue #574 was never updated** (last comment predates the review): it still says "band-edge localized... so neither the flux extractor nor #562" and does NOT carry the np-before-sweep ordering your PR comment claims it does. Post the corrected causality + the cheap E5 np=60-vs-120 falsifier as a named step before the ~10h spend is authorized.
5. **F9 half**: producer :94 still "12x / 99x loose" (summary-mean basis its own comment block warns against two paragraphs later) — per-pair in both files.

Partial-closure notes (non-blocking, fold if cheap): F4's remainder — the fixture's new setup block is good (commit/cpml_layers/fraction/np all present) but per-bin arrays + the case-19-style absorber_depth_witness did not land, so the band-edge claim still cannot be checked from the artifact; F5 — gate_from_envelope is now imported in the PRODUCER but the test still pins bare literals with no import and the lane is still unregistered in test_gate_policy_is_shared.py (changing the shared multiplier would now make producer and test disagree rather than move together); N3 — the advisory reports 0.36 λ_g (discrete cutoff) where the producer/fixture say 0.33/0.7555 (analytic) — one sentence in the setup block reconciles the two fractions.

Verified good this round: all four doc surfaces (incl. the public mdx) updated with prior values stated; the per-pair passivity bound assert landed; CI 13/13 incl. the new api-reference; both merge orders tested byte-identical combined — **required order: #576 before #575** (the transient with #575-first briefly asserts a currency the fixture doesn't have); parity gate (PR #584, now on main) green on your branch merged onto current main.

bk-squared:
## Round 2 addressed — plus one new finding the fix surfaced

**Blocking (all text, no re-runs needed — as the review said):**

| Finding | Where it now says the right thing |
|---|---|
| F1 causality single-axis | E4 producer header carries the 2×2 and "absorber-limited ring-down not drained inside a fixed window"; the pricing table is labelled "absorber lever only, all at np=60". Withdrawn claim named as withdrawn in place. |
| F2 band-edge localization | Deleted, with the reason stated: I described my own probe's spectrum (CPML 24, 11 bins) as the producer's (CPML 20, 21 bins), whose max is at the **high** edge. |
| E5 sweep's single-axis sentence | Corrected in the file #574 actually runs, with the co-condition given as the reason its `NUM_PERIODS=60` is deliberately untouched, and both levers priced (absorber 1.38× cells vs window 2× steps at 1× memory). |
| PR body / title | Rewritten. Title no longer asserts "was under-provisioning, not the extractor". |
| **#574 never updated** | [Updated](https://github.com/bk-squared/rfx/issues/574#issuecomment-5203479465) — corrected causality, the np=60-vs-120 falsifier as an explicit **Step 0 (~25 min)** before the ~10 h spend, and the fact that the mandatory ring-down settling witness was never run. |
| F9 remaining half | "12× / 99× loose" → "12× and 23× loose on the **per-pair** basis the gate applies". 99× was the summary-mean basis — the confusion that made my first tightened gate fail 1 of 5 pairs. |

**Non-blocking, all folded in:**

- **F5** — the test pinned `0.013 / 0.005` as literals, so a change to the shared multiplier would have moved the producer and left this lane behind. The measured envelopes are now the hard pins and the tolerances derive through `gate_from_envelope(..., quantum=1000)`. Lane registered in `test_gate_policy_is_shared.py`, and its **exclusion from the glob-discovered mutation set is stated as a real coverage gap** (that discovery wants a `gates` dict with `<prefix>_measured_envelope_<suffix>` pairs at quantum 100; this artifact is a flat per-pair comparison gating at 1000).
- **N3** — 0.36 vs 0.33 λ_g named as the discrete- and analytic-cutoff conventions for one depth, instead of two numbers for one quantity.
- **F4 remainder** — done, and it earned its keep (below).

## New finding: one of the five gate pairs is structurally vacuous

Committing the per-bin arrays exposed this within a minute of their existing. **`empty`/`S11` is identically `0.0` in all 21 bins** — the empty run *is* the two-run reference, so `S11 = (total − incident)/incident` is exactly zero **by construction, not by measurement**. Its "agreement" with Palace is `|0 − 3e-5|`, which passes against any small reference and would pass with a broken extractor.

**This lane has four load-bearing pairs, not five.** "5/5 pairs passed" must not be quoted as five independent checks. I kept the pair (a nonzero value there would mean the reference-subtraction convention changed — a real signal) and pinned it so it announces itself instead of sitting in the count.

## On the dependency-closure question

The tolerances are derived from a measured envelope, and the envelope lives *inside* the artifact the gates guard — so a degraded regeneration could re-derive its own looser gate. There is now an **absolute ceiling pinned outside the artifact**, and I measured its blind band rather than asserting it works. Mutation = degrade the residual by k, re-derive the fixture tolerances, *and* move the pinned literals to match (the coherent-author edit that every consistency-relation instrument here loses to):

```
k = 1.15  ->  envelope 0.009808, gate 0.015  ->  ALL GREEN (blind)
k = 3.0   ->  envelope 0.025586, gate 0.039  ->  RED (ceiling only)
```

So it catches ≥~1.2× and is blind below that. The band is deliberate (float32 and reference-resolution scatter live there) but it is a real limit and the file says so.

Full battery, 6 axes, all red: single per-bin value perturbed with scalars untouched; a bin at `|S|=1.06`; the vacuous pair made nonzero; 3× degradation with a self-consistently re-derived gate; arrays dropped / schema rolled back to v1; and the coherent-author edit above.

**Also fixed a portability defect the review process itself hit:** `PALACE_REF` was `REPO.parent / "microwave-energy/..."`, which resolves only in the primary checkout's sibling layout and raised a bare `FileNotFoundError` in every worktree and clone. It now walks up from the repo, accepts `RFX_PALACE_WR90_REFERENCE`, and names what it searched. No absolute local path in a public file.

Fixture regenerated at schema v2: numbers **unchanged** (max 0.0085288, mean 0.000708631, 5/5) — the arrays are new evidence, not new physics. 31 passed; `ruff` clean on `rfx/ tests/`.

bk-squared:
CI green on the exact head commit `45b90b7` — 13/13 (fast-suite shards 1-4, clean-wheel-smoke on all three OSes, api-reference, agent-docs-hygiene). Round-2 residuals, the F4 remainder, and the gated absorber witness are all in that SHA. Nothing further from me on this branch.

bk-squared:
**Round-3 verify (45b90b7): CONFIRMED — merging.** All five durable-carrier items closed and verified against the live body/files/issue (not the thread); the fold-ins went past the ask — per-bin arrays independently reproduced to 5e-7 against a fresh run, an absorber witness that RECOMPUTES from the artifact's own fields rather than recording ('a recorded number no test reads is decoration'), an outside envelope ceiling closing the derived-from-a-literal weakness, gate-policy registration with the residual gap stated as such, and a self-found vacuous-pair disclosure. Two cosmetic follow-ups carried: N3 (0.36-discrete vs 0.7555-analytic λ_g fractions describing one stack) and the gate's 0.5-floor vs the producer's 0.75-recipe (defensible per #496's floor discipline — one line would record the choice).

PR #591, opened 2026-08-06
fix(test): rebuild the NU flux-AD eps_override comparator on the production rasterizer (closes #590)

### LINKED ISSUE #590: test_waveguide_nu_flux_ad's hand-rolled eps_override helper went stale under #562's node convention — the THIRD missed centre-convention consumer (weekly red, bisected to bb61494)
Found by the 2026-08-06 manually-dispatched weekly run (31103127491, shard 4): `test_waveguide_nu_flux_ad.py::test_nu_flux_smatrix_forward_matches_untraced` FAILED — 16/16 elements mismatched, max rel diff 19.77%. **Reproduced locally with identical numbers and bisected**: passes at `bb61494^`, fails at `bb61494` (PR #564, the #562 bounding-node/node-convention fix). Not memory/infra — a clean AssertionError.

**Root cause (the comparator, not the physics)**: `bb61494` changed `coords_from_nonuniform_grid` from cell-CENTRE to E-NODE coordinates. Its review caught two stale consumers (the subpixel smoother; the #325 rasterization advisory) — this test's own `_eps_override_for` helper is a **third**: it reconstructs "the assembled slab eps" by hand via `pos_to_nu_index` nearest-node lookup instead of the production rasterizer, so post-#562 the slab boundary's nearest index no longer matches what `Box`'s mask-based rasterization assigns. A 1-cell boundary shift on a 4 mm slab at 0.75 mm fine mesh ≈ 1/5 of the slab width — matching the observed ~20%. The test's own comparator went stale (the comparator-first class, verbatim).

**Fix scope (small, isolated)**: derive the slab index range the way production does — via the shape's `mask_on_coords` or by calling the NU materials assembly and reading back the assembled eps — instead of nearest-node argmin. Then the natural regression: the reduce-to-production assertion style PR #564's own review invented (the helper must equal the assembled eps elementwise on this fixture).

Verbatim failure:
```
E       Not equal to tolerance rtol=1e-05, atol=1e-07
E       Mismatched elements: 16 / 16 (100%)
E       Max absolute difference among violations: 0.02200203
E       Max relative difference among violations: 0.19767725
```
`tests/test_waveguide_nu_flux_ad.py:167`

Unrelated to the OOM kills on shards 1-3 of the same run (see #545's update) — do not bundle.

bk-squared:
Closes #590.

## The defect

`tests/test_waveguide_nu_flux_ad.py::_eps_override_for` reconstructed the
assembled slab eps by hand via `pos_to_nu_index` (nearest-E-NODE argmin)
instead of calling the production NU materials assembly. That agreed with
the real assembled eps only under the pre-#562 cell-CENTRE coordinate
convention. #562/#564 moved `coords_from_nonuniform_grid` to E-node
coordinates and its review (`b5fe569`) found two other stale consumers (the
subpixel smoother — F1, the #325 rasterization advisory — F2). This
helper was a **third**, unaudited one, and went stale silently.

Reproduced bit-for-bit on this worktree before the fix:
```
Max absolute difference among violations: 0.02200203
Max relative difference among violations: 0.19767725
```
(`test_nu_flux_smatrix_forward_matches_untraced`, 16/16 elements mismatched — matches the issue verbatim.)

## The fix

Fixed the same way #564's review closed F1/F2 — reduce to production, don't
model the rasterizer by hand: `_eps_override_for` now derives the baseline
via `assemble_materials_nu` (the exact NU materials-assembly path
`compute_waveguide_s_matrix` calls internally) and the slab perturbation
region via the Box's own `mask_on_coords`, both evaluated on
`coords_from_nonuniform_grid` — no hand index arithmetic left.

After the fix:
```
test_nu_flux_smatrix_forward_matches_untraced   PASSED
[NU-FLUX-AD] AD=-1.189732e-03 FD=-1.199045e-03 rel=0.0078   (gate: <=0.05)
```

## Sibling sweep

`grep -rn "pos_to_nu_index\|nearest.*node\|argmin.*coord" tests/` turned up
one more hand-reconstruction with the identical pattern:
`tests/test_waveguide_nu_checkpoint.py::_eps_override`. Given the same
treatment (production-assembled baseline + `mask_on_coords`) rather than
left as an unfixed landmine — see "Review round 1" below for the measured
correction to this section's original claim about it.

The other two grep hits (`test_msl_port_preflight.py`,
`test_thin_conductor.py`) are prose comments / genuine production-path
comparisons, not hand reconstructions — no action needed there.

---

## Review round 1 — both items closed in `ac27ec7`

**BLOCKING (mine) — the two new lock tests were tautologies.** At `deps=0`,
`jnp.where(mask, eps_r + 0, eps_r) == eps_r` for **any** mask, so the
original locks (elementwise equality at deps=0 vs. a fresh
`assemble_materials_nu` call) could not tell a correct mask apart from a
shifted, all-True, or all-False one. The reviewer demonstrated a 1-node
`jnp.roll` shift of the slab mask passes every test in both files while
moving a measured AD gradient 11% (`-1.190e-3 -> -1.059e-3`), invisible to
the AD/FD gate because both sides of that comparison move together under
the same shifted mask.

Fixed per the reviewer's prescription: both locks now perturb at
`deps=0.5` and assert that the increased cells are exactly the cells that
are genuine slab material in the production assembly
(`materials.eps_r == _SLAB_EPS_R`), not just that deps=0 is a no-op.
Reproduced the mutation witness myself before pushing — both fixtures red
the new assertion under a 1-node `jnp.roll`:
```
test_waveguide_nu_flux_ad.py fixture:   Mismatched elements: 224 / 14592 (1.54%)
test_waveguide_nu_checkpoint.py fixture: Mismatched elements: 420 / 14880 (2.82%)
```

**MEDIUM — the sibling's "wrong physical location" claim above was FALSE
on that fixture.** Measured independently before writing this correction:
OLD (`pos_to_nu_index`) and NEW (`mask_on_coords`) masks occupy the **same
x-planes** `[28, 29, 30]`. The 114 differing cells are exactly the `y=A` /
`z=B` **hi-face** node planes — `Box.mask_on_coords`'s own half-open
`[lo, hi)` volume rule excluding its own hi face at the box's full-cross-
section extent, not a #562 coordinate-convention artifact (the lo-face
planes `y=0`, `z=0` are included in both OLD and NEW). A gradient computed
with the OLD mask vs. the NEW mask on this fixture is **bit-identical**:
`-0.023071737959980965` both, 0.00% difference. So this rewrite on the
checkpoint sibling is **latent-divergence cleanup** — removing the same
hand-rolled-comparator class before it could go stale under a future
change — not a fix for a live defect on this fixture. The headline
`test_waveguide_nu_flux_ad.py` fixture's diagnosis (genuine extra lo-face
cell from the #562 shift, 19.8% forward mismatch) is untouched and
correct. Both the commit message and `_eps_override`'s docstring now carry
the measured correction.

**LOW — dropped `_eps_override_for`'s dead `domain_x` parameter** (unused
in the function body since before #590) and updated its three call sites.
No signature-parity reason to keep it — the sibling `_eps_override` never
had this parameter.

## Verification (final)

- `pytest tests/test_waveguide_nu_flux_ad.py tests/test_waveguide_nu_checkpoint.py -m "slow or not slow"` → **8 passed**
- 1-node `jnp.roll` mutation witness on both fixtures → **both red** the new mask-shape assertions (shown above)
- `ruff check tests/test_waveguide_nu_flux_ad.py tests/test_waveguide_nu_checkpoint.py --select E,F,W --ignore E501,F401,E741,E731,E701,E702,E402` → clean

R3: memory=PR #564's own review reply (`b5fe569`, the reduce-to-production
pattern for F1/F2) reapplied to a third and fourth consumer of the same
convention; this PR's own round-1 review (independently re-verified every
claim — mask x-plane overlap, hi-face-only diff set, bit-identical
OLD/NEW gradient — before acting on it) | R2-attempts=1 per round (single
prescribed fix each time, not a repeated mechanism hypothesis) |
falsifier=the failing-then-passing reproduction (defect) plus the 1-node
`jnp.roll` mutation on both fixtures (lock tests), both reproduced before
pushing

Refs #564, #562

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/tests/test_waveguide_nu_checkpoint.py
+++ b/tests/test_waveguide_nu_checkpoint.py
@@ -78,12 +78,44 @@ def test_nu_checkpoint_no_longer_fenced_and_forward_identical():
 
 
 def _eps_override(sim, deps):
-    from rfx.runners.nonuniform import pos_to_nu_index
+    """NU-grid eps override: production-assembled eps + deps inside the slab.
+
+    #590 sibling: mirrors ``test_waveguide_nu_flux_ad.py::_eps_override_for``
+    — derive the baseline from ``assemble_materials_nu`` (the production NU
+    materials-assembly path) and the slab region from the Box's own
+    ``mask_on_coords``, never a hand ``pos_to_nu_index`` nearest-node
+    reconstruction (which went stale under #562's node-coordinate
+    convention change and was silently unprotected here since this test
+    only compares two runs that share the same override, not against a
+    production-default run).
+    """
+    from rfx.runners.nonuniform import assemble_materials_nu
+    from rfx.geometry.rasterize_grid import coords_from_nonuniform_grid
+
     grid = sim._build_nonuniform_grid()
-    eps = jnp.ones(grid.shape, dtype=jnp.float32)
-    i_lo = pos_to_nu_index(grid, (_SLAB_LO, _A / 2, _B / 2))[0]
-    i_hi = pos_to_nu_index(grid, (_SLAB_HI, _A / 2, _B / 2))[0]
-    return eps.at[i_lo:i_hi, :, :].set(_SLAB_EPS + deps)
+    materials, _debye_spec, _lorentz_spec, _pec_mask = assemble_materials_nu(sim, grid)
+    coords = coords_from_nonuniform_grid(grid)
+    slab_entry = next(e for e in sim._geometry if e.material_name == "diel")
+    slab_mask = slab_entry.shape.mask_on_coords(coords.x, coords.y, coords.z)
+    return jnp.where(slab_mask, materials.eps_r + deps, materials.eps_r)
+
+
+def test_eps_override_matches_production_assembly():
+    """#590 sibling regression: lock ``_eps_override`` to the production NU
+    materials assembly (the reduce-to-production pattern PR #564's review
+    used for F1/F2) — an elementwise assertion the helper's deps=0 baseline
+    equals a fresh, independent ``assemble_materials_nu`` call on the
+    identical fixture. Not ``@pytest.mark.slow``: no FDTD run, so it reds
+    in the fast lane if this helper ever drifts back to a hand-rolled
+    index reconstruction.
+    """
+    from rfx.runners.nonuniform import assemble_materials_nu
+
+    sim = _nu_sim()
+    grid = sim._build_nonuniform_grid()
+    materials, _, _, _ = assemble_materials_nu(sim, grid)
+    override = _eps_override(sim, jnp.asarray(0.0, dtype=jnp.float32))
+    np.testing.assert_array_equal(np.asarray(override), np.asarray(materials.eps_r))
 
 
 @pytest.mark.slow
--- a/tests/test_waveguide_nu_checkpoint.py
+++ b/tests/test_waveguide_nu_checkpoint.py
@@ -84,10 +84,22 @@ def _eps_override(sim, deps):
     — derive the baseline from ``assemble_materials_nu`` (the production NU
     materials-assembly path) and the slab region from the Box's own
     ``mask_on_coords``, never a hand ``pos_to_nu_index`` nearest-node
-    reconstruction (which went stale under #562's node-coordinate
-    convention change and was silently unprotected here since this test
-    only compares two runs that share the same override, not against a
-    production-default run).
+    reconstruction.
+
+    Measured correction (PR #591 review): on THIS fixture the OLD
+    (``pos_to_nu_index``) and NEW (``mask_on_coords``) masks occupy the
+    SAME x-planes (28, 29, 30) — the #562 node-coordinate convention shift
+    that broke the sibling ``test_waveguide_nu_flux_ad.py`` fixture does
+    NOT reach this one. The 114 cells that differ are exactly the y=A and
+    z=B hi-face node planes: Box's half-open ``[lo, hi)`` volume rule
+    excludes its own hi face, and this box's y/z extent is the full
+    PEC-walled cross-section, so those planes are the PEC-wall nodes
+    ``pos_to_nu_index`` had no shape-aware reason to exclude. Directly
+    measured: a gradient computed with the OLD mask and one computed with
+    the NEW mask on this fixture are IDENTICAL, ``-2.307174e-02`` both
+    (0.00% difference) — this rewrite is latent-divergence cleanup (the
+    same class of hand-rolled comparator removed before it COULD go stale
+    under a future change), not a fix for a live defect on this fixture.
     """
     from rfx.runners.nonuniform import assemble_materials_nu
     from rfx.geometry.rasterize_grid import coords_from_nonuniform_grid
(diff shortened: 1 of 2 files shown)


bk-squared:
## Review response (SHA `ac27ec7`)

Addressed all four items from the REQUEST-CHANGES review.

### BLOCKING — tautological lock tests, fixed

Confirmed the reviewer's finding independently before acting: at `deps=0`,
`jnp.where(mask, eps_r + 0, eps_r) == eps_r` for **any** mask, so the original
lock (elementwise equality at deps=0) could not distinguish a correct mask
from a shifted/all-True/all-False one.

Both locks now perturb at `deps=0.5` and assert that the cells which increased
are exactly the cells that are genuine slab material in the production
assembly (`materials.eps_r == _SLAB_EPS_R`), per the reviewer's prescription
(adapted per file — `_SLAB_EPS_R` / `_SLAB_EPS`, `_eps_override_for` /
`_eps_override`).

**Mutation witness, reproduced myself** (1-node `jnp.roll` shift of the slab
mask along x, both fixtures):

```
=== test_waveguide_nu_flux_ad.py fixture ===
MUTATION CORRECTLY REDS the new assertion:
Arrays are not equal
Mismatched elements: 224 / 14592 (1.54%)

=== test_waveguide_nu_checkpoint.py fixture ===
MUTATION CORRECTLY REDS the new assertion:
Arrays are not equal
Mismatched elements: 420 / 14880 (2.82%)
```

Both new assertions red under the mutation that silently passed all 8 tests
before this fix.

### MEDIUM — sibling narrative correction

Verified independently (not taken on the reviewer's word) before writing
anything:

- OLD (`pos_to_nu_index`) and NEW (`mask_on_coords`) masks on the checkpoint
  fixture occupy the **same x-planes**: `[28, 29, 30]`.
- The 114 differing cells are exactly the union of the `y = A` (index 30) and
  `z = B` (index 7) **hi-face** node planes — confirmed cell-for-cell against
  that hypothesis (114/114 match). These are excluded by
  `Box.mask_on_coords`'s own half-open `[lo, hi)` volume rule at the box's own
  y/z extent (the box spans the full PEC-walled cross-section), not a #562
  coordinate-convention artifact — the lo-face planes (`y=0`, `z=0`) are
  correctly included in both OLD and NEW.
- Gradient measured with the OLD mask vs. the NEW mask on this fixture:
  `-0.023071737959980965` both, **bit-identical**, 0.00% difference.

The PR's original claim ("the slab silently sat at the wrong physical
location") was wrong for this fixture. Corrected in both the commit message
and the `_eps_override` docstring: this rewrite is **latent-divergence
cleanup** — removing the same hand-rolled-comparator class before it could go
stale under a future change — not a fix for a live defect on this fixture.
The headline `test_waveguide_nu_flux_ad.py` fixture's diagnosis (genuine
extra lo-face cell from the #562 shift, 19.8% forward mismatch) is untouched
and correct.

### LOW — dead `domain_x` parameter

Dropped from `_eps_override_for`'s signature (it was never used in the
function body, even pre-#590) and updated its three call sites
(`_s21_mag2`, `test_nu_flux_smatrix_forward_matches_untraced`, the lock
test). No signature-parity reason to keep it — the sibling `_eps_override`
never had this parameter.

### Verification

- `pytest tests/test_waveguide_nu_flux_ad.py tests/test_waveguide_nu_checkpoint.py -m "slow or not slow"` → **8 passed**
- 1-node `jnp.roll` mutation witness on both fixtures → **both red** the new assertions (shown above)
- `ruff check tests/test_waveguide_nu_flux_ad.py tests/test_waveguide_nu_checkpoint.py --select E,F,W --ignore E501,F401,E741,E731,E701,E702,E402` → clean

Pushed as `ac27ec7` on top of `fd9bb2d`.

PR #592, opened 2026-08-06
fix(ci): split weekly slow-tests into shard+highmem lanes, add live RSS sampler (issue #545 step 3)

### LINKED ISSUE #545: Weekly slow lane: runner-shutdown cancellations are RECURRING — shards 1+3 killed two consecutive weeks (07-27, 08-03)

bk-squared:
## Summary

Implements the 3-part resolution path from issue #545's 2026-08-06 discriminator-run comment (run 31103127491, first run with both the #566 rebalance and the RSS reporter live). That run proved the recurring shard-1/2/3 kills (exit 143, zero assertions/exceptions) are OOM: every killed shard's process had already crossed **12-15 GB RSS high-water** before its own death. The fatal tests were **memory-heavy (3-24 GB) but time-cheap (38-100s)**; pytest-split's `duration_based_chunks` balances on time only, so it freely relocates these tests between shards on every `.test_durations` regeneration — a coin flip for which shards OOM next.

**This is NOT "Closes #545"** — the issue stays open until a green weekly run proves the split actually prevents the OOM kill. See the **predicted partial outcome** below before treating the next run as a refutation either way.

### 1. `highmem` marker (registered in `pyproject.toml`)

Marked the 7 tests confirmed memory-heavy in the discriminator run. **Important provenance caveat, added after review**: the `[rss]` post-test hook only prints *after* a test completes, so for the 3 tests that were themselves the ones killed (no PASSED suffix in the original run), their own footprint was never captured in CI — the number recorded is the *preceding* test's high-water reading, i.e. "the process was already at N MB entering this test."

| test | existing marker(s) | measured |
|---|---|---|
| `test_forward_tfsf_oblique_differentiable` | `slow` | **own footprint unmeasured in CI** (it was shard 1's kill); process already at 14,976 MB entering it (that's the preceding coax test's reading). Measured directly: 23.78 GB peak locally (taskset -c 0-1), 23,964 MB on independent standalone remeasurement |
| `test_oblique_reflection_magnitude_differentiable` | `slow` | **own footprint unmeasured in CI** (shard 2's kill); process already at 12,187 MB entering it (preceding msl-referee test's reading) |
| `test_ram_gradient_ad_vs_fd` (parametrized sigma/eps) | `slow` | `[sigma-1.0]`: **own footprint unmeasured in CI** (shard 3's kill); process already at 14,854 MB entering it (preceding mu_r test's reading). `[eps-4.0]` shares the same checkpointed forward and is marked with it (pytest marks apply per-function, not per-param) |
| `test_coax_reflection_grad_finite_and_fd_consistent` | `slow_physics` | +10.8 GB delta (self-attributed — this test completed normally) |
| `test_auto_eps_msl_gradient_matches_fd_mini_referee` | **none** (no slow/slow_physics/gpu) | +8.1 GB delta (self-attributed); independently remeasured at ~8.9 GB |
| `test_oblique_complex_gamma_differentiable` | `slow` | +10.7 GB delta (self-attributed) |
| `test_mu_r_gradient_ad_vs_fd` | `slow` | +3.3 GB delta (self-attributed) |

`highmem` is **additive only**: `pyproject.toml`'s `addopts` (the line *above* the markers list) still filters `not gpu and not slow and not slow_physics`, untouched. No test's default-fast-lane membership changes — `test_auto_eps_msl_gradient_matches_fd_mini_referee` still runs in the default/PR-gating lane exactly as before; `highmem` only pulls it out of the *weekly* 4-shard rotation. (That test's continued exposure in the fast lane, with no RSS diagnostics there, is filed separately as #594.)

### 2. Weekly workflow split (`.github/workflows/validation.yml`)

- The 4 `slow-tests` shard jobs now select `-m "not gpu and not highmem"`.
- New `slow-tests-highmem` job runs `-m "highmem and not gpu"` **serially** (no shard split, no xdist) so only one multi-GB peak exists on the runner at any instant. Same `RFX_WEEKLY_RSS=1` env, plus `-s` (see step 3). 30-minute timeout.
- **Corrected runner ceiling: ~16 GB, not ~7 GB.** Two witnesses: shard 1 survived past 14,976 MB (an entering-state reading, not a death); the ~8.9 GB msl-referee highmem test passes on `ubuntu-latest` in the default fast lane on every PR.
- **Predicted outcome (not yet proven — this split has not run in CI)**: the 4-shard job should stop OOMing. `slow-tests-highmem` will likely **still** see `test_forward_tfsf_oblique_differentiable` exit 143 — serialization removes only the ~4 GB of accumulated baseline from co-located tests, not that one test's own ~24 GB footprint, which alone exceeds the ~16 GB ceiling. This is *not* "several tests each ~2x a ~7 GB ceiling" — it's "one test ~1.5x a ~16 GB ceiling, the rest fit." If that test alone dies in its own job, that is **still an improvement** (a single named, reproducible test death is a strictly better signal than a random shard death among ~1500 tests) — not a refutation of the split. Next steps per #545 item 3, in order: (a) an explicit `checkpoint_segments` count for that test's forward; (b) a larger-runner label for `slow-tests-highmem`; (c) exclude it from CI and track it as a manual/local-only regression check.
- `.test_durations` needs no regeneration — its now-unused entries for the deselected `highmem` tests are simply not seen by pytest-split's `-m` selection, not stale.

### 3. Live RSS sampler (closes the discriminator gap, `conftest.py`)

The discriminator run's writeup disclosed a remaining gap: `ru_maxrss` only updates in the post-test hook, so the RSS at the exact SIGKILL instant is never captured. Added `_RSSLiveSampler`: a background **daemon thread** (stdlib `threading` only) polling `/proc/self/status` VmHWM every ~2s, printing `[rss-live] VmHWM N MB [during <nodeid>]` (`flush=True`) whenever the high-water mark advances past the threshold.

**Review caught this shipping dead**: pytest's own capture swallows a background thread's bare `print()` even with `flush=True` — measured 0 `[rss-live]` lines at the workflow's exact `-v -ra` flags, 0 under a `timeout -s KILL` rehearsal of the real SIGKILL scenario, 0 even with `--capture=tee-sys`. **Fixed** by adding `-s` (`--capture=no`) to both weekly pytest steps. Proven with a subprocess-level test pair that spawns a real child `pytest`, self-SIGKILLs it mid-test, and inspects stdout from *outside* pytest's own capture layer:

```
# with -s: [rss-live] VmHWM 208 MB during ...test_rss_live_sampler_helper_allocates_and_self_sigkills
#          survives in the streamed stdout; process then Killed (exit 137)
# without -s (identical scenario): zero [rss-live] output
```

Also added: an independent-oracle band check (`_read_proc_vmhwm_mb` vs `resource.getrusage().ru_maxrss`) so a kB/MB unit-conversion mutation can't silently pass, and a narrow `(ValueError, OSError)` guard around `_sample_once` scoped only to the interpreter/pytest-teardown closed-stream race (not a bare `except Exception` — see `feedback_broad_except_eats_async_timeout`, #555, where a broad except swallowed a real SIGALRM timeout and hung a process).

## Verification

- `pytest tests/test_weekly_rss_reporter.py -v`: **18 passed, 1 skipped** (the SIGKILL helper only runs as a subprocess target by design; includes the new independent-oracle band check and the two subprocess SIGKILL-survival tests).
- All 7 `highmem` tests (8 nodeids incl. sigma/eps parametrization) run and **pass functionally**: `8 passed, 2 warnings in 647.23s (0:10:47)`.
- `ruff check --select E,F,W --ignore E501,F401,E741,E731,E701,E702,E402` on every touched file: clean.
- **Partition proof** (`--collect-only`; environment-dependent, invariant holds either way):
  - This dev environment (no optional `cad` extras): old `-m "not gpu"` = **3320**; new shard selection `-m "not gpu and not highmem"` = **3312**; new highmem selection `-m "highmem and not gpu"` = **8**. 3312+8=3320, no overlap, no loss.
  - Reviewer's environment (optional extras installed): 3328 = 3320 + 8 — same invariant, different baseline total.
  - Per-shard `--splits 4` breakdown also sums exactly to the shard-selection total, confirming pytest-split runs cleanly against the existing `.test_durations` file with no regeneration.

## Follow-up filed

#594: the ~8.1-8.9 GB `test_auto_eps_msl_gradient_matches_fd_mini_referee` (marked `highmem` here) runs in the default fast lane every PR with no `RFX_WEEKLY_RSS` diagnostics and the same pytest-split shard-reassignment exposure #545 diagnosed in the weekly lane. Proposes either RSS diagnostics on `pr-tests.yml`'s `fast-suite`, or pinning that test's shard assignment. Not fixed in this PR — scoped as a triage issue.

## R3 self-audit

1. Contradicted by known memory/feedback? — none found after grep; `feedback_dont_over_fragment_prs.md` supports landing all 3 resolution steps in one PR; `feedback_broad_except_eats_async_timeout` (#555) directly informed the narrow-except design in `_run`; `feedback_provenance_mismatch_class.md` directly named the 3-comment misattribution bug review caught.
2. R2 diminishing-returns loop? — n/a, CI-plumbing fix responding to an already-diagnosed OOM verdict, not a physics/extractor investigation loop.
3. Falsifier: the next scheduled/manual `validation.yml` run — `slow-tests` (4-shard) should complete without exit 143; `slow-tests-highmem` is **predicted** to still lose `test_forward_tfsf_oblique_differentiable` to exit 143 specifically (not a refutation, see above) while the other 6 highmem tests should pass in-band.

## Test plan

- [x] `pytest tests/test_weekly_rss_reporter.py -v` — 18 passed, 1 skipped
- [x] Run all 7 `highmem`-marked tests directly — 8/8 nodeids passed (647s)
- [x] Subprocess SIGKILL-survival test pair — both pass, manually reproduced
- [x] `--collect-only` partition proof (invariant holds across environments)
- [x] `ruff check` on all touched files — clean
- [x] Filed follow-up issue #594 for the fast-lane structural risk
- [ ] **Falsifier (next weekly run)**: `slow-tests` completes cleanly; `slow-tests-highmem` shows the predicted single-test death (or, better than predicted, none)

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -15,25 +15,54 @@ jobs:
   # Weekly run 31103127491 (2026-08-06, first run with both the #566 rebalance
   # AND the RSS reporter live) proved the recurring shard-1/2/3 kills (exit 143,
   # zero assertions/exceptions) are OOM: every killed shard's process had
-  # already crossed 12-15 GB RSS high-water before its own death, 2-2.5x the
-  # ~7 GB hosted-runner-class ceiling. The fatal tests in each case were
-  # memory-heavy (8-24 GB) but TIME-cheap (38-100s); pytest-split's
-  # duration_based_chunks balances on time only, so it freely relocates these
-  # tests between shards on every .test_durations regeneration -- a fresh
-  # durations file is a coin flip for which shards OOM next. The fix is to
-  # pull the 7 confirmed memory-heavy tests (marked `highmem`, see pyproject
-  # markers + issue #545's discriminator-run comment for the per-test
-  # deltas/peaks) out of the time-balanced shard rotation entirely and run
-  # them in their own job, one process at a time, so at most one multi-GB
-  # peak exists on the runner at any instant.
+  # already crossed 12-15 GB RSS high-water before its own death. pytest-split's
+  # duration_based_chunks balances on time only, so it freely relocates the 7
+  # confirmed memory-heavy (3-24 GB, TIME-cheap at 38-100s each) tests between
+  # shards on every .test_durations regeneration -- a fresh durations file is a
+  # coin flip for which shards OOM next. The fix is to pull those 7 tests
+  # (marked `highmem`, see pyproject markers + issue #545's discriminator-run
+  # comment for the per-test deltas/peaks) out of the time-balanced shard
+  # rotation entirely and run them in their own job, one process at a time, so
+  # at most one multi-GB peak exists on the runner at any instant.
+  #
+  # CORRECTED runner ceiling: PR #592 review re-measurement puts the actual
+  # ceiling at ~16 GB, not ~7 GB -- two witnesses: shard 1 SURVIVED past
+  # 14,976 MB (its own entering-state reading, not a death), and the ~8.9 GB
+  # msl-referee highmem test (test_auto_eps_msl_gradient_matches_fd_mini_referee)
+  # passes on ubuntu-latest in the DEFAULT fast lane on every PR. So this is
+  # NOT "several tests each ~2x an ~7 GB ceiling" -- it is ONE test
+  # (test_forward_tfsf_oblique_differentiable, ~24 GB own footprint, ~1.5x the
+  # ~16 GB ceiling) that will not fit under any serialization, plus 6 tests
+  # that do fit under ~16 GB and only needed to stop competing for headroom
+  # with the rest of a 3300-test shard.
+  #
+  # PREDICTION (not yet proven -- this split has not run in CI): the 4-shard
+  # slow-tests job should stop OOMing now that the highmem tests are gone.
+  # slow-tests-highmem will likely still see
+  # test_forward_tfsf_oblique_differentiable exit 143 -- serialization removes
+  # the ~4 GB of accumulated baseline from co-located tests, which is not
+  # enough to fit a ~24 GB single-test footprint under a ~16 GB ceiling. If
+  # that happens, this IS still an improvement (a single named, reproducible
+  # test death is a strictly better signal than "one of ~1500 tests in a
+  # randomly balanced shard died somewhere"), not evidence the split failed.
+  # Next steps per issue #545 item 3, in order of preference: (a) an explicit
+  # checkpoint_segments count for that test's forward, to see whether its ~24
+  # GB tape is compressible; (b) a larger-runner label for slow-tests-highmem
+  # if (a) does not land it under ~16 GB; (c) exclude it from CI entirely and
+  # track it as a manual/local-only regression check.
   #
   # .test_durations interaction: the 4-shard job's selection (`not gpu and
   # not highmem`) now excludes the 7 highmem tests, so duration_based_chunks
-  # balances only over the remaining ~3312 tests. No regeneration of
-  # .test_durations is needed for this split -- pytest-split's
-  # --collect-only entry simply won't see the deselected highmem tests, and
-  # the file's existing per-test duration entries for them (if present) are
-  # just unused by this selection, not stale/wrong.
+  # balances only over the rest of the "not gpu" suite (~3300 tests; the exact
+  # count depends on which optional extras -- e.g. `cad` -- are installed, but
+  # the partition invariant, shard-selection-count + highmem-selection-count
+  # == old not-gpu-selection-count with no overlap, holds regardless: verified
+  # both at 3320 = 3312 + 8 without optional extras and at 3328 = 3320 + 8
+  # with them installed, --collect-only, PR #592). No regeneration of
+  # .test_durations is needed for this split -- pytest-split's --collect-only
+  # entry simply won't see the deselected highmem tests, and the file's
+  # existing per-test duration entries for them (if present) are just unused
+  # by this selection, not stale/wrong.
   slow-tests:
     runs-on: ubuntu-latest
     # 90 not 60: the slowest shard reached ~52 min on 2026-07-27 and the slow
@@ -73,6 +102,19 @@ jobs:
       # kill from a pure infra shutdown if a shard is killed again — no-op
       # elsewhere (fast-suite/pr-tests.yml do not set this).
       #
+      # -s (issue #545 step 3, PR #592 review): REQUIRED for the live sampler's
+      # background-thread prints to survive at all. Measured directly: under
+      # pytest's default fd-level capture (no -s), a completing test's
+      # [rss-live] lines are captured into pytest's internal per-test buffer
+      # and never surface -- 0 lines, even under a `timeout -s KILL` rehearsal
+      # of the exact SIGKILL scenario this sampler exists for, and even with
+      # `--capture=tee-sys`. Only `-s` (`--capture=no`) sends the thread's
+      # writes straight to the real stdout fd, where flush=True can actually
+      # do its job. See
+      # tests/test_weekly_rss_reporter.py::test_live_sampler_line_survives_sigkill_with_dash_s
+      # (and its without-`-s` regression-pin sibling) for the subprocess-level
+      # proof of this property.
+      #
       # "not highmem" (issue #545 step 3): the 7 confirmed memory-heavy tests
       # move to the slow-tests-highmem job below, see the block comment above
       # this job for why.
@@ -80,7 +122,7 @@ jobs:
         env:
           RFX_WEEKLY_RSS: "1"
         run: >
-          pytest -v -ra
+          pytest -v -ra -s
           -m "not gpu and not highmem"
           --splits 4 --group ${{ matrix.group }}
           --ignore=tests/test_meep_crossval.py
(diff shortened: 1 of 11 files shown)


bk-squared:
## Review response (2 blocking items addressed)

**CRITICAL fixed** -- the live sampler's prints are swallowed by pytest's own capture unless `-s` is passed. Added `-s` to both weekly pytest steps in `validation.yml` (kept `-v -ra`), and added a subprocess-level SIGKILL test pair that spawns a real child `pytest`, self-SIGKILLs it mid-test, and inspects stdout from **outside** pytest's own capture layer:

- `tests/test_weekly_rss_reporter.py::test_live_sampler_line_survives_sigkill_with_dash_s` -- **PASS**
- `tests/test_weekly_rss_reporter.py::test_live_sampler_line_lost_without_dash_s` (regression pin for why `-s` is required) -- **PASS**

Manual standalone repro of the same scenario:

```
# with -s:
tests/test_weekly_rss_reporter.py::test_rss_live_sampler_helper_allocates_and_self_sigkills [rss-live] VmHWM 208 MB during tests/test_weekly_rss_reporter.py::test_rss_live_sampler_helper_allocates_and_self_sigkills
Killed (exit 137)

# without -s (identical scenario):
tests/test_weekly_rss_reporter.py::test_rss_live_sampler_helper_allocates_and_self_sigkills
Killed (exit 137)   # zero [rss-live] output
```

Also added: an independent-oracle band check (`_read_proc_vmhwm_mb` vs `resource.getrusage().ru_maxrss`) so a kB/MB conversion mutation can't silently pass, and a narrow `(ValueError, OSError)` guard around `_sample_once` in `_run` scoped only to the interpreter/pytest-teardown closed-stream race (not a bare `except Exception` -- per `feedback_broad_except_eats_async_timeout`, #555).

**HIGH (honesty) fixed** -- the PR body/workflow comments now PREDICT that `test_forward_tfsf_oblique_differentiable` is likely to still exit-143 in the serial `slow-tests-highmem` job: its own footprint (23,964 MB on your standalone remeasurement, agreeing with the earlier 23.78 GB taskset number) exceeds the corrected ~16 GB runner-class ceiling (witnesses: shard 1 survived past 14,976 MB; the ~8.9 GB msl-referee highmem test passes every PR in the fast lane) even fully serialized. Reframed from "several tests 2x a ~7 GB ceiling" to "one test ~1.5x a ~16 GB ceiling, the rest fit." Next steps per #545 item 3 are now spelled out in order: explicit `checkpoint_segments` count for that test, a larger-runner label, or exclusion.

**Same-pass fixes**:
- 3 of 7 highmem citation comments were misattributing a *preceding* test's high-water reading to the killed test itself (the `[rss]` post-test hook only prints after a test completes, so a killed test's own line never fires) -- fixed in `test_forward_tfsf_differentiable.py`, `test_oblique_fresnel_magnitude.py`, `test_ram_multilayer_inverse_design.py`.
- "~7 GB ceiling" -> "~16 GB" throughout.
- Partition counts now state the collection environment (3320/3312/8 without optional extras vs 3328/3320/8 with them here) instead of a hardcoded "~3312"; the invariant (shard-count + highmem-count == old-count, no overlap/no loss) holds either way.
- `pyproject.toml`: "addopts below" -> "addopts above" (it's the line above markers); "8-24 GB" -> "3-24 GB" (mu_r's +3.3 GB is the true minimum).
- Filed #594 for the follow-up you flagged: the ~8.1-8.9 GB msl-referee highmem test runs in the default fast lane every PR with no `RFX_WEEKLY_RSS` diagnostics and the same shard-reassignment exposure -- scoped as a triage issue, not fixed here (proposes both remedies you named: RSS diagnostics on fast-suite, or pinning).

Full verification: `pytest tests/test_weekly_rss_reporter.py` -> 18 passed, 1 skipped (the SIGKILL helper only runs as a subprocess target, by design). `ruff check` clean on all touched files.

New SHA: `20d7592`

Issue #589, opened 2026-08-06
coax-MSL transition (issue 489 leg 4): settled run, MSL column-power question, reciprocity/S22 measurement


bk-squared:
## Background

Issue #489's leg 4 (`compute_coax_msl_transition`, the coax<->MSL transition lane) is the one leg of the #489 dev track that stays **EXPERIMENTAL, diagnostic-only** even after #489's other three legs (referee, mesh-refinement, AD) closed and the two-port core's label lifted (PR #588). This issue tracks leg 4's own continuing work, since #489 itself is closed and should not reopen for it.

Attempt 2 (PR #585, merge `da443f5`) landed the ladder/clearance/API-decoupling instrument fixes and confirmed the gamma-vs-beta discriminant (provisional, run-length-invariant), but left reciprocity, `|S22|`, and max`|S|` **UNMEASURED AT THIS SETTLING** — the 45000-step committed fixture only reached -19.7/-17.9 dB ring-down settling, short of this repo's -40 dB rule (`CLAUDE.md` ring-down settling witness).

## What this issue tracks

1. **The settled run** (predeclared in `tests/test_coax_msl_transition.py::SETTLED_RUN_RECORD`, currently `status: "UNRUN"`): rerun the committed attempt-2 fixture at `target_n_steps=135000` to clear (or measure against) the -40 dB rule and get an actual pass/fail read on reciprocity/`|S22|`/max`|S|`. The driver (`scripts/diagnostics/coax_msl_transition_settled_run.py`) and VESSL YAML (`scripts/vessl_coax_msl_transition_settled_run.yaml`) landed in PR #588; the run itself was deliberately NOT submitted by that PR (mounted-checkout pattern — needs `main` to carry the driver first). Submit after PR #588 merges.
2. **The MSL-driven column-power open question** (PR #585 review finding B5): `sum_j |S_ij|^2` for driven port `msl` is mostly far below 1 (0.0018-0.204 at 20000 steps, 0.0104-1.218 at 45000 steps) on a nominally lossless structure, with one bin crossing 1 at the less-settled checkpoint — scaling-invariant (survives the drive-amplitude-gap retraction), so this is real evidence of *something*, not a normalization artifact. Where does the MSL column's power go? Not answered by attempt 2.
[...]

PR #597, opened 2026-08-07
feat(sparams): extra_flux_monitors= energy-witness opt-in + the #589 flux-adjudication driver

bk-squared:
Implements the instrument half of the **issue #589 pre-declaration** (2026-08-07 comment): face-resolved flux accounting on the settled attempt-2 fixture, to adjudicate the MSL-driven column-power deficit between **(a) physically departing power** and **(b) extraction blindness**. No fixture is promoted, no gate is changed, and **no VESSL run is submitted by this PR** — submission follows the mounted-checkout rule (driver must be on main) and a coordinate-addendum comment on #589 that precedes it.

## What this adds

- **`extra_flux_monitors=` opt-in** on `compute_coax_msl_transition` and `compute_coaxial_two_port`: entries are the objects `add_flux_monitor` registers (built on a scratch `Simulation`, same domain); each internal drive run accumulates the requested Poynting-flux planes with **fresh per-drive accumulators**, and per-drive spectra return name-keyed on `result.flux_monitors`. The registered-monitor guard is **unchanged**, and the internal run call is byte-identical when the opt-in is unset (kwargs added only when present — the existing monkeypatch fake-run test keeps its explicit signature).
- `build_flux_monitor_cfgs(..., entries=)` override (sim still consulted for `_freq_max`/`_domain` only).
- `flux_monitors: dict | None = None` on both result dataclasses; class docstring records that the #589 pre-declaration supersedes the closed-box variant as the chosen instrument.
- **Driver** `scripts/diagnostics/coax_msl_flux_adjudication.py`: C1 (MSL through-line control — flux-measured P_inc, plane-invariance, launch-spill background, settling), C2 (coax two-port control — attenuation-model gate vs the run's own Re(γ), shield tightness), target (settled 135000-step recipe + 6-face box, aperture partition, shifted-sibling monitors for the face-shift gate, two-plane witness). Controls hard-gate the target (rc=2 skips it). The decision rule is the issue comment's, verbatim; the header carries the face coordinate table the pre-declaration delegated to this PR.
- **VESSL YAML** (tracked by the same force-add exception as the settled-run/gpu-suite/crossval YAMLs); it re-runs the bit-identity witness in-job before the physics stages.

## Witnesses — measured, not promised

- **Non-perturbation witness** (`test_extra_flux_monitors_do_not_perturb_s`, slow_physics): S **bit-identical** with and without monitors on the attempt-2 fixture; settling_db identical; spectra finite and name-keyed. Measured pre-commit on CPU: `1 passed in 40.05s`.
- Fast lane: guard + entry-validation tests, `2 passed in 2.14s`.
- Flux sign convention verified live (top-drive two-port reads negative net flux at a mid-line plane → "positive = +axis" holds; the driver's outward multipliers key off it).
- 3-stage driver smoke at toy scale: exit 0; fail-loud paths verified (C1 gate failure blocks the target; missing `c1_result.json` blocks the target; unsettled toy physics returns NON-CLOSING with the witness numbers visibly wrong, as they should be).
- ruff (CI select/ignore set): clean.

R3: memory=ledger leg-4 entries + power-flux fence (witness-not-extractor stated in both docstrings) + rf_core_audit half-step class (face-shift gate is the declared falsifier) + gate-binds-artifact (witness measured) | R2-attempts=0 (instrument build; the lane's one pre-declared attempt fires at run time) | falsifier=bit-identity test green + smoke rc/gate behavior verified pre-commit

### DIFF
--- a/docs/guides/api_symbol_inventory.json
+++ b/docs/guides/api_symbol_inventory.json
@@ -2434,7 +2434,8 @@
       "feed_impedance",
       "cond_warn",
       "strict_passivity",
-      "skip_preflight"
+      "skip_preflight",
+      "extra_flux_monitors"
     ],
     "compute_coaxial_line_reflection": [
       "termination",
@@ -2473,7 +2474,8 @@
       "feed_impedance",
       "cond_warn",
       "strict_passivity",
-      "eps_scale"
+      "eps_scale",
+      "extra_flux_monitors"
     ],
     "compute_mixed_s_matrix": [
       "n_steps",
--- a/rfx/api/_sparams.py
+++ b/rfx/api/_sparams.py
@@ -269,6 +269,52 @@ def _warn_msl_wave_split_unreliable(
 _SETTLING_WITNESS_DB = -40.0
 
 
+def _validate_extra_flux_monitor_entries(entries, domain, fn_name):
+    """Light re-validation of ``extra_flux_monitors=`` entries (#589 opt-in).
+
+    Entries are the objects ``Simulation.add_flux_monitor`` registers,
+    typically built on a scratch ``Simulation`` sharing this sim's domain —
+    ``add_flux_monitor`` already validated them against THAT domain, so this
+    only re-checks the one property that can silently diverge between the
+    two sims (the normal-axis coordinate inside THIS domain). Energy-witness
+    channel only: spectra come back on ``result.flux_monitors``; nothing
+    here feeds the S-parameter math (the registered-monitor guard below is
+    unchanged — this extractor still builds its own DFT planes).
+    """
+    if not entries:
+        return
+    axis_to_index = {"x": 0, "y": 1, "z": 2}
+    seen = set()
+    for pe in entries:
+        for attr in ("name", "axis", "coordinate"):
+            if not hasattr(pe, attr):
+                raise TypeError(
+                    f"{fn_name}(): extra_flux_monitors entries must be the "
+                    "objects Simulation.add_flux_monitor() registers "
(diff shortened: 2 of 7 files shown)


bk-squared:
**Adversarial review round 1 (90b5db1): HOLD from all three independent passes → all findings fixed in 3389bd7.**

Three independent review passes (pre-declaration fidelity / code correctness / test-CI adequacy) each returned HOLD. The catch that mattered most: **C1's launch-spill background included the xm face, which on the through-line control carries the full transmitted power (~0.95 with synthetic through-spectra) — the control would have failed unconditionally on an arithmetic bug**, burning this lane's one pre-declared attempt. Also caught: arm (b) unreachable under a tripped two-plane witness (contamination rule asymmetric); the posted spill subtraction reported-but-not-applied; `--stage target` bypassing both control gates; C2's posted W1 closure gate missing; a claimed-but-absent monitor-coordinate assert; W2 being algebraically identical to W1 (tautology-as-witness class); three shift members landing 1 cell from the CPML inner edge; and the api-reference drift red (the pinned-signature gate doing its job on the new kwarg).

Every fix is in 3389bd7 and each departure from the original posted wording is DECLARED in the #589 addendum that will be posted before submission (spill composition excluding xm, W2 reframed as bookkeeping, C2 W1 added, one-sided shifts, own-drive γ model). Re-verification: gate-refusal path rc=2, κ warning trigger fires, toy target NON-CLOSING as expected, `check_api_reference` OK, ruff clean, fast tests green.

Round 2 = CI on 3389bd7; merge on green.

PR #602, opened 2026-08-07
feat(waveguide): energy ring-down settling witness on the S-matrix path (#538)

### LINKED ISSUE #538: compute_waveguide_s_matrix does not populate settling_db — the energy-based ring-down witness exists only on the lumped/MSL path
## Summary

The repo's settling rule (CLAUDE.md, simulation_methodology) prefers an ENERGY-based ring-down witness for resonant numbers taken at fixed `num_periods` on an open CPML domain: terminal energy in dB below the post-source peak, rule < −40 dB. The machinery exists — `settling_db` on the S-matrix result, with the aggregate truncation warning in `rfx/api/_sparams.py` — but it is populated only by the **lumped/MSL** extraction path (`MSLSMatrixResult.settling_db`, the path crossval case 07 used to commit "settling witness −72 dB").

`compute_waveguide_s_matrix` returns `settling_db = None` unconditionally. Verified across three independent full generations of crossval case 19 (a five-iris WR-90 resonant filter, exactly the structure class the rule targets): every row commits `settling_db: null`.

## Why it matters

Case 19 had to substitute the independent-axis criterion — a `num_periods` 400/600/800 scan holding f0/BW to one bin, plus per-run column-power cleanliness, with the under-settled np=200 run demonstrably firing the criterion at colpow 1.207. That substitute is defensible (truncation shows up as non-passivity first on this structure, measured), but:

- it costs two extra full FDTD runs per generation where the energy witness would be free;
- the doubling axis is a *stability* argument, not the mandated *energy* argument, and the settling rule itself notes that indirect comparisons can pass transient mush;
- any future waveguide-path resonant case (filters, cavities, diplexers) inherits the same gap.

## Suggested fix

Port the settling computation to the waveguide path: end/peak field-energy ratio at the port probe planes per driven run, same `_SETTLING_WITNESS_DB` threshold and the same aggregate warning. The lumped/MSL implementation looks directly reusable — the waveguide path already records per-port time series for the DFTs.

## Provenance

Found during the independent numeric audit of crossval case 19 (#499): the case's claim_scope briefly asserted the energy witness was committed, the audit showed the fields were null in every row across three generations, and the claim has been corrected to name this gap instead.

bk-squared:
Closes #538. The waveguide S-matrix path now carries the same energy ring-down settling witness as the lumped/MSL path — per driven run, from the `v_probe_t` time series the scan **already records** for the DFT extraction, so nothing enters the jitted graph and S cannot be perturbed.

## What changed

- `settling_db_from_port_records` (new, `waveguide_port.py`): worst end/peak `|V_probe|²` tail ratio across a driven run's port probe planes; NaN under tracing (same guard as MSL).
- `return_settling=` opt-in on **all three** extractors (plain / flux / normalized) → `(S, settling_db)`; default `False` keeps every existing caller and the AD contract tests unchanged. Flux/normalized record the **device** run's witness (the matched straight reference would flatter it).
- `compute_waveguide_s_matrix`: every normalize branch threads the witness into the new `WaveguideSMatrixResult.settling_db` and fires the same aggregate truncation warning as MSL. NU sibling not covered — documented and tracked on #538.

## Measured

- All three normalize modes populate `[-5.8, -1.9]` dB on the WR-90-class two-port fixture at `num_periods=6` (identical across modes — same drives/records) and the truncation warning **fires** on that deliberately unsettled record.
- `num_periods` 4 → 16 drives the witness more negative — the direction falsifier is a committed test.
- Public path run-to-run bit-deterministic with the witness attached. New test file: `3 passed in 19s` (fast lane).
- `check_api_reference` OK (regenerated — the signature pin caught the new kwarg, as designed); ruff clean.

Once this is on main, crossval case 19's next regeneration can retire its `num_periods`-doubling substitute criterion in favor of the direct energy witness (noted on #538; not done preemptively here).

R3: memory=gate-can-bind-artifact (structural non-perturbation + determinism pin + existing S gates) + agreement-is-not-independence (device run, not the reference) | R2-attempts=0 | falsifier=direction test + real truncation-warning firing

### DIFF
--- a/docs/guides/api_symbol_inventory.json
+++ b/docs/guides/api_symbol_inventory.json
@@ -956,7 +956,8 @@
         "aniso_eps",
         "conformal_weights",
         "aniso_inv_eps",
-        "checkpoint_segments"
+        "checkpoint_segments",
+        "return_settling"
       ]
     },
     "extract_waveguide_sparams": {
--- a/docs/guides/sparameter_support_matrix.md
+++ b/docs/guides/sparameter_support_matrix.md
@@ -358,10 +358,12 @@ outside the single-dielectric-slab configuration described next.
   29%): max interior diff `0.0320 ns` against a pinned-constant gate of
   `0.042 ns` (the raw diff against the exact closed-form derivative,
   `0.0248 ns`, is recorded for context but is not the gated number). A
-  record-length settling witness (`num_periods` 60 vs 120 --
-  `compute_waveguide_s_matrix` has no built-in energy-based `settling_db`
-  for the waveguide port family, unlike the MSL calculator) agrees to
-  `0.000 ns`. A domain-size invariance run (`+100 mm` growth) holds the pass
+  record-length settling witness (`num_periods` 60 vs 120; at the time of
+  that run `compute_waveguide_s_matrix` had no built-in energy-based
+  `settling_db` -- since issue #538 the uniform single-mode waveguide
+  lanes carry the same energy-based `settling_db` field and -40 dB
+  aggregate warning as the MSL calculator, with the NU and multimode
+  branches still un-adopted) agrees to `0.000 ns`. A domain-size invariance run (`+100 mm` growth) holds the pass
   verdict (`0.0266 ns`, still under the gate). Three genuinely independent
   falsifiers (skipping the phase-unwrap step, dropping the leading minus
   sign, and using the wrong `L_eff` -- domain length instead of
--- a/rfx/api/_sparams.py
+++ b/rfx/api/_sparams.py
@@ -351,7 +351,8 @@ def _warn_if_ringdown_truncated(
         "truncation artifacts wherever the structure is resonant — expect "
         "spurious |S| poles and passivity violations. Increase num_periods "
         "until the witness is below −40 dB before quoting any S value "
-        "(see MSLSMatrixResult.settling_db).",
+        "(see the result's settling_db field — MSLSMatrixResult and, since "
+        "issue #538, WaveguideSMatrixResult).",
         stacklevel=2,
     )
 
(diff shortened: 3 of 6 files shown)


bk-squared:
**Adversarial review round 1 (9fec5af): 1 BLOCKING + 1 MAJOR + 7 MINOR from three independent passes → all fixed in dbbf9aa.**

The catch that mattered — and it was MEASURED, not argued: the witness sampled only `v_probe_t`, but the plain/normalized extraction DFTs the **ref-plane records**; on an iris-resonator fixture the single-series witness read **8.3 dB more settled** than the worst record the extraction actually consumes (a V-node/I-antinode standing wave makes a single-quantity witness a concrete false pass at the −40 dB line). This is the ledger's "two routes sharing the suspect quantity" class, in its witness form: the witness must read the records the extraction CONSUMES.

Fixes: worst over **all four** recorded series per port; flux/normalized variants take the worst of the **device AND reference** runs (the reference produces the normalization — its truncation corrupts the same S; NaN propagates instead of being `max()`-eaten); scope limits stated (modal records only; global-max peak means a weakly-coupled high-Q interior resonance can pass while its narrowband S feature is under-resolved — inherited from the MSL witness this mirrors); the stale support-matrix "has no built-in `settling_db`" parenthetical rewritten as historical; None-enumeration now names both un-adopted paths (NU + uniform multimode); union return annotations; direct extractor-level flag-off/on bit-identity pair replaces the determinism-only pin; measured docstring numbers re-taken with the 4-series witness at the executed parameter (**[-2.3, -1.9] dB** at num_periods=4.0 — less settled than the old single-series read, the review's predicted conservative direction). `3 passed`; api gate + ruff clean.

Round 2 = CI on dbbf9aa; merge on green.

